PR #35, opened 2015-01-20
Pawel

pjanowski:


chemistry/structure.py
```diff
@@ -626,6 +626,11 @@ def print_atoms(atom, coords):
                 return a, dict(), [a.xx, a.xy, a.xz]
         else:
             raise Exception("Should not be here!")
+        # atom names justified according to pdb specs
+        for atom in self.atoms:
+            if not len(atom.name) == 4 and \
+               not len(Element[atom.atomic_number]) == 2:
```
swails: Why `not x == y` instead of `x != y`?

swails:
This commit is breaking several of the tests.  The tests that it is breaking, in particular, is the PDB writing tests.  The way these tests work is to write a PDB to a `cStringIO.StringIO` object (so it's all done in memory) and then feed that object right back to `read_PDB` and make sure that the resulting structures are as equal as they should be (i.e., numbers need not be the same if you asked `write_pdb` to renumber the atoms and residues... likewise with altloc's).

But the change you made modifies `atom.name` of each atom in the Structure to pad with whitespace.  When `read_PDB` process the resulting PDB file, it strips all string properties (which is something you added in this PR), so the atom names of the _new_ `Structure` instance (with no whitespace) are no longer equal to the atom names of the _old_ `Structure` instance (to which you added whitespace when writing the PDB file).

The `write_pdb` method should never change any aspect of the `Structure` class (in C++ parlance, `write_pdb` should be `const`).  Can you move the logic detecting how the string *should* look down to where the writing is actually taking place (e.g., [here](https://github.com/swails/ParmEd/blob/master/chemistry/structure.py#L653) and [here](https://github.com/swails/ParmEd/blob/master/chemistry/structure.py#L637)).

A few more comments: the test for atom name length should be `<4` (rather than `!=4`) when deciding whether or not to pad with whitespace (since we need to protect against long atom names here).  And while you're at it, can you go ahead and make sure you truncate the atom name to 4 characters?  No point in slowing down the writer with a conditional, just always take the slice.  So something like this:

```Python
    if len(pa.name) < 4 and len(Element[pa.atomic_number]) != 2:
        aname = ' %-3s' % pa.name
    else:
        aname = '%-4s' % pa.name
    dest.write(atomrec % (anum, pa.name[:4], ..., res.name[:3], ...)
```

In case you're not aware -- you can make these changes to the branch you submitted the PR from (pawel?) and push them to your repository.  They will automatically be added to this PR and re-run the tests.

pjanowski:
Hi Jason,

Sure I can do this. Just two comments though.

On Tue, Jan 20, 2015 at 5:15 PM, Jason Swails <notifications@github.com>
wrote:

> The write_pdb method should never change any aspect of the Structure
> class (in C++ parlance, write_pdb should be const). Can you move the
> logic detecting how the string *should* look down to where the writing is
> actually taking place (e.g., here
> <https://github.com/swails/ParmEd/blob/master/chemistry/structure.py#L653>
> and here
> <https://github.com/swails/ParmEd/blob/master/chemistry/structure.py#L637>
> ).
>
> The reason I didn't want to do this (it's what I did first), is that it
forces me to repeat the code in 6 or so places.

> A few more comments: the test for atom name length should be <4 (rather
> than !=4)
>
No problem but what's the difference? If x can only be from 0 to 4
characters long, than what's the difference if I do x<4 or x!=4?

Thanks!
Pawel

swails:
> forces me to repeat the code in 6 or so places.

I admit that my loop unrolling was a premature optimization.  The separate code paths for the renumber and non-renumber loops can be fused, and the conditional moved inside the loop.  That should cut down the code duplication considerably.  If you don't feel comfortable refactoring that part, I'd be happy to submit a PR to *your* PR (that's how I can contribute to this PR).  Otherwise, go ahead and bring the renumber conditional into the write loop.

> No problem but what's the difference? If x can only be from 0 to 4
> characters long, than what's the difference if I do x<4 or x!=4?

You are assuming that x can only be from 0 to 4 characters long.  That's true if a `Structure` is instantiated from `read_PDB` or even `AmberParm` (assuming that `AmberParm`'s FORMAT specifier for ATOM_NAME stays at 20a4).  But that's not true in general.  Specifically, CHARMM files support atom names of arbitrary length (and the CHARMM equivalent to gaff -- CGenFF -- uses atom names and types often in excess of 5 or 6 characters).

I also wrote some code instantiating a `Structure` class and adding atoms with the aim of allowing that code to make the PDB for me.  I named the atoms 'C0%d' where %d was the serial number.  When the serial number got above 100, the atom name overflowed and VMD choked.  In cases like these, it's always better to take the potentially route that requires fewer assumptions (assuming similar/same computational cost of each route, of course).

pjanowski:
Hi Jason,

On Tue, Jan 20, 2015 at 6:10 PM, Jason Swails <notifications@github.com>
wrote:

> forces me to repeat the code in 6 or so places.
>
> Only 4 times :)

> I admit that my loop unrolling was a premature optimization. The separate
> code paths for the renumber and non-renumber loops can be fused, and the
> conditional moved inside the loop. That should cut down the code
> duplication considerably. If you don't feel comfortable refactoring that
> part, I'd be happy to submit a PR to *your* PR (that's how I can
> contribute to this PR). Otherwise, go ahead and bring the renumber
> conditional into the write loop.
>
I think it's easier if you refactor the loop. Otherwise I'll just end up
putting just as many conditionals around all the rnums and anums and so
forth. You know better what each of those is actually counting. But maybe
instead of submitting a PR to the PR can't you just accept my PR and then
make your own modification?

Pawel

chemistry/structure.py
```diff
@@ -634,6 +634,8 @@ def print_atoms(atom, coords):
                     anum = (atom.idx + 1 + nmore) % 100000
                     rnum = (res.idx + 1) % 10000
                     pa, others, (x, y, z) = print_atoms(atom, coords)
+                    if len(pa.name) < 4 and len(Element[pa.atomic_number]) != 2:
+                        pa.name = ' %-3s' %pa.name[:4]
                     dest.write(atomrec % (anum , pa.name, pa.altloc,
```
swails: This is where you should put `pa.name[:4]`

swails:
The tests are still failing because you are still modifying the atom.name attribute in write_pdb.  Note my suggestion:

```Python
    if len(pa.name) < 4 and len(Element[pa.atomic_number]) != 2:
        aname = ' %-3s' % pa.name
    else:
        aname = '%-4s' % pa.name
    dest.write(atomrec % (anum, pa.name[:4], ..., res.name[:3], ...)
```

You need to assign to a local variable (in my example, `aname`) rather than using `pa.name`, since `pa` is a pointer to the atom destined to be printed (i.e., _p_ rinted _a_ tom).  Also, the truncation needs to happen in the `dest.write` call.  As it stands, the truncation to 4 letters is put in a code path only reachable by atom names with 3 or fewer letters...

swails:
> I think it's easier if you refactor the loop. Otherwise I'll just end up
> putting just as many conditionals around all the rnums and anums and so
> forth. You know better what each of those is actually counting. But maybe
> instead of submitting a PR to the PR can't you just accept my PR and then
> make your own modification?

Agreed.  It'd be nice if you could fix the issues making the tests fail, then I'll merge the PR and simplify the code so it doesn't give everyone modifying it a headache :).  Thanks!

pjanowski:
Jason,

Sorry about these tests failing. I may not be around much tomorrow so if
you want to go ahead and change the pa.name call, go ahead. Otherwise I'll
do it as soon as I can (probably Thursday).

Pawel

swails:
No problems at all -- I don't have a pressing need for this, so I'll wait until you get to it.  If you'd rather I do it, that's fine, but I'm happy to let you do it if you want.

PR #44, opened 2015-01-24
Fix mask parser

swails:


swails:
The main problem fixed here is that specifying multiple atom names or numbers would get overwritten such that only the last name is recognized.  For instance:

```
:1-3,6@CA,C,O,N
```

was being interpreted as

```
:1-3,6@N
```

This was not present in the last release.  However, residue _names_ did get overwritten; i.e.

```
:1-6,ALA@N
```
Was being interpreted as:
```
:ALA@N
```
This _was_ in the most recent release.

This PR also includes many enhancements to the Amber mask unit testing so this shouldn't happen again.  For even complicated masks, the Amber parser here now completely matches the cpptraj parser.

PR #45, opened 2015-01-28
CHARMM PDB file support

swails:
This adds support for reading (and optionally writing) a SEGID written in columns 73 to 76 (inclusive) in the PDB file. The standard makes no specification for these columns, so we can assign those columns to be a "segid" while still supporting fully compliant PDB files.

Also provide a "`charmm` keyword for `Structure.write_pdb` and `structure.write_PDB` (that defaults to `False`) that will optionally print out the SEGID in columns 73 through 76 (left-justified).

Finally, this changes the `system` string that the `CharmmPsfFile` parsed into this segid so that PSF -> PDB will be more in line with what CHARMM users expect (assuming they use `charmm=True` when writing a PDB file).

Unrelated changes include truncating a bunch of strings in the PDB writing routine to make sure they do not overflow their allotted columns.

### DIFF
--- a/chemistry/charmm/psf.py
+++ b/chemistry/charmm/psf.py
@@ -213,7 +213,7 @@ class CharmmPsfFile(Structure):
             atid = int(words[0])
             if atid != i + 1:
                 raise CharmmPSFError('Nonsequential atoms detected!')
-            system = words[1]
+            segid = words[1]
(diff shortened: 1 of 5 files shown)


swails:
This also adds a unit test for reading and writing SEGID to the PDB file (testing that it doesn't by default and that it does when requested to in a way that the parser understands).

PR #46, opened 2015-02-02
Move omm

swails:
This moves createSystem to `Structure`, effectively merging (and eliminating) `chemistry.charmm.openmmloader` and `chemistry.amber.openmmloader` (and eliminating the need to add a `chemistry.gromacs.openmmloader` if and when that time comes.

All OpenMM tests pass on my machine (they are not run on Travis... I should probably fix that...)


swails:
Of the diffstat here, 5337 lines are simply new test files.

PR #54, opened 2015-02-12
Docs

swails:
Improved docs.


swails:
This PR also fixes a couple bugs, particularly in the chamber action and the stretch-bend API, which needed to be generalized to account for 2 distinct force constants.

PR #80, opened 2015-03-14
Try to upgrade Travis testing to pull in OpenMM and work through conda

swails:


kyleabeauchamp:
Cool.  AFAIK, I think that MSMBuilder currently has the "cleanest" version of our CI system.  It might be useful to build the layout to match that as much as possible.  In particular, I think having the simplified `.travis.yml` and the `devtools` directory are useful for keep things as compartmentalized as possible.  

https://github.com/msmbuilder/msmbuilder/blob/master/.travis.yml

kyleabeauchamp:
Actually, the MDTraj stuff seems to be pretty clean now.

kyleabeauchamp:
PS: I'm totally not trying to micromanage here, but if the layout is nearly identical you can usually convince Robert to help troubleshoot when the Travis CI toolchain runs into issues.

swails:
Not at all, I appreciate the feedback (and if you actually want to use ParmEd, I'd welcome anything that would make it more useful to your work).  My goal is to ultimately get the Travis system looking (and acting) more like what MDTraj does, but I need to wrap my head around what's happening there more (my last issue on MDTraj clearly demonstrates my lack of full understanding there).

The most important thing in this PR is to get the OpenMM stuff tested in Travis, since I've been doing some work there.

kyleabeauchamp:
Yeah, just let me know if you have questions.  I'm currently looking into migrating some of my present GAFF benchmark to use parmed for extracting charges from the prmtop file.

swails:
Thanks -- same to you (as far as having any questions about what ParmEd can do).  Extracting and manipulating charges was actually the original thing I wrote ParmEd for

PR #86, opened 2015-03-19
Fix HAngle constraint assignment.

swails:


kyleabeauchamp:
cool, thanks!

swails:
NP.  The problem in OpenMM is different, but I found it.  I'll issue a PR there tomorrow, since fixing it "right" is more involved than hacking an immediate fix.

In any case, `AmberParm` has nicer behavior than `AmberPrmtopFile`, anyway (like taking an inpcrd/restart in the constructor, assigning default periodic boundaries from the inpcrd rather than the prmtop, substantially faster reading courtesy of a C++ extension, and the ability to modify the topology/parameters before calling `createSystem`)

PR #94, opened 2015-04-03
Add option to parmout and chamber to write VMD-compatible prmtop files

Eigenstate:
VMD has a very strict interpretation of the prmtop format, and ignores the format fields when reading one in. This means that bond data is not successfully parsed from prmtop files that parmed outputs, preventing visualization of the results.

I added an option, -vmd to chamber and parmout that will result in the format fields of the prmtop being what VMD expects. This has the same function as the -vmd flag to standalone chamber.

Example usage:
action = parmout(parm, "test.prmtop test.inpcrd vmd")

### DIFF
--- a/ParmedTools/ParmedActions.py
+++ b/ParmedTools/ParmedActions.py
@@ -238,16 +238,22 @@ class parmout(Action):
     """
     Final prmtop written after all actions are complete
     """
-    usage = '<prmtop_name> [<inpcrd_name>] [netcdf]'
+    usage = '<prmtop_name> [<inpcrd_name>] [netcdf] [vmd]'
     supported_subclasses = (AmberParm,)
-
+    
     def init(self, arg_list):
+        print("INITING\n")
         self.filename = arg_list.get_next_string()
         self.rst_name = arg_list.get_next_string(optional=True)
         if arg_list.has_key('netcdf'):
             self.netcdf = True
         else:
             self.netcdf = None
+        if arg_list.has_key('vmd'):
+            print("Found vmd key\n")
+            self.vmd = True
+        else:
+            self.vmd = None
 
     def __str__(self):
         if self.rst_name is not None:
@@ -261,6 +267,9 @@ class parmout(Action):
         if self.rst_name is not None:
             if not Action.overwrite and os.path.exists(self.rst_name):
                 raise FileExists('%s exists; not overwriting.' % self.rst_name)
+        if self.vmd is not None:
+            print("CONVERTING TO VMD COMPAT\n")
+            self.parm.convert_to_vmd_compat()
         self.parm.write_parm(self.filename)
         if self.rst_name is not None:
             self.parm.write_rst7(self.rst_name, netcdf=self.netcdf)
@@ -3583,6 +3592,7 @@ class chamber(Action):
                       before applying it. This might lead to larger prmtop
                       files, but for large parameter sets will dramatically
                       shorten the running time of the chamber action
+        - -vmd        Writes a VMD-compatible prmtop file
 
     If the PDB file has a CRYST1 record, the box information will be set from
     there. Any box info given on the command-line will override the box found in
@@ -3590,7 +3600,7 @@ class chamber(Action):
     """
     usage = ('-top <RTF> -param <PAR> [-str <STR>] -psf <PSF> [-crd <CRD>] '
              '[-nocmap] [-box a,b,c[,alpha,beta,gamma]|bounding] [-radii '
-             '<radiusset>]')
+             '<radiusset>] [-vmd]')
     needs_parm = False
 
     def init(self, arg_list):
@@ -3693,6 +3703,7 @@ class chamber(Action):
             raise InputError('Illegal radius set %s -- must be one of '
                              '%s' % (self.radii, ', '.join(GB_RADII)))
         self.condense = not arg_list.has_key('nocondense')
+        self.vmd_compat = arg_list.has_key('-vmd')
 
     def __str__(self):
         retstr = 'Creating chamber topology file from PSF %s, ' % self.psf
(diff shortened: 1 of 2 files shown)


chemistry/amber/amberformat.py
```diff
@@ -814,15 +834,24 @@ def write_parm(self, name):
             for i in xrange(len(self.parm_data[self.charge_flag])):
                 self.parm_data[self.charge_flag][i] *= CHARGE_SCALE
 
+        if self.vmd_compat is not None:
+            self.delete_flag('CTITLE')
```
swails: This should be moved into `convert_to_vmd_compat` -- `write_parm` shouldn't modify the parm object (permanently at least).

chemistry/amber/amberformat.py
```diff
@@ -814,15 +834,24 @@ def write_parm(self, name):
...
         # write data to prmtop file, inserting blank line if it's an empty field
         for flag in self.flag_list:
             new_prm.write('%%FLAG %s\n' % flag)
             # Insert any comments before the %FORMAT specifier
-            for comment in self.parm_comments[flag]:
-                new_prm.write('%%COMMENT %s\n' % comment)
+            # except VMD hates comments
+            if self.vmd_compat is None:
+              for comment in self.parm_comments[flag]:
+                  new_prm.write('%%COMMENT %s\n' % comment)
```
swails: Let's just have `convert_to_vmd_compat` delete all comments and get rid of the `vmd_compat` flag altogether.

chemistry/amber/amberformat.py
```diff
@@ -788,6 +787,27 @@ def set_version(self):
 
     #===================================================
 
+    def convert_to_vmd_compat(self):
```
swails: `AmberFormat` is a base class to the actual topology file objects.  `AmberParm` inherits from `AmberFormat` and both `ChamberParm` and `AmoebaParm` inherit from `AmberParm`.  This method is specific to chamber topology files (and will not in fact work with any other flavor).  So can you move this function to the `ChamberParm` class in `chemistry/amber/_chamberparm.py`?

ParmedTools/ParmedActions.py
```diff
@@ -261,6 +265,8 @@ def execute(self):
         if self.rst_name is not None:
             if not Action.overwrite and os.path.exists(self.rst_name):
                 raise FileExists('%s exists; not overwriting.' % self.rst_name)
+        if self.vmd is not None:
+            self.parm.convert_to_vmd_compat()
```
swails: I don't think that `parmout` should modify the topology file -- it goes against what I would expect to happen.  So if you want the `vmd` option in `parmout` (and `outparm`), perhaps make a copy of `self.parm` and convert that to VMD compatibility... but this probably won't support `AmoebaParm` correctly yet...

swails:
The main reason I didn't add this capability when I first implemented the action is because I figured people should just use the PSF file with VMD when visualizing, and a VMD-compatible chamber prmtop would run in Amber and just give *wrong* results -- but not obviously-wrong-or-crash results.

But then I ran into a case from CHARMM-GUI where cross-linked molecules needed to be rearranged in order to work with the `ATOMS_PER_MOLECULE` and `SOLVENT_POINTERS` sections of the prmtop, so the PSF file won't have the same atom ordering as the resulting topology file, so this is a good addition.

A couple questions/comments:

* The `vmd_compat` flag is causing problems.  This attribute doesn't exist unless the `convert_to_vmd_compat` function is called, but it's used in `write_parm` (which means that `write_parm` fails with an `AttributeError` any time you call it on a prmtop file that hasn't been passed to `convert_to_vmd_compat` (that's why the Travis CI tests are failing).  It would be best if `chamber` just took this option and called `convert_to_vmd_compat` and made all of the necessary changes to implement this change directly in that function (rather than offloading some of the work to `write_parm`).
* `convert_to_vmd_compat` is specific to chamber topologies, so it should only be an instance method on `ChamberParm`.
* What, if anything, should we do to "mutilate" the resulting VMD-compatible prmtop so it's valid for visualization but will crash when simulating (make the bond force constants all incredibly large, for instance)?

BTW -- nice to see you on Github!  I hope you're enjoying Stanford

Eigenstate:
Hmmmm. I was actually not aware that the VMD-compatible prmtops aren't read in correctly by sander/pmemd, which is kind of concerning.

Since I've been messing with the VMD code a lot lately, I think I will just fix their prmtop parser so this sort of workaround isn't necessary, and you can table this pull request until then.

Grad school is great! Also github is addictive. Hope you're well :)

swails:
> Hmmmm. I was actually not aware that the VMD-compatible prmtops aren't read in correctly by sander/pmemd, which is kind of concerning.

Well they're read in "correctly" -- but they're read in as a "regular" prmtop file (which means the CHARMM-specific CMAP and improper torsion parameters) are ignored.  So it's "close enough" to a real force field that the simulation doesn't blow up, but obviously without the CMAPs and impropers, it's just wrong.  So there is no error message, the simulation doesn't fail, and the results are not obviously wrong -- the trifecta of "how not what to act like".

> Since I've been messing with the VMD code a lot lately, I think I will just fix their prmtop parser so this sort of workaround isn't necessary, and you can table this pull request until then.

Cool.  If it's of any use, ParmEd has a fairly generalized C++ prmtop parser (that's actually used to accelerate prmtop reading in Python).  The code's in `src/readparm.cpp`, and should be fairly robust against future prmtop extensions.  But if this PR would help you with your work, then I'm more than happy to get it merged in when it's fixed up a bit.

> Grad school is great! Also github is addictive. Hope you're well :)

Good to hear (and agreed -- Github is addictive :)).

swails:
this is stale and unmergeable at the moment, so I'm going to close it.

PR #96, opened 2015-04-06
Try to bring ParmEd's Travis-CI more in line with what mdtraj does

swails:
Particularly with respect to conda. Now that I've used conda a bit more, I
(think I) understand a bit better what to do. Still a bit different, since the
test suite is not a submodule that can be executed directly via nose like mdtraj
is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,13 +2,9 @@ language: python
 
 env:
     matrix:
-        - PYTHON_VERSION=2.6 NO_NUMPY=yes
-        - PYTHON_VERSION=2.6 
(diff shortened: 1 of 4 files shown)


swails:
Next step is to have Travis build the Conda packages

kyleabeauchamp:
We can also just do this on our Jenkins if desired, you'd just need to make
a pr to conda recipes

swails:
Whatever is most convenient.  Now that I've been using conda/miniconda, it should hopefully be easier to  hash this out on Friday.

PR #104, opened 2015-04-13
Initial implementation of support for NAMD binary files. This is not hea...

radakb:
...vily

tested and is mostly designed for a few specific tasks, such as adding or
removing blocks of atoms from .coor and .vel files.

### DIFF
--- a/chemistry/namd/namdbinfiles.py
+++ b/chemistry/namd/namdbinfiles.py
@@ -1,0 +1,95 @@
+"""
+This module contains classes for reading and writing (single frame) NAMD binary
+files.  The main functionality is currently aimed at manipulation rather than
+analysis.
+"""
+from struct import unpack, pack
+
+import numpy as np
+
+
+class NamdBinFile(object):
+    """From the NAMD manual:
+
+        NAMD uses a trivial double-precision binary file format for 
+        coordinates, velocities, and forces ... The file consists of the atom 
+        count as a 32-bit integer followed by all three position or velocity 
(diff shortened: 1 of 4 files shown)


chemistry/namd/namdbinfiles.py
```diff
@@ -0,0 +1,95 @@
...
+        return self.values
+
+
+class NamdBinVel(NamdBinFile):
+    """Class to read or write NAMD "binvelocities" files.
+
+    NAMD internal units are assumed. These can be converted to 
+    Angstrom / picosecond by multiplying by NamdBinVel.PDBVELFACTOR.
+    """
+    PDBVELFACTOR = 20.45482706
```
swails: I don't see where this is used.

chemistry/namd/namdbinfiles.py
```diff
@@ -0,0 +1,95 @@
...
+
+    @property
+    def velocities(self):
+        return self.values
```
swails: Shouldn't this be

```Python
@property
def velocities(self):
    return self.values * self.PDBVELFACTOR
```

?

radakb: I suppose that might be expected behavior. I figured it could be left to the user.

swails:
This seems to be superseded by #108.  For future reference, you can push more commits to a branch you issued a PR from to add to that pull request (i.e., you don't need to create a new branch and issue a new PR every time you want to make a change).

Closing in favor of #108.

PR #108, opened 2015-04-16
Namd files

radakb:
Just adding a few convenience functions for modifying NamdBinFiles on the fly.
Also started working on rudimentary tests.


swails:
There have been a couple changes that are incompatible since this PR was issued, but the conflicts should be very easy to resolve (they *should* just be in `chemistry/__init__.py` and `setup.py` since another `chemistry` sub-package has been added.

There are two issues with this PR right now, both should be easy enough to resolve:

1. Python 3 support is broken.  Python 3 introduced a split between the `bytes` and `str` types (in Python 2, they are the same).  So while reading from a file opened in mode `'r'` in Python 2 yields bytes (since they are strings), doing the same in Python 3 tries to decode them in the default encoding (typically UTF-8).  But what we really want here are raw bytes.  So open the files in mode `'rb'` for reading and `'wb'` for writing to fix this issue (I think).
2. Currently numpy is an optional dependency (and one of the builds in the Travis build matrix tests without numpy), so the unprotected `import numpy` in the `namd` package breaks `chemistry` when numpy is unavailable.

hainm:
@swails 
it's great if you can write one page for ParmEd developers ;)

swails:
A good idea, but harder to do IMO than standard user docs.  Most projects I work with don't have dev docs per se outside of "read the code" (and those that do I haven't found the dev docs to be particularly helpful).

radakb:
Sorry, got distracted from this for quite some time. I think everything should be above board now.

hainm:
@radakb ``parmed`` API changes a lot since you PR your code. 

Especially ``import chemistry`` should be ``import parmed``.

swails:
Also, numpy is now a requirement, so you can just import it now.

swails:
But for this PR, you need to merge master into it and resolve the conflicts (if there are any).

swails:
Ultimately it would be nice to have automatic file-type detection for reading NAMD coordinate and velocity files (so it works with `load_file`).  This is going to be tricky, though, since velocity and coordinate files have the same exact format.  We'd need some kind of heuristic to distinguish between coordinates and velocities that doesn't scale horribly with the number of particles...

I'll have to think about it a little bit.  For now, I'll go ahead and merge this.

PR #110, opened 2015-04-24
Implement fancy indexing for `Structure`

### LINKED ISSUE #78: Enable fancy indexing for `Structure`
One thing that would be nice when manipulating structures is to make them indexable and slice-able.  Here are some ideas, all of which can be implemented side-by-side based on the argument type of the indexer:

- `struct[str]` -- Treat the string as an Amber selection mask
- `struct[int|slice|iter]` -- Select atoms. If it's an `int`, get a single atom. If it's a `slice`, get that slice of atoms.  If it's an iterable, make sure it's length `natom` and return the atoms whose indices evaluate to `True`.
- `struct[int|slice|iter, int|slice|iter]` -- First item in tuple is a residue selector.  Second item in tuple is an atom selector, whose semantics are identical to the last bullet.
- `struct[int|slice|iter, int|slice|iter, int|slice|iter]` -- First item in tuple is now a chain selector. Second item in tuple is a residue selector. Third item in the tuple is an atom selector.  Again, identical semantics as above.

A couple comments:

- Amber masks do not make sense in the context of the chain, residue, and atom slicing, so it is an independent mode.
- The duplication of the topology and parameter data means the slicing will be pretty slow, so that should be kept in mind.
- The idea is that the resulting selection will be a data copy (not much you can do to get around that within the current framework), but it will be a full description with force field and topology data (assuming it originally came with that), so you can immediately dispatch it to OpenMM or the sander API to carry out MM investigations.

If anyone else cares about this feature, input is welcome.

swails:
It is not the most efficient way of doing things, but it's reasonably simple and
fairly expressive. There are a number of hopefully intuitive ways of indexing a
subset of atoms from a given Structure:

By atom
-------
`struct[10]` Returns the 11th Atom instance in the structure

By atom slice
-------------
`struct[:10]` Returns new Structure with 1st 10 atoms and all their parameters

By Amber mask
-------------
`struct[':1-5@CA']` Returns a new Structure with CA atoms of the 1st 5 residues

By chain, residue, and atom indexing/listing/slicing (indexes start from 1 and
are sequential); selections return Structure (if selecting more than 1 atom)
------------------------------------------------------------------------------
`struct[['A', 'B'], 10￼5, :]` all atoms of residues 10-100 (every 5) of chains A and B
`struct[[1, 2, 4, 8, 9], :5]` First 5 atoms of residues 1, 2, 4, 8, and 9
`struct[:,:,:]` Select all atoms

By mask array (mask_array must be same length as # of atoms)
------------------------------------------------------------
`struct[mask_array]` All indices whose values are boolean True map to a selected atom

By selected atom array
----------------------
`struct[selected]` All indices in "selected" correspond to a selected atom.

Closes #78

### DIFF
--- a/chemistry/geometry.py
+++ b/chemistry/geometry.py
@@ -25,10 +25,6 @@ from __future__ import division
 
 from chemistry import unit as u
 from chemistry.constants import TINY, DEG_TO_RAD, RAD_TO_DEG
-try:
-    from itertools import izip as zip
-except ImportError:
-    pass # Py3, zip is already izip
 from math import pi, cos, sin, sqrt, acos
 try:
     import numpy as np
--- a/chemistry/structure.py
+++ b/chemistry/structure.py
@@ -62,6 +62,11 @@ try:
 except ImportError:
     pd = None
 
+try:
+    basestring
+except NameError:
+    basestring = str
+
 # Try to import the OpenMM modules
 try:
     from simtk.openmm import app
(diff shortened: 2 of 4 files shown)


hainm:
nice. Do you have private travis test? I can't see why this commit failed.

Hai

swails:
It should be publicly visible.  Travis failed because I think some of my test cases require a little too much memory for a 4-way build matrix to run at once.  Travis sporadically kills the tests, and I have to re-run them when it happens -- it's unpredictable when they get killed.

The solution, I think, is to use smaller systems for the ParmEd-OpenMM tests (so it doesn't require quite as much memory).  Kind of a PITA to do that, though, so I've been punting on it.

swails:
The tests are all passing except for the one that Travis killed for whatever stupid reason.  I really need to switch to smaller OpenMM test systems. Ugh.

Issue #109, opened 2015-04-24
[FR] Interface with PyRosetta API


cxhernandez:
I posted this on the [MDTraj page earlier](https://github.com/mdtraj/mdtraj/issues/806), but @kyleabeauchamp thought it might be more appropriate here.

It might be nice to have a class that could extract information from a PyRosetta pose object:

``` python
from rosetta import *
from chemistry.rosetta import RosettaPose

init()
pose = pose_from_seq('AAAAAAAA')
ala = RosettaPose(pose)

sys = ala.createSystem(...)
# Continue on to simulate system in OpenMM
```

Let me know if you think this might be worth a PR. Thanks!

swails:
> Let me know if you think this might be worth a PR.

Yea, definitely.  I don't know anything about Rosetta, myself, so it would take me a little while to get to the point where I could add this myself.

If you wanted to have a go at the implementation, I'd be more than happy to field any questions if the (docs)[http://parmed.github.io/ParmEd] are either inadequate or too poorly organized (and of course, doc improvements are *always* welcome).

swails:
This is mostly done in #111.  Let's keep this open with a checklist of what to do next:

- [ ] Update Sphinx docs for this new functionality
- [ ] Add a test case for the "round-trip" Rosetta->ParmEd->OpenMM/mdtraj

cxhernandez:
I'll get started with the Sphinx today.

p.s. I just seeded a simulation of a mutated peptide from WT using this and OpenMM. Seems to work just fine!

swails:
Awesome.  Just out of curiosity, where did you get the parameters?  Did they come from OpenMM (via `ForceField`), or did you get them out of Rosetta somehow?

cxhernandez:
I just used `ForceField`. Rosetta seems to have [ways of extracting CHARMM-like parameters](https://www.rosettacommons.org/node/3725), but I don't know if it's the full set.

cxhernandez:
Should the "round trip" test actually assert anything or just make sure we can start a simulation and load the results into MDTraj?

@kyleabeauchamp

kyleabeauchamp:
Maybe a nan check?  I don't think there's a right answer here, we just want to make sure the molecule is not royally screwed up.

kyleabeauchamp:
E.g. Make sure we can do a single step of MD with a short timestep and a Langevin integrator?

PR #111, opened 2015-04-25
Rosetta

cxhernandez:
This PR adds the ability to load a `Pose` object from PyRosetta into a ParmEd `Structure`, as referenced in https://github.com/ParmEd/ParmEd/issues/109.

Example usage:

``` python
>>> from rosetta import *
>>> from chemistry import load_rosetta

>>> init()

>>> pose = pose_from_sequence(25*"A")
>>> struct = load_rosetta(pose)
>>> struct
<Structure 253 atoms; 25 residues; 252 bonds; NOT parametrized>
```

Still need to add some tests, but let me know what you think @swails @kyleabeauchamp


cxhernandez:
Doesn't seem to like Rosetta's proline for some reason...

swails:
> Doesn't seem to like Rosetta's proline for some reason... I think it's because it includes a virtual atom.

Like a `VirtualSite` in OpenMM?  ParmEd actually supports those (called `ExtraPoint` -- it is an `Atom` subclass).

chemistry/rosetta/pose.py
```diff
@@ -0,0 +1,79 @@
...
+
+from chemistry.exceptions import RosettaError, RosettaWarning
+from chemistry.periodic_table import AtomicNum, Mass
+from chemistry.structure import Structure
+from chemistry.topologyobjects import Atom, Bond
+
+try:
+    from rosetta import Pose, AtomID
+except ImportError:
+    RosettaWarning('Cannot find the PyRosetta module.')
```
swails: Usually what I do is `pass` (or set what I imported to `None`), and then raise an `ImportError` when I try to use functionality in a module that is not available.  See what I do [here](https://github.com/ParmEd/ParmEd/blob/master/chemistry/openmm/reporters.py) with OpenMM.  The only reason I used a decorator was because I had many functions that depended on OpenMM and only wanted to put the check in one place.  In this case, Rosetta is only used once so the decorator approach is probably overkill.

chemistry/rosetta/pose.py
```diff
@@ -0,0 +1,79 @@
...
+        if not isinstance(pose, Pose):
+            raise RosettaError('Object is not a PyRosetta Pose object.')
+
+        struct = Structure()
+        struct.experimental = struct.journal = struct.authors = ''
+        struct.keywords = struct.doi = struct.pmid = ''
+        struct.journal_authors = struct.volume_page = ''
+        struct.title = ''
+        struct.year = None
+        struct.related_entries = []
```
swails: ParmEd makes no guarantees that this metadata exists -- I had need for some of the PDB metadata in the past, so I added it to my PDB parser and just stuck on those attributes.  Only the PDB and CIF parsers add this metadata, so I don't think it's necessary here.

chemistry/rosetta/pose.py
```diff
@@ -0,0 +1,79 @@
...
+
+class RosettaPose(object):
+
+    @staticmethod
+    def load(pose):
+        """ Load a Pose object and return a populated `Structure` class
+        """
+
+        if not isinstance(pose, Pose):
+            raise RosettaError('Object is not a PyRosetta Pose object.')
```
swails: Is `TypeError` more appropriate here?

chemistry/rosetta/pose.py
```diff
@@ -0,0 +1,79 @@
...
+                    try:
+                        for nbr in conf.bonded_neighbor_all_res(AtomID(atno,
+                                                                       resid)):
+                            if nbr.rsd() <= resid and nbr.atomno() < atno:
+                                struct.bonds.append(
+                                    Bond(struct.atoms[sum(
+                                        [pose.residue(i).natoms()
+                                         for i in xrange(1, nbr.rsd())])
+                                        + nbr.atomno() - 1],
+                                        atom))
```
swails: From a voice of experience -- this kind of approach can be an unexpected bottleneck (by avoiding the `sum`, I cut the Amber prmtop parsing down by 60% in d5fffbb763203a5881a592eea4c7575c3bdf3be7).

If performance is not an issue, then there's no reason to change this; just something to keep in mind.

swails:
This looks great!  Thanks!

Some questions about tests:

- Do you want to add them to this PR or another one?
- Is there any way to install Rosetta in Travis for CI?
- If not, perhaps we can prod @kyleabeauchamp or @choderalab to donate some time on their Jenkins server for some extra testing?

kyleabeauchamp:
To start with, I will install pyrosetta this afternoon on my Ubuntu 14.04 desktop and run the tests manually.

cxhernandez:
Okay! Now ```s = load_rosetta(pose_from_sequence("PARMED"))``` works :smiley:

cxhernandez:
As far as installing Rosetta on Travis, we could host a tarball of the binaries someplace (the Rosetta servers are extremely slow), and all the CI would have to do is download it, unpack, and then source a simple bash script.

cxhernandez:
Also, thanks for all the helpful advice @swails!

kyleabeauchamp:
AFAIK we don't have license permissions to redistribute Pyrosetta, so that would require some complex work with private repos and travis keys.  IMHO, we might be forced to simply test locally.  It might also be possible to get some tests running on our Jenkins systems.

cxhernandez:
The license seems to be LGPLv2.

kyleabeauchamp:
Are you sure?

http://www.pyrosetta.org/dow
```
A Rosetta license is required in order to download and use PyRosetta. Licensing is free for academic and non-profit institutions and is available to commercial users for a fee. Academic and commercial licensing of PyRosetta is handled with the standard Rosetta license through Rosetta Commons. 
```

https://c4c.uwc4c.com/licenses/88

cxhernandez:
Yeah, sorry, I realized I was in the wrong directory just after posting that.

kyleabeauchamp:
I ran the example code and saved to PDB, seemed to work fine.  

Regarding tests, IMHO what I'd love is to make sure we can build a protein, solvate it, parametrize it with the amber FF14 force field, and then run it in either sander or openmm.  

The reason I think it's worth doing a  full "round-trip" test is that there are possible issues with atom and residue names which could break either rosetta, amber, or openmm.

cxhernandez:
+1  for "round-trip" test

swails:
> Regarding tests, IMHO what I'd love is to make sure we can build a protein, solvate it, parametrize it with the amber FF14 force field, and then run it in either sander or openmm.

+1 here, too.

I can help with the sander side of things (since it's not *completely* trivial to convert a `Structure` into an `AmberParm` subclass (or more precisely, to instantiate an `AmberParm` from a `Structure`).

So what I see as missing is 3 things:

1. A test case, even if it's disabled in Travis.  How do you want to do this, @cxhernandez?  You can add it to this PR if you want; you can add the example at the top here to this PR and the more sophisticated "Rosetta->ParmEd->OpenMM/sander" test in a separate PR (which is what I would do); or we can punt on all of the tests for another PR (I'd rather have at least *something* in soon, though)
2. Attribution -- @cxhernandez, you can add your name and affiliation either in the `rosetta.py` docstring or comments at the top.
3. Sphinx docs (this can wait)

I have also been looking into downloading and install PyRosetta and I'm looking forward to playing around with this.

kyleabeauchamp:
FWIW, this is what I added to my bashrc to get PyRosetta working on my anaconda python:

```
source /home/kyleb/Downloads/PyRosetta.namespace.ubuntu.release-53/SetPyRosettaEnvironment.sh
export LD_LIBRARY_PATH=$PYROSETTA:$LD_LIBRARY_PATH
```

swails:
Thanks for the tip.  Which version did you download??  Their download page is a mess -- there are like 80 options with incomprehensible version numbers.  (Who thought that was a good idea?)

kyleabeauchamp:
PyRosetta.namespace.ubuntu.release-53.tar.bz2

kyleabeauchamp:
http://graylab.jhu.edu/download/PyRosetta/archive/release/namespace.ubuntu/PyRosetta.namespace.ubuntu.release-53.tar.bz2

swails:
Any reason you chose the namespace over the monolithic option?

kyleabeauchamp:
I actually downloaded both but it seemed like the namespace version had a more recent release number (53).

cxhernandez:
Does this look right?

kyleabeauchamp:
Test runs on my machie

cxhernandez:
Is there a way to turn the test off if the module doesn't load. I tried:

``` python
if __name__ == '__main__':
    if init:
        unittest.main()
```
which works in my python2.7 env, but is failing on travis.

kyleabeauchamp:
Example here: https://github.com/choderalab/gaff2xml/blob/master/gaff2xml/tests/test_openeye.py

kyleabeauchamp:
I think you can also do a raise SkipTest, should be an example somewhere in gaff2xml or in mdtraj

cxhernandez:
:+1:

swails:
> Is there a way to turn the test off if the rosetta module doesn't load? I tried:
> 
```
> if __name__ == '__main__':
>     if init:
>         unittest.main()
```
> which works in my python2.7 env, but is failing on travis.

I'm a little late to the party here, but the reason it fails on Travis is because `nose` imports all of the test modules and goes sniffing for classes and methods it thinks are tests (and then calls them).  The code you added above is only run when the test file is executed directly (or via `python -m`).  Your solution is a good one (my approach is just to delete the methods and classes that run tests that don't satisfy prerequisites, but the `skipIf` way is better).

test/test_chemistry_rosetta.py
```diff
@@ -0,0 +1,48 @@
...
+
+def _unpackLen(obj):
+    return len(list(obj))
+
+
+class TestRosetta(unittest.TestCase):
+    """ Tests loading of a Rosetta pose object """
+
+    @skipIf(not init, "Cannot test load_rosetta module without PyRosetta.")
+    def testLoadPose(self):
```
swails: Add a docstring describing the test?  When you run `nosetests -vs test_chemistry_rosetta.py`, the docstring is printed followed by "ok" or "FAIL", and I like having that description, personally.

test/test_chemistry_rosetta.py
```diff
@@ -0,0 +1,48 @@
...
+        if PDBFile:
+            pdb = PDBFile(get_fn('ala_ala_ala.pdb'))
+            self.assertEqual(_unpackLen(struct.topology.bonds()),
+                             _unpackLen(pdb.topology.bonds()))
+        else:
+            pdb = read_PDB(get_fn('ala_ala_ala.pdb'))
+
+        self.assertEqual(_unpackLen(struct.topology.atoms()),
+                         _unpackLen(pdb.topology.atoms()))
+
```
swails: This doesn't actually work if OpenMM is not installed.  You conditionally use `read_PDB` from ParmEd if `PDBFile` from OpenMM is not available, but then you go ahead and request the OpenMM `Topology` instance from the ParmEd PDB file, anyway.  This will raise an `ImportError` if OpenMM is not available.

Also, this only tests against OpenMM's `Topology` class, not really directly against ParmEd's `Structure`.  I'd suggest splitting this in 2 tests, one that tests directly against `Structure.atoms` and `Structure.residues` and the other that tests against OpenMM's `Topology`, and do the `skipIf(not PDBFile)` for the OpenMM test.

The ParmEd test would look something like this:

```Python
    ...
    self.assertEqual(len(struct.atoms), len(pdb.atoms))
    self.assertEqual(len(struct.residues), len(pdb.residues))
    # No test for bonds, yet (it's functionality I hope to add soon)
```
Then the other test would compare `struct` directly with `pdb.topology` from OpenMM.

What do you think?  I'd be happy to submit a PR to you showing you what I mean if you'd like.

swails:
Looks great!  I don't have PyRosetta installed on any of my machines yet so I can't test it. If you think this PR is good to go, let me know and I'll merge it.

cxhernandez:
Not quite yet, I think. There's something weird still with how prolines are converted.

VMD output for sequence "PARMED" loaded from ParmEd-generated PDB:
``` tcl
Info) Finished with coordinate file parmed_test.pdb.
Info) Analyzing structure ...
Info)    Atoms: 96
Info)    Bonds: 152
Info)    Angles: 0  Dihedrals: 0  Impropers: 0  Cross-terms: 0
Info)    Bondtypes: 0  Angletypes: 0  Dihedraltypes: 0  Impropertypes: 0
Info)    Residues: 6
Info)    Waters: 0
Info)    Segments: 1
Info)    Fragments: 1   Protein: 1   Nucleic: 0
```
VMD output for sequence "PARMED" loaded from Rosetta-generated PDB:
``` tcl
Info) Finished with coordinate file rosetta_test.pdb.
Info) Analyzing structure ...
Info)    Atoms: 95
Info)    Bonds: 147
Info)    Angles: 0  Dihedrals: 0  Impropers: 0  Cross-terms: 0
Info)    Bondtypes: 0  Angletypes: 0  Dihedraltypes: 0  Impropertypes: 0
Info)    Residues: 6
Info)    Waters: 0
Info)    Segments: 1
Info)    Fragments: 1   Protein: 1   Nucleic: 0
```

swails:
What does Rosetta do with the extra point?  Is it added to the PDB file or not?

cxhernandez:
It doesn't seem to export the `VirtualSite`. When I delete that line from the the ParmEd all the number match up. The structures still look odd, but at least they're the identical.

image

cxhernandez:
Would it be best practice to omit the `VirtualSite` by default, and just add a kwarg to the function so the user can add them in if they want to?

swails:
> Would it be best practice to omit the VirtualSite by default, and just add a kwarg to the function so the user can add them in if they want to?

My initial feeling was no (include them by default and provide an option to turn them off), but I may be changing my mind.  Here is my main argument for keeping them by default:

Virtual sites are treated like any other atom -- indeed, whether they are set as a `VirtualSite` in OpenMM or not has no bearing on the computed energies.  The only difference is that their forces are transferred to the atoms defining their frame of reference (conserving overall torque).  ParmEd uses the bond pattern to define the frame (and translates that into a formulation that OpenMM understands).  It's also often nice to be able to visualize exactly where those extra points are.

On the other hand, if "standard" residues start getting decorated with extra points and such, it could easily break other programs.

This is a discussion that is orthogonal to this PR, though, so I think we can merge and bring up another issue to discuss the handling of printing virtual sites to PDB and PDBx/mmCIF files.  What do you think?

cxhernandez:
Sounds good to me!

swails:
Awesome.  Thanks for this!

PR #112, opened 2015-04-26
Add an option to skip tests of very large systems.

swails:
One of the chamber systems I tested was ~60K atoms. When running 4 instances
of the tests, some would occasionally get killed presumably for blowing Travis
memory limits (since it had to store all of the raw data, all of the data
structures, the OpenMM data structures, and their copies in the running
Context).  They are still useful tests, IMO, since it gets every component of
the CHARMM force field in the chamber files. So add an environment variable
option for disabling large tests, and set it in the Travis environment.

Also, `unittest.skipIf` (and `unittest.SkipTest`) don't exist before Python 2.7,
and ParmEd supports at *least* Python 2.6. So add a hacktastic replacement for
the `skipIf` decorator to test/utils.py that behaves *almost* the same way
(basically, if the test is skipped, the decorator prints a message and returns;
so it still prints "ok" at the end, unlike `unittest.skipIf`).

It can't currently decorate a class... we'll see if that's required.


swails:
The Py<=26 `skipIf` implementation *should* now handle decorating classes in Python 2.6 and earlier.

PR #118, opened 2015-04-28
PyRosetta Docs

cxhernandez:
just to get the ball rolling on the rest of https://github.com/ParmEd/ParmEd/issues/109


chemistry/rosetta/pose.py
```diff
@@ -58,18 +58,17 @@ def load(pose):
...
-                                      occupancy=0.0, bfactor=0.0,
-                                      altloc='', number=atnum, rmin=rmin,
-                                      epsilon=epsilon)
+                    atom = ExtraPoint(**params)
                 else:
-                    atom = Atom(atomic_number=atomic_number, name=atname,
-                                charge=0.0, mass=mass, occupancy=0.0,
-                                bfactor=0.0, altloc='', number=atnum,
-                                rmin=rmin, epsilon=epsilon)
-                atom.xx, atom.xy, atom.xz = tuple(at.xyz())
```
swails: Did you mean to get rid of this line?  I don't see where you set the atomic coordinates anywhere else (apologies if I'm just blind)

swails:
I'm not a Sphinx expert, either (I think Robert is much better than I am).  Maybe try replacing `load_rosetta` with `chemistry.load_rosetta`?

I suspect that without that, Sphinx doesn't know what namespace to look for it in.  At some point in the docs you put

```
currentmodule: chemistry.rosetta
```

which probably makes `chemistry.rosetta` the default namespace to look through.  Since that's not where `load_rosetta` resides, I suspect that symbol can't be resolved.

swails:
P.S. Just let me know whenever you decide this is ready to merge.

chemistry/rosetta/pose.py
cxhernandez: Shoot! No you're not–– good catch!

swails:
Maybe it's worth adding a test to make sure the coordinates get moved around correctly somehow?

cxhernandez:
I agree. I'll get on that.

cxhernandez:
Also, changing the module to chemistry in the Sphinx docs fixed the link– thanks!

swails:
I think you still need an entry to rosetta.rst in index.rst, right?

cxhernandez:
Yeah. Any preference for where it goes in the list?

``` markdown
   Core classes used to represent topologies <topologyobjects>
   The core Structure class <structure>
   Working with units <dimensional_analysis>
   The Amber file classes <amber>
   The CHARMM file classes <charmm>
   Using the ParmEd programs (``parmed.py`` and ``xparmed.py``) <parmed>
   The ParmEd API <parmed_api>
   OpenMM Functionality <openmm>
```

swails:
between `Working with units` and `Using the ParmEd programs` somewhere makes most sense to me.

kyleabeauchamp:
1.  Get rid of the platform and props arguments to OpenMM, which are never defined
2.  Get rid of the constraint tolerance argument to OpenMM IMHO--it's pretty arcane for novices
3.  How about adding the following, which applies a mutation?  Just a thought, but it's only two extra lines and it totally showcases what we can do here.

```
from toolbox import mutate_residue

[...]

pose = pose_from_sequence(seq)
mutate_residue(pose, 5, "K")
```

cxhernandez:
+1 for using mutate_residue in the example

kyleabeauchamp:
Need to fix this too, it's wayyy too big and breaks the PBC:

```
nonbondedCutoff=10*nanometers,
```

cxhernandez:
Oops, yeah that was supposed to be a 1

cxhernandez:
What do you think @swails?

doc/omm_rosetta.rst
```diff
@@ -0,0 +1,82 @@
...
+    pose = pose_from_sequence(seq)
+
+
+Mutate a residue
+~~~~~~~~~~~~~~~~
+
+Want to mutate the fifth residue to a lysine? PyRosetta
+makes it as easy as::
+
+    from toolbox import mutate_residue
```
swails: It's really `from toolbox import mutate_residue`?  `toolbox` isn't part of the `rosetta` namespace?  That's almost as crazy as using the name `chemistry` as a package :).

doc/omm_rosetta.rst
```diff
@@ -0,0 +1,82 @@
...
+    system = ff.createSystem(mod.topology, nonbondedMethod=PME,
+                             nonbondedCutoff=1*nanometers,
+                             constraints=HBonds)
+    integrator = LangevinIntegrator(300*kelvin, 1/picoseconds,
+                                    2*femtoseconds)
+
+    system.addForce(MonteCarloBarostat(1*bar, 300*kelvin))
+
+    simulation = Simulation(mod.topology, system, integrator)
+    simulation.context.setPositions(positions)
```
swails: Should this be `struct.positions`?  I may have missed where `positions` was set earlier, but the way I'm reading this now the user would get a `NameError`.

cxhernandez: mod.positions– but a `NameError` nonetheless.

swails: Oh right.  `mod.positions` contains the solvent positions (`struct.positions`) does not.

swails:
@cxhernandez -- would it be easy to run this example yourself and check that it works?  And maybe also add it to `examples/rosetta` once it does?

The example is awesome, btw.

cxhernandez:
I ran it for 500 steps instead of 50000, to save some time, but it works!

cxhernandez:
Should I reference the example in this tutorial?

swails:
> I ran it for 500 steps instead of 50000, to save some time, but it works!

Awesome! (that's what I would've done, too)

> Should I reference the example in this tutorial?

Sure -- you can say that the user can find it in `examples/rosetta` or something.  I should do the same for Amber and CHARMM if I haven't already.

cxhernandez:
Done and done!

cxhernandez:
whoops I wasn't thinking when I typed that shebang

swails:
This looks great

PR #121, opened 2015-04-29
Gromacs support

swails:
Update by merging master.


swails:
This PR does *not* add gromacs support. It is simply updating the WIP branch `gromacs_support` with the latest developments in `master`.

swails:
Can you put this in a separate issue?  It's unrelated to this PR, which is about to be merged and closed.

Slicing is clearly slow for fully parametrized `Structure`s (particularly `AmberParm`) -- the approach is suboptimal, but the easiest way to get it to work correctly with all subclasses.  It does a full copy then a strip of the ignored atoms.  There are clearly better approaches which it seems I will have to implement to make this a useful feature for larger systems.

hainm:
ah, sorry. I though this was only very minor issue so I don't create a new one. 

Hai

swails:
It's definitely not a minor issue. Even so, issues are cheap to open and easy to close. Always err on the side of new issue (especially when it's unrelated to other open issues)

hainm:
got it.

swails:
You're being too nice calling a 10 second slice "minor" ;). I'll profile it shortly to see if there is an easy optimization. 

Thanks for the report in any case.

Issue #123, opened 2015-04-30
Post talk slides


kyleabeauchamp:
John and were talking about how your talk slides had a lot of good "big picture" commentary that was unavailable in the online documentation.  It might be useful to post those slides online and link to them in the README or the docs

swails:
Sure.  Something like what you did for MSMBuilder? (e.g., http://rawgit.com/kyleabeauchamp/MSMBuilderTalk/master/index.html)

I'm also all for doc improvements, if anyone wants to make suggestions (or PRs) related to that

kyleabeauchamp:
I'd be happy with even just dumping the keynote slides somewhere and adding a link--no need for you to go through reformatting.  Then we can clean up the actual docs in the future

swails:
I can do that.  I liked your slide deck in MSMBuilder, so if I find it easy to do that I will. Otherwise I'll just dump a PDF version of the Keynote slides to the website and add a link.

PR #125, opened 2015-04-30
Add a "future directions" and link to PDF slides in homepage.

swails:
This should address #123 to some degree.

The webpage is already live on http://parmed.github.io/ParmEd/html/index.html

@kyleabeauchamp -- how's this?


kyleabeauchamp:
+1

Issue #126, opened 2015-05-01
Spacing in mol2 output


kyleabeauchamp:
I've noticed that the spacing between columns seems to be somewhat nonconventional.  Driver script, input, and output below.  

```
import chemistry

filename = "./A.mol2"
struct = chemistry.load_file(filename)
mol2file = chemistry.formats.Mol2File()

struct.name = "XYZ"

mol2file.write(struct, "./out.mol2")
```

```
head A.mol2

@<TRIPOS>MOLECULE
hexane
   20    19     1     0     0
SMALL
No Charge or Current Charge


@<TRIPOS>ATOM
      1 C1         -0.7500   -0.7230   -0.3380 c3        1 MOL     -0.092500
      2 C2         -0.8330   -0.7250   -1.8560 c3        1 MOL     -0.079500
      3 C3         -1.8740   -1.7250   -2.3580 c3        1 MOL     -0.078600
      4 C4         -1.9540   -1.7250   -3.8850 c3        1 MOL     -0.078600
      5 C5         -2.9950   -2.7250   -4.3870 c3        1 MOL     -0.079500
      6 C6         -3.0780   -2.7280   -5.9050 c3        1 MOL     -0.092400
      7 H1         -0.0020   -0.0040    0.0130 hc        1 MOL      0.032300
      8 H2         -0.4660   -1.7140    0.0280 hc        1 MOL      0.032300
      9 H3         -1.7150   -0.4600    0.1070 hc        1 MOL      0.032300
     10 H4          0.1520   -0.9730   -2.2690 hc        1 MOL      0.038000
```

```
head out.mol2
@<TRIPOS>MOLECULE

20 19 1 0 1
SMALL
USER_CHARGES
@<TRIPOS>ATOM
1 C1 -0.7500 -0.7230 -0.3380 c3 1 XYZ -0.0925
2 C2 -0.8330 -0.7250 -1.8560 c3 1 XYZ -0.0795
3 C3 -1.8740 -1.7250 -2.3580 c3 1 XYZ -0.0786
4 C4 -1.9540 -1.7250 -3.8850 c3 1 XYZ -0.0786
5 C5 -2.9950 -2.7250 -4.3870 c3 1 XYZ -0.0795
6 C6 -3.0780 -2.7280 -5.9050 c3 1 XYZ -0.0924
7 H1 -0.0020 -0.0040 0.0130 hc 1 XYZ 0.0323
[...]

PR #130, opened 2015-05-02
Fix and improve mol2 file writing.

swails:
Fixes mol2 file writing so it works with ResidueTemplateContainer objects as
well as Structure instances.  Still needs some improvement with mol_name
handling (mol2 files offer too many different names for their units... sigh)

It also puts the output in a more tabular format that is easy to grok.  This
format will overflow slightly with over 1M atoms or very long atom names (or
atom type names), but it will still maintain its whitespace delimited behavior
(it just won't look quite as tabular).  But for any files that large, grokking
by eye is unlikely to be nearly as important (and it won't look horrible, it
just won't look as good).


swails:
Closes #126

kyleabeauchamp:
Thanks, I updated my installation and re-ran the one test I have that covers this `nosetests openmoltools -v --exe -a 'parmed'`

PR #134, opened 2015-05-04
Use Vec3 instead of tuples for `positions` for better OMM compatibility

swails:
I suspect this is the culprit in pandegroup/openmm#891


swails:
The failure is a very minor round-off issue.

PR #138, opened 2015-05-11
Gromacs support

swails:
The Gromacs topology file parser is now working, and correctly parses out and populates a `Structure` instance with bonds, angles, dihedrals, exceptions, cmaps, impropers, and even R-B torsions.  Atom types are assigned, but no other parameter types are yet assigned.

Next stage is to implement a Gromacs topology file *writer*, then start to parse in parameter types.

This also adds one test, but more are planned.

### DIFF
--- a/chemistry/__init__.py
+++ b/chemistry/__init__.py
@@ -11,7 +11,7 @@ from chemistry import unit, utils
 from chemistry.topologyobjects import *
 from chemistry.residue import *
 from chemistry.structure import Structure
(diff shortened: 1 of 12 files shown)


hainm:
nice. I will pull after you add test case for Gromacs to play with pytraj too.

swails:
There is a test case already. I just want to add another one. 

Also this PR is not for the master branch (it's for the ParmEd/gromacs_support branch)

PR #156, opened 2015-05-21
Both ftype=3 and ftype=5 are RB torsions

leeping:


swails:
I did not know this.  Thanks.

leeping:
On second thought, even though the function is exactly the same, I'm not sure if there are any parameter transformations before Gromacs actually calculates it.  The Gromacs manual says this:

image

Since I've never seen `ftype=5` used, I'd rather be more conservative and disable it, and just stick with `ftype=3`.  I'll include it in the next PR.

PR #157, opened 2015-05-21
DPPC test case

leeping:
This adds the DPPC test case, and disables `funct = 5` for now because (1) I've never seen it used, and (2) I'm not sure if Gromacs transforms the parameters internally.


leeping:
Note that this test case currently doesn't have matching energies between Gromacs, OpenMM and AMBER.  I think setting `[ pairtypes ]` will help, but the electrostatic energies are also off by about 575 kJ/mol between Gromacs and AMBER (OpenMM mixes it all into `NonbondedForce` so I can't tell the difference.)

PR #152, opened 2015-05-21
Gromacs support

swails:
This PR adds full support for Gromacs topology files and GRO coordinate files to ParmEd.  It also adds a *lot* of other features designed during the process of supporting Gromacs and bug fixes for issues found while supporting Gromacs.  A most likely incomplete list:

- Parse fully parametrized Gromacs topology files
- Parse GRO files
- Fix issue in which distance-based Amber masks would only work on AmberParm instances (now it works for all Structure instances)
- Fix a buggy `update_dihedral_exclusions` method which is often useful when converting between formats
- Fix some issues in `Structure` copying when splitting dihedrals
- Assign parameters in CHARMM PSF files from a *copy* of the parameter set rather than the parameter set itself. This prevents multiple CHARMM PSF instances from claiming the same parameter type instances, which will cause problems (since they can only be "bound" to one list at a time).
- Drastically improves slice performance for `Structure` (about 10x faster)
- Implements split, `+`, and `*` for `Structure` complete with documentation
- Adds Gromacs examples and updates existing examples to make sure it works correctly
- Updates README with contributors
- Update all OpenMM tests to work with the CPU platform, and update tolerances to reduce the occurrence of false failures

Anything else I forgot?

PR #160, opened 2015-05-23
Include a calculation with explicit 1-4 pairs

leeping:


leeping:
https://github.com/ParmEd/ParmEd/pull/152

PR #162, opened 2015-05-29
Add new files to 12B test case

leeping:
This also includes @davidlmobley's `11a` test case, along with the test scripts (which can be simplified.)


leeping:
https://github.com/ParmEd/ParmEd/pull/152

PR #165, opened 2015-06-01
Python 3 compatible except syntax

zonca:
This works also in Python 2.6 and 2.7, **not** in 2.5


swails:
I've already dropped Python 2.6 support and lower. [1]  Officially supported are Python 2.7 and Python 3.3+.

I clearly didn't test xparmed.py thoroughly enough following #124 :).  Thanks.

[1] I know of no issues preventing Python 2.6 from working, but I use `unittest` features only present in Python 2.7 and above, so I can't test 2.6.  Python 2.7 is really the only Python 2 that people should be using anymore, anyway.

PR #167, opened 2015-06-03
Ack.  Update version numbers.  I need a better way of doing this.

swails:


rmcgibbo:
I have used `versioneer` for a couple of new repos, and it's kind of clever, but it didn't seem worth it to retrofit already estabilished projects. https://github.com/warner/python-versioneer

hainm:
for only 2 numbers needed to change, I prefer to manually update.

swails:
Man, I wish I knew about versioneer earlier.  Changing it by hand is fine... as long as I remember to do it.  I personally like the idea of pulling version info from the commit hash for between-release developmental versions.  @rmcgibbo -- is versioneer easy to use?

rmcgibbo:
Yeah, it's straightforward. Very convention-over-configuration.

Issue #169, opened 2015-06-04
2.0 release?


swails:
What do people think about an official release?  There are obviously still lots of things worth adding, but there hasn't been an official release in a very long time.

I cut a beta pre-release in order to put out a conda package for people that needed one, but a real release is warranted, IMO.  I wanted to wait for the gromber command, but now that's done.

kyleabeauchamp:
+1, maybe just wait until someone has tested out the conda packages to make sure there aren't any fixes needed there.

swails:
Sounds good -- @jchodera, have you used the conda packages yet?  And if so, do they work for you?

I'll also try them.

jchodera:
I was able to `conda install parmed` on my `osx` machine, but I don't actually know how to *import* it in Python to check which version number I got.  Apologies, but I'm a total n00b here.

I see `parmed.py` on my path (which is it not just called `parmed` if it is an executable?) though.

davidlmobley:
You want 'chemistry'



—

swails:
The version number is wrong because I forgot to change it.  It will say something like "15.1", unfortunately.  I'm thinking about moving to versioneer on @rmcgibbo's suggestion.

But to do that:

```Python
import chemistry as chem
chem.__version__
```

I'm also thinking about adding a version module that gives more information (kind of like what numpy and Python itself have).  The main motivation here is that ParmEd has its AmberTools versioning and its standalone versioning (yuck).  It's currently a mess, and I don't want any version ambiguity now that more people are starting to use it.

swails:
But more than simply installing it, if you plan on using it on Travis that would be more of a stress-test.

jchodera:
That works!
```
>>> import chemistry
>>> chemistry.__version__
'15.1'
```
Some suggestions (though I am sure you have heard them before):
* It would be useful to have your package name match your import name.  `conda install parmed` to do `import chemistry` makes my brain hurt.
* The version numbers should probably match too.  I'm very surprised that installing version 2.0beta0 yields version 15.1.

jchodera:
> But more than simply installing it, if you plan on using it on Travis that would be more of a stress-test.

@ChayaSt is in the process of setting up travis-ci for her [torsion parameterization tool](https://github.com/choderalab/Torsions), so she'll use `parmed` for that!

swails:
> It would be useful to have your package name match your import name. conda install parmed to do import chemistry makes my brain hurt.

[...]

Issue #85, opened 2015-03-18
repo organization


kyleabeauchamp:
I keep finding it counterintuitive that we have two top level package imports (`import chemistry, ParmedTools`).  Also hard to figure out which things are in which import.

Obviously this is probably a hard-to-change legacy issue, so feel free to close this issue.

swails:
I have no problems reorganizing.  The API is stabilizing, but not fully stable yet, I don't think (I'm about ready to ensure backwards compatibility for all classes except those in `chemistry.tinker`, but I'm not set on package/module arrangement).

I separate these top-level packages in my head -- `chemistry` has all of the parsers, core classes, etc. while `ParmedTools` contains the interpreter machinery and GUI widgets.  The interpreter actions are really the only thing you'd want to import from there, but that works primarily on Amber prmtops (several have been generalized to work on `Structure`, but there's a definite Amber bias in `ParmedTools` that doesn't exist in `chemistry`).

No other package layout seemed superior to me, but as it's been like this for ca. 4 years now, I'm too familiar with it to be a good judge.  I'm open to any ideas/suggestions to improve the layout.

kyleabeauchamp:
I guess my intial expectation was something like:

```
from parmed import chemistry, etc
```

But definitely ask around, as other users might have better thoughts.  Lee Ping and Robert also might be the people to ask.

swails:
Hmm... that's an idea.  I'll have to figure out how to get that to play nicely with `parmed.py` -- it would be too easy to accidentally import the script the way things are laid out now (and neither `parmed.py` nor `xparmed.py` are import-friendly).

Maybe move `parmed.py` into `parmed/__main__.py` and dump a script into the bindir by `setup.py` somehow...

swails:
I thought about this a little more and came up with a potential solution that would implement the behavior you're looking for while still maintaining backwards compatibility... but it doesn't feel all that elegant.

I could make `parmed.py` importable (with its execution protected behind an ``if __name__ == '__main__':`).  So it would look something like this:

```Python
#!/usr/bin/env python
import chemistry
import ParmedTools

if __name__ = '__main__':
    <script>
```

That way, `parmed.py` works both as the "top-level ParmEd namespace" as well as the ParmEd interpreter, but `chemistry` and `ParmedTools` still live in the global import namespace as well.  So then you could do stuff like

```Python
from parmed import chemistry
from parmed import ParmedTools
```

as well as what I usually do

```Python
[...]

PR #173, opened 2015-06-07
Change imports from chemistry -> parmed and ParmedTools -> parmed.tools

swails:


swails:
OK.  This reorganization is finished as described in #85 (and #169).  This represents a major API breakage, with the major differences detailed below:

- `chemistry` package renamed to `parmed`
- `ParmedTools` package moved to `tools` subpackage in `parmed`
- `ParmedTools.ParmedActions` module was renamed to `actions` in the `parmed.tools` package
- `fortranformat` package was moved into `parmed.utils` (but this is only used internally for the AmberParm class).
- Got rid of the redundant (and deprecated) `openmmloader` modules in the `amber` and `charmm` packages (I don't think anybody was using them)

I also took this opportunity to clean up the exception hierarchy and stem the proliferation of exceptions.  I can probably reduce this a little bit more, but it's good for now.

I also added a Raises section to the docstrings of more functions to indicate what possible exceptions may be raised.

Finally, the `setup.py` script will look through all directories in `sys.path` and remove any `chemistry` or `ParmedTools` packages that are present.

The summary of changes necessary to adapt to this new version are:

```Python
import chemistry
# replaced by
import parmed

from chemistry import foo, bar
# replaced by
from parmed import foo, bar

import ParmedTools
# replaced by
import parmed.tools

import ParmedTools.ParmedActions
# replaced by
import parmed.tools.actions
```

Easy fixes are:

```Python
import parmed as chemistry # exactly replaces "import chemistry"
import parmed.tools as ParmedTools # exactly replaces "import ParmedTools"
```

Or in cases where you do `from chemistry import <stuff>`, just replace `chemistry` with `parmed`.

hainm:
+1

davidlmobley:
@swails - could you make a new version tag for this, perhaps v2.0 beta1 or something? I would like to specify what version of ParmEd is required for `openmoltools` and currently I have to say something complicated (v2.0 beta after update #173 or something along those lines).

swails:
You mean cut a new pre-release?  The `parmed.__version__` tag already says "2.0beta1".

davidlmobley:
"Cut a new pre-release" was what I meant, but if you incremented the
version tag at the same time as this change, that would be adequate since
then I can just say 2.0beta1. Is that what happened?

swails:
Done.

PR #177, opened 2015-06-09
parmed hn: fix permission deny for logging + py3 compat for exception

hainm:
Situation: I can not use parmed.py from another user's folder (got permission error).

### DIFF
--- a/parmed/scripts.py
+++ b/parmed/scripts.py
@@ -179,10 +179,9 @@ def clapp():
                                                 now.hour, now.minute, now.second))
                 parmed_commands.setlog(logfile)
                 close_log_file = True
-                could_open_logfile = True
             except:
-                print ("could not open logfile. Skip logging")
-                could_open_logfile = False
+                sys.stderr.write("Could not open logfile. Skip logging")
+                close_log_file = False
         # Loop through all of the commands
         try:
             try:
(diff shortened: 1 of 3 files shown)


parmed/scripts.py
```diff
@@ -171,13 +171,18 @@ def _print_warnings(message, category, filename, lineno, file=None, line=None):
...
+                now = datetime.datetime.now()
+                logfile.write('# Log started on %02d/%02d/%d [mm/dd/yyyy] at '
+                              '%02d:%02d:%02d\n' % (now.month, now.day, now.year,
+                                                now.hour, now.minute, now.second))
+                parmed_commands.setlog(logfile)
+                close_log_file = True
+                could_open_logfile = True
+            except:
+                print ("could not open logfile. Skip logging")
+                could_open_logfile = False
```
swails: Why not get rid of `could_open_logfile` and just set `close_log_file = False` here instead?

Also, can you print `Could not open logfile. Skip logging` to `stderr instead of stdout?

swails:
Good catch!

parmed/scripts.py
hainm: I just updated.

hainm:
uhm, I should use "stderr" more often.

swails:
Thanks!

PR #183, opened 2015-06-10
Properly recognize OMM versions that are too old.

swails:
Gives a helpful error message if OMM is available, but is too old to work with ParmEd.


swails:
And this is why decorators are awesome.

swails:
The current check gets the minimum requirements.  The version is a string, and developmental versions can be challenging to parse.  It's not as easy as Python versions to check.  This is the most reliable way to make sure that the minimum version requirement is met for ParmEd.

PR #200, opened 2015-06-13
Add the ability to save a raw copy of a downloaded PDB

### LINKED ISSUE #196: write_pdb?
motivation: 

```python
import parmed as pmd
pdb = pmd.download_PDB("4PY3")
pdb.write_pdb("4PY3.pdb", mode='original') # just like download pdb file to disk.
```

I thought about this when using ParmEd to get b-factor. But I can't not find the method, so I wish to open the original pdb file and look inside.

Why not downloading myself or using other's program? I am too lazy. :D (open chrome, download pdb, browse to pdb file, use vim to open, ...). 

btw, did I miss that `parmed` read b-factor too?

swails:
Added a "saveto" argument

Fixes #196

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -4,6 +4,7 @@ PDBx/mmCIF files.
 """
 from __future__ import division, print_function, absolute_import
 
+from contextlib import closing
 import io
 import itertools
 import ftplib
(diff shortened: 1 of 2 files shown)


hainm:
+1

PR #208, opened 2015-06-19
Units

swails:
Get numpy.ndarray to work with multiplying units (and Quantities).  Also adds a test.

### DIFF
--- a/doc/dimensional_analysis.rst
+++ b/doc/dimensional_analysis.rst
@@ -113,11 +113,7 @@ You can create a unit in one of two ways:
            Quantity(value=1, unit=/(meter**3))
 
 Seems easy enough, and I usually use method 2 when turning a ``float``, ``int``,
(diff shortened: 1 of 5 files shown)


swails:
The failure here is caused because ParmEd replaces its `unit` package with OpenMM's when it's available (to make sure that it plays nicely with OpenMM).  This change hasn't yet propagated down to OpenMM, so the `openmm-dev` conda package is overriding ParmEd's unit package with a version that doesn't yet support numpy multiplication.

hainm:
merge with failing? (I read your above comment too)

swails:
Yea.  The issue is OpenMM, and I've already submitted the fix there.  Just have to wait for it to propagate to their `openmm-dev` conda build and that will fix ParmEd.

The only reason the tests are there are to prevent bugs from creeping in -- the one failure (which is temporary) is expected and not indicative of a bug, and will go away soon.  It's a minor irritant.

PR #211, opened 2015-06-23
Coordinates

swails:
This is a substantial change to how coordinate handling is done in ParmEd.

- `numpy` is now required, and `coordinates` is now always a `numpy.ndarray` attached to `Structure`
- `Structure` parsers generate a `coordinates` ndarray of shape (natom, 3)
- All trajectory (including single-frame restart "trajectory") file formats generate a `coordinates` ndarray of shape (frame, natom, 3)
- Code relying on coordinate processing has been simplified to take advantage of numpy always being there
- `box` attribute is now always a ndarray (as opposed to *usually* being numpy, which ended up masking a couple bugs -- consistency *really* is the best policy).


swails:
The main thing I don't like here is that some files yield coordinates as a shape `(frame, atom, 3)` while others are `(atom, 3)`, which complicates duck typing.

Possible alternatives:

1. Make every `coordinates` attribute `(frame, atom, 3)`, and those with one frame will always have frame = 1.
2. Have every `coordinates` attribute be `(atom, 3)` for the first frame for every file type, including trajectories, and have a `get_coordinates(frame='all')` method that will return a specific frame's coordinate set with shape `(atom, 3)` or `'all'` coordinates with shape `(frame, atom, 3)`.
3. Leave it what it is now and make people that want to take coordinates from either a PDB/GRO file *or* a restart/trajectory/charmm crd file have to test the shape of the resulting `coordinates`.

The more that I think about it, the more I am leaning toward 2...

hainm:
@swails what is parmed's motivation to have `(frame, atom, 3) shape`? handle a trajectory?

swails:
Yes.  ParmEd has readers for GRO files (which can specify a trajectory), PDB files (which can specify an ensemble as well), and NetCDF trajectories.  Given ParmEd's automatic format detection capabilities, it would be nice for duck-typing to work here.

hainm:
cool. so I can add

```python
import pytraj as pt
traj = pt.load(filename, top_name, engine='parmed')
```

swails:
I wouldn't, but it's possible.  ParmEd doesn't support any formats that mdtraj and cpptraj don't, and doesn't support basic things like iterloading.

You're better off using a real trajectory processing library if you want to analyze a trajectory.  The classes in ParmEd are primarily a convenient way to access the information for use with `Structure` and to provide a means to query velocity and force trajectories (I've used this for debugging before) and support OpenMM reporters (again, supports force and velocity trajectories).

hainm:
i see.

hainm:
hey man,

what do you mean about `cached coordinates`?

swails:
The coordinates are attributes of each atom -- `atom.xx, atom.xy, atom.xz`.  But `Structure` keeps around a private copy of the coordinates in the `_coordinates` attribute, which is the "cached" coordinates.  This is the trick that is used in "memoization" that speeds up a calculation or something (i.e., expensive results are cached and returned when requested so as to avoid having to recompute them).

But if you use cached results, you need to be able to tell when the cache is potentially wrong and then eliminate the cache so the correct results can be regenerated.  That's what my latest commits did.

hainm:
thanks. I used `memoization` before but it's hard to debug. I've heard people suggesting to use `memoization` if the function is pure. 

I will look at your code to see your trick. thx.

swails:
memoization is really just for when a particular quantity is expensive to compute and the added memory used by the cached results is not too large.

There's no point in memoizing a *very* cheap calculation -- it buys you nothing (and often looking up the memoized value may be just as slow as recomputing it, depending on where it's stored).

In a basic sense, writing a trajectory file is a form of memoization -- it's faster to read that file than it is to regenerate the data ;).

hainm:
I used memoizing before to load 3D-array`xyz` coordinates of TrajectoryIterator (which is quite expensive).

swails:
Then it sounds like a good use of memoization

PR #213, opened 2015-06-24
Remove a keyword argument that was causing DihedralType() to crash

leeping:
Hi Jason,

In creating the .frcmod object, I ran into the following error message:

```
leeping@not0rious:~/projects/VSP27-Protein/XMLConvert$ ./oxml-aprm.py Checked/a99sc-v2-r1/a99sc-v2-r1.xml
  File "./oxml-aprm.py", line 124, in <module>
    parameters._DihedralTerm(pk=k, phase=f, periodicity=p, dihtype=dtyp))
  File "/home/leeping/local/lib/python2.7/site-packages/parmed/amber/parameters.py", line 87, in __init__
    improper=improper)
TypeError: __init__() got an unexpected keyword argument 'improper'
```

It turns out that `DihedralType` doesn't take an "improper" argument although `Dihedral` does.  I believe that deleting the keyword argument in `amber/parameters.py` doesn't have any ill effects.  Can you confirm this?

Thanks,

- Lee-Ping

### DIFF
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -83,8 +83,7 @@ class _DihedralTerm(object):
             self.type = dihedraltype
         else:
             self.type = DihedralType(pk / self.idivf, float(periodicity),
-                                    float(phase), float(scee), float(scnb),
-                                    improper=improper)
+                                     float(phase), float(scee), float(scnb))
 
     def parmline(self, multiterm=False):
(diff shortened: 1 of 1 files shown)


swails:
Yes, I believe this is correct.  And there's a test for this functionality, so it clearly hasn't broken anything.  I'm ultimately going to get rid of this class and replace it with a proper `ParameterSet` subclass, but after #211 is done.

Thanks!

PR #223, opened 2015-06-30
Pep8

hainm:
i've tried `autopep8` for `pytraj`, so trying i for parmed too. overall, the new changes are ok but sometimes `pep8` or `autopep8` is very strict with 80-character line.

It's ok if you reject this PR. Just testing.

Hai


parmed/amber/_amberparm.py
```diff
@@ -316,8 +319,8 @@ def from_structure(cls, struct):
         inst.LJ_radius = [0 for i in range(ntyp)]
         inst.LJ_depth = [0 for i in range(ntyp)]
         for atom in inst.atoms:
-            inst.LJ_radius[atom.nb_idx-1] = atom.atom_type.rmin
-            inst.LJ_depth[atom.nb_idx-1] = atom.atom_type.epsilon
+            inst.LJ_radius[atom.nb_idx - 1] = atom.atom_type.rmin
+            inst.LJ_depth[atom.nb_idx - 1] = atom.atom_type.epsilon
```
swails: I think this is actually *not* PEP-8 (I think PEP-8 says no spaces between operands when it's inside selections.  Maybe it's just inside function definitions/function calls.  I can't remember.  I don't like the spaces there, though.

hainm: yeah, I think both of them are fine.

https://www.python.org/dev/peps/pep-0008/
```python
Yes:

ham[1:9], ham[1:9:3], ham[:9:3], ham[1::3], ham[1:9:]
ham[lower:upper], ham[lower:upper:], ham[lower::step]
ham[lower+offset : upper+offset]
ham[: upper_fn(x) : step_fn(x)], ham[:: step_fn(x)]
ham[lower + offset : upper + offset]
```

swails:
I *know* the test files were not PEP-8 (I ignored the 80-column rule there).  And I don't want to make those files PEP-8.

A lot of these changes I like, but the ones I commented on I don't.  Is there any way to selectively choose which PEP-8 changes autopep8 will make?

hainm:
I have not tried different options with autopep8 yet. Hopefully it would be smart.

i used pep8 for pandas files and there are lots of codes not following the recommended rule, especially the whitespace trailing. :d

Hai

hainm:
ah I think you can use `--ignore errors` option

https://github.com/hhatto/autopep8

PR #226, opened 2015-07-01
Attempt to make TrackedList.__getitem__ a little cheaper.

swails:
This gives a slight improvement in indexing a `TrackedList`.  It seems that `TrackedList.__getitem__` is one of the routines that `supercell.parm7` spends the most time in, which indicates it's unnecessarily expensive.  I should probably stop overriding it and just settle with a slice returning a `list` instead of a `TrackedList`...  We'll see how much that breaks.

In the meantime, this gives a ~10% performance boost (reducing parsing time from 10 to 9 seconds on my desktop).  Not too much.

At this point, the bottleneck *seems* to be `Atom` instantiation. There's nothing that I can see to address that short of migrating to Cython.  (Of course, you could just use PyPy instead :)).

### DIFF
--- a/parmed/topologyobjects.py
+++ b/parmed/topologyobjects.py
@@ -3979,10 +3979,9 @@ class TrackedList(list):
         return TrackedList(list.__mul__(self, fac))
 
     def __getitem__(self, thing):
-        retval = list.__getitem__(self, thing)
-        if hasattr(thing, 'indices'):
-            return TrackedList(retval)
-        return retval
+        if isinstance(thing, slice):
+            return TrackedList(list.__getitem__(self, thing))
+        return list.__getitem__(self, thing)
 
     def __getslice__(self, start, end):
         return TrackedList(list.__getslice__(self, start, end))


hainm:
have you ever tried parmed with pypy? I saw its benchmark a while ago and did not have a good impression.

swails:
Yes, I tried a long time ago, and I went ahead and tried it again just now (numpypy is becoming more and more mature).

I used this Python script:

```Python
import parmed as pmd
from time import time

start = time()
pmd.load_file('supercell.parm7')
end = time()
print('Loading supercell.py took %s seconds' % (end - start))
```

Running with Python:

```
swails@batman ~ $ python timeit.py 
Loading supercell.py took 12.9179570675 seconds
```

Running with Python *removing* the `_rdparm.so` accelerated reader (so it's pure Python):

```
swails@batman ~ $ python timeit.py 
Loading supercell.py took 16.5119521618 seconds
```

Running with pypy:

```
swails@batman ~ $ pypy timeit.py 
/usr/lib64/pypy/site-packages/numpy/linalg/_umath_linalg.py:1005: UserWarning: no cffi linalg functions and no _umath_linalg_capi module, expect problems.
  warn('no cffi linalg functions and no _umath_linalg_capi module, expect problems.')
Loading supercell.py took 3.52634000778 seconds
```

So the same pure Python code, unchanged, runs 5x faster with PyPy than with CPython.  And this is real-world Python code, not something hand-tuned to emphasize the awesomeness of PyPy.  It really is quite impressive.

hainm:
wow, that's nice.

hainm:
uhm, no `pypy` with conda in official binstar channel yet.

swails:
Yea, there never will be.  conda is a package manager for CPython (which is the reference Python language implementation).  It was designed specifically *for* CPython and designed to make it easier to distribute extensions for CPython.

pypy is a separate implementation of the Python language (written in RPython).  It's entirely unrelated to CPython.  It doesn't implement the CPython C-API, which means it doesn't work with Cython or many standard extensions.  It's an alternative interpreter altogether. (And it takes a *long* time to compile).  But I've used it for quite a few things when Python was simply too slow for me and I didn't feel like redoing all my work in C++.

PR #227, opened 2015-07-01
Fix documentation and building with pypy

swails:


swails:
Looks like one of the items in the build matrix stalled for whatever reason.  Everything is passing, though.

PR #232, opened 2015-07-02
Fix PDB file parsing when residues have 2 atoms with the same name

swails:
They were previously identified as "alternate locations". But not all PDB files
obey the unique-atom-names-in-each-residue rule, so don't stash atoms as
alternate locations unless the altloc is *actually* defined.


davidlmobley:
It's hard to say that all PDB files do ANYTHING consistently, other than make mistakes. ;)

swails:
> It's hard to say that all PDB files do ANYTHING consistently, other than make mistakes. ;)

Indeed :)

PR #239, opened 2015-07-04
Import tools into the top-level parmed namespace.

### LINKED ISSUE #238: import `tools` to `parmed` namespace?
hey man,

I saw in the amer topic "[AMBER] another leap inconsistency", you posted an example here.

```python
pmd.tools.energy(y, 'omm', 'decompose').execute()
```
but `tools` is not available when `import parmed as pmd`.
(I need to do `from parmed import tools`). If this is what you intend to import to main namespace, we can add to `__init__` file: `from parmed import tools`, right? :p 

I did this and try to PR but got error

```python
  File "./setup.py", line 41, in <module>
    import parmed
  File "/u1/haichit/programs/parmed/parmed/__init__.py", line 17, in <module>
    from parmed import tools
  File "/u1/haichit/programs/parmed/parmed/tools/__init__.py", line 7, in <module>
    from .actions import *
  File "/u1/haichit/programs/parmed/parmed/tools/actions.py", line 6, in <module>
    from parmed import (Bond, BondType, Angle, AngleType, Dihedral, DihedralType,
ImportError: cannot import name 'load_file'
```

the cause of error is you import `parmed` in `setup.py` to get `__version__`. Is this a good idea to import here? (like get version from string parsing).

Hai

swails:
Fixes #238

### DIFF
--- a/parmed/__init__.py
+++ b/parmed/__init__.py
@@ -23,3 +23,7 @@ load_rosetta = rosetta.RosettaPose.load
 
 download_PDB = formats.PDBFile.download
 download_CIF = formats.CIFFile.download
+
+# The tools package depends on *everything*, so import this at the end to avoid
+# circular imports.
+from parmed import tools


parmed/__init__.py
```diff
@@ -23,3 +23,7 @@
 
 download_PDB = formats.PDBFile.download
 download_CIF = formats.CIFFile.download
+
+# The tools package depends on *everything*, so import this at the end to avoid
```
hainm: I see.

PR #241, opened 2015-07-05
Add Structure.save function

### LINKED ISSUE #106: add universal `save` for Parm object
I am playing with ParmEd in pytraj namespace

    >>> from pytraj import io
    >>> p_top = io.load_ParmEd("./Tc5b.top")
    >>> p_top
        <AmberParm 304 atoms; 20 residues; 310 bonds; parametrized>

     >>> # it's really nice if ParmEd has this
     >>> p_top.save("Tc5b.cif")
     >>> p_top.save("Tc5b.psf")
     >>> p_top.save("Tc5b.crazy_ext", fmt='psf') 
     >>> # if fmt is not None, ParmEd will set filetype of fmt

(I know there are methods like `write_cif` in ParmEd)

swails:
This allows you to write the desired format in a *single* place, with any
necessary conversions being done for you automatically. It also handles
Amber-style formats as well as possible, determining in most instances whether a
structure can be cast as a ChamberParm when AmberParm is not flexible enough.
File type is determined from filename extension *or* explicit direction.

This also fixes writing PSF files from Structure instances that were *not*
parsed from PSFs (since the `groups` list did not have a `nst2` attribute.

This fixes #106.

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -1292,16 +1292,16 @@ class Structure(object):
 
     #===================================================
 
-    def save(self, fname, kind=None, **kwargs):
+    def save(self, fname, format=None, **kwargs):
         """
         Saves the current Structure in the requested file format. Supported
         formats can be specified explicitly or determined by file-name
         extension. The following formats are supported, with the recognized
-        suffix and ``kind`` keyword shown in parentheses:
+        suffix and ``format`` keyword shown in parentheses:
 
             - PDB (.pdb, pdb)
             - PDBx/mmCIF (.cif, cif)
-            - Amber topology file (.prmtop/.parm, amber)
+            - Amber topology file (.prmtop/.parm7, amber)
             - CHARMM PSF file (.psf, charmm)
             - Gromacs topology file (.top, gromacs)
             - Gromacs GRO file (.gro, gro)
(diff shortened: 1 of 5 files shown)


swails:
Still needs tests.

hainm:
+1

parmed/structure.py
```diff
@@ -1293,6 +1292,104 @@ def split(self):
 
     #===================================================
 
+    def save(self, fname, kind=None, **kwargs):
```
hainm: why not using 'format="pdb"' instead of using 'kind'?

swails: Format is a better choice. Kind was just the first word that came to mind when I was coding it.

swails:
Travis spuriously killed one of the build matrix.  Merging anyway.

hainm:
it's really convenient to use parmed now.

think about reading gaussian output, .ac, ... files? I am working on modified nuc, there are many steps and programs involve to derive new pars. I wish I only need to remember "parmed". lol.

swails:
I'd like to have parsers for everything, but Gaussian is below many others on the list currently.

leeping:
Hi Hai,

I think that reading / writing quantum chemistry formats is challenging but doable.  The output depends on the type of job being run (energy, gradient, optimization, and properties being a few examples).  There is also the question of what information we should read from the file (e.g. SCF energy, MP2 correlation energy, Mulliken charges, intermediate geometries from optimization being a few examples).  When writing Gaussian input, one needs to specify the type of job being run and other parameters such as total memory. This is all interesting stuff to think about.

Thanks,

- Lee-Ping

hainm:
thanks

Hai

PR #250, opened 2015-07-07
Make cls arg of copy method optional

swails:


swails:
I guess this isn't as easy of a change as I thought.  I'm gonna drop this.

hainm:
@swails 

does 
```python
from copy import copy
pdb0 = copy(pdb)
```
return a deep copy?

swails:
It should.

PR #253, opened 2015-07-10
add load mol2, pdb, .. as Structure

hainm:
What do you think?

### DIFF
--- a/parmed/tools/parmlist.py
+++ b/parmed/tools/parmlist.py
@@ -34,7 +34,7 @@ class ParmList(object):
             if parm in self._parm_names:
                 raise DuplicateParm('%s already in ParmList' % parm)
             try:
-                parm = load_file(parm)
+                parm = load_file(parm, structure=True)
             except FormatNotFound:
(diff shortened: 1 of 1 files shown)


swails:
All keywords that are not consumed by `load_file` are passed on to the parsing function of the parser class.  Only `Mol2File` takes a `structure` argument, so this change actually breaks the `parm` command for every kind of file *except* mol2's.  So this won't work.

I think you need to use `Mol2File.id_format` to determine if it *is* a Mol2 or not, then only pass the `structure` kwarg if that routine returns `True`.

hainm:
thanks. So I should do the same thing for pdb file?

anyway. Seems complicated. I will use ipython instead. :D  (just because `parmed a.mol2` is faster than `ipython`.

swails:
> thanks. So I should do the same thing for pdb file?

No.  mol2 files are the only ones that *can* parse to a `Structure` that don't by default, so only the Mol2 has special handling.  I'll issue a PR for this soon.

hainm:
thanks.

PR #254, opened 2015-07-10
Try to migrate to using Travis's container-based system

swails:
Should give faster builds and more memory (!!)


hainm:
@swails can you  briefly tell about this? "Try to migrate to using Travis's container-based system"

swails:
See https://github.com/omnia-md/omnia-md/issues/16#issuecomment-120420397 where I detailed my experiences with ParmEd as well as Travis's webpage: http://docs.travis-ci.com/user/migrating-from-legacy/?utm_source=legacy-notice&utm_medium=banner&utm_campaign=legacy-upgrade

You can see that link yourself in a message that pops up at the top of the Travis build website for pytraj.  See https://travis-ci.org/pytraj/pytraj/jobs/70357487 for instance.

hainm:
thanks.

hainm:
ah, so 134 files are those test files you wanted to add before (but couldn't due to memory)?

swails:
No, no.  Those 134 test files I had to add because `sudo apt-get install gromacs` doesn't work anymore.  Which means I could no longer rely on Gromacs topology files being present on the Travis build environment.  So I had to add them to my repository to keep Travis happy.

What occasionally happened to me before was that random builds in my build matrix would get killed, presumably because all 4 jobs would just happen to line up such that their combined memory usage was over the allowed limit, so one of them would get killed.  This is much less likely to happen to me now since (a) the containers have 33% more RAM available to them and (b) I removed one of the builds from my build matrix.

hainm:
so why don't you git clone their repo on travis?

swails:
Because all I need are ~1MB of parameter files

hainm:
I see. thanks.

rmcgibbo:
Gromacs is now available on travis: https://github.com/travis-ci/apt-package-whitelist/pull/695

swails:
Thanks!

Issue #256, opened 2015-07-11
summary for structure


hainm:
I think it's good to add `summary` for structure too. parmed has this in interactive mode.

hainm:
cpptraj has this

```python
In [53]: t.dihedral_info()
#       Mask [*] corresponds to 34 atoms.
#Dihedral    pk     phase pn                atoms
         1:  2.000 0.00  1.0 :1@O     :1@C     :2@N     :2@H     (12,11,13,14) O -C -N -H
E        2:  2.500 3.14  2.0 :1@O     :1@C     :2@N     :2@H     (12,11,13,14) O -C -N -H
         3:  0.000 0.00  2.0 :1@C     :2@N     :2@CA    :2@HA    (11,13,15,16) C -N -CX-H1
         4:  0.156 0.00  3.0 :1@HB3   :1@CB    :1@CA    :1@C     (10,7,5,11) HC-CT-CX-C
         5:  0.156 0.00  3.0 :1@HB2   :1@CB    :1@CA    :1@C     (9,7,5,11) HC-CT-CX-C
         6:  0.156 0.00  3.0 :1@HB1   :1@CB    :1@CA    :1@C     (8,7,5,11) HC-CT-CX-C
         7:  0.156 0.00  3.0 :1@HA    :1@CA    :1@CB    :1@HB1   (6,5,7,8) HP-CX-CT-HC
         8:  0.156 0.00  3.0 :1@HA    :1@CA    :1@CB    :1@HB2   (6,5,7,9) HP-CX-CT-HC
         9:  0.156 0.00  3.0 :1@HA    :1@CA    :1@CB    :1@HB3   (6,5,7,10) HP-CX-CT-HC
        10:  0.000 0.00  2.0 :1@HA    :1@CA    :1@C     :1@O     (6,5,11,12) HP-CX-C -O
```

parmed has this in interactive mode but not in normal (python) mode. 

Did I miss anything?

hainm:
this is from `parmed` in python mode

```python
In [67]: p.dihedrals
[...]

PR #260, opened 2015-07-11
Fix up print methods in parmed.tools

swails:
- Avoids O(N^2) string concatenation algorithms for building the string return
  values (which is a consequence of Python strings being immutable). Uses the
  recommended `join()` method instead.
- Changes `printXYZ.__str__` to `__repr__` (and returns the same thing for the
 ` __str__` method) in order to make these methods play nicer in the interactive
  Python interpreter.

### DIFF
--- a/parmed/tools/actions.py
+++ b/parmed/tools/actions.py
@@ -817,14 +817,17 @@ class printInfo(Action):
 
             self.found = True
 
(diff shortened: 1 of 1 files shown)


swails:
Fixes latest comments in #256

PR #262, opened 2015-07-12
*Always* remake the parm data when calling write_parm

swails:
Previously, this was only called if any of the parameter arrays detected
changes, such as deleting an item (like a bond, angle, dihedral, etc.) or one of
the parameter types.  However, *changing* one of the parameter types would not
trigger AmberParm to recognize a change, so the raw data would not be rebuilt
upon writing.  Since this is probably fairly common, we might as well *always*
flush updates to the raw data, even if it's not necessary.


test/test_parmedtools_actions.py
```diff
@@ -2134,23 +2138,21 @@ def testSetMolecules(self):
...
         PT.setMolecules(parm).execute()
 
     def testNetCharge(self):
         """ Test netCharge for AmoebaParm (charge is the monopole) """
         act = PT.netCharge(amoebaparm)
-        chg = act.execute() # check this part of the API
+        chg = act.execute() # check the netCharge.execute return value
         self.assertEqual(str(act), 'The net charge of :* is %.4f' % chg)
         self.assertAlmostEqual(chg, 0.0)
         chg = PT.netCharge(amoebaparm, ':WAT').execute()
```
hainm: can you explain the meaning of 'execute ' word in parmed context?

swails: See [here](http://parmed.github.io/ParmEd/html/parmed_api.html#using-the-actions) and [an example of a typical action](https://github.com/ParmEd/ParmEd/blob/master/parmed/tools/actions.py#L706-L795)

hainm: thx

PR #271, opened 2015-07-14
Enhancements to Residue class and MOL2 file parsing

### LINKED ISSUE #268: .mol2 file charge precision
Hi There,

Would be possible to change the .mol2 format writer and add a parameter that specifies the precision of the written charges? Sometimes the AM1BCC charge calculation does not give exact integer total charge and this could cause problem to tleap. 

Cheers,

Gaetano

swails:
This closes #268 by adding a `fix_charges()` method to `ResidueTemplate` and `ResidueTemplateContainer` in order to force charges to exactly the closest integer value (or, in the case of `ResidueTemplate`, to whatever charge value you want, since residues like terminal amino acids actually have non-integer charges separately, but combine to give an integer charge in the Amber force field).

This implements the API detailed in #268 as well as a corresponding one for `ResidueTemplateContainer` for convenience (which goes through every `ResidueTemplate` and calls `fix_charges()` independently).

This also adds support for parsing MOL2 files from places like the ZINC database that contain MOL2 files with multiple `@<MOLECULE>` sections for each residue template one after the other.

It also adds a keyword `split` in `Mol2File.write` that will output a Mol2 file in this format (i.e., with each residue or residue template written to separate `@<MOLECULE>` sections written one after the other).

The last thing I want to do here is to add a universal `save()` method to `ResidueTemplate` and `ResidueTemplateContainer` that will allow users to save them as any one of the supported file formats (at the moment, just Amber OFF files and mol2 files).

Things left to do
---------------------
- [x] Add universal `save` method to `ResidueTemplate`
- [x] Add universal `save` method to `ResidueTemplateContainer`


davidlmobley:
This sounds great. We'll test soon. Also +1 for handling .mol2 files with multiple molecules - those are common (i.e. typically anyone with pharma connections uses those frequently).

davidlmobley:
Also, am I reading this right that ultimately we should be able to go from a .mol2 file to an .off file? (Or vise versa?) This has often plagued me.

swails:
> Also +1 for handling .mol2 files with multiple molecules - those are common (i.e. typically anyone with pharma connections uses those frequently).

Yea, I saw them a lot after I wrote the original mol2 parser.  At least now they get parsed correctly :).

> Also, am I reading this right that ultimately we should be able to go from a .mol2 file to an .off file? (Or vise versa?) This has often plagued me.

Yes, this can be done (and could be done before this PR).  tleap can also do it, though.  This will be made much easier once I implement the TODO in this PR.

nividic:
Hi Jason,

After to a closer look to the problem we do not think that this is a definitive solution to the issue. We have tested the cyclohexane.mol2 file where the charge fix potentially could still cause problems. Before the fix the net total charge is 0.0005. After the fix the new .mol2 file produces a net charge of -0.000004. This means that if we add 10^4 molecules together this can still trigger the warning/error in tleap. The problem comes from the round off errors during the .mol2 writing process where each charge is rounded to 6 decimal figures. Below I have pasted a possible workaround to your originally fix_charges() solution:

    def fix_charges(self, to=None):

        if not self.atoms:
            raise ValueError('Cannot fix charges on an empty residue')
        net_charge = sum(a.charge for a in self.atoms)
        if to is None:
            to = round(net_charge)
        if net_charge == to:
            return self

        smear = (to - net_charge) / len(self)
        
        sum_c = 0.0
        for atom in self:
            atom.charge = round(atom.charge + smear,6)
            sum_c += atom.charge
        
        smear = int(round(sum_c)) - sum_c
        
        self.atoms[0].charge += smear
        
        return self
        
In this case all the atom charges are rounded to 6 decimal figures and then a new smear is evaluated and added, for example, to the first atom. In this case the net written .mol2 file charge is < 10^(-16)  

Please let us know what you think.

Cheers,

Gaetano

swails:
Good point, I hadn't thought about that.  Your solution looks to be a good one, although I will likely make the precision a `fix_charges` parameter (and actually have it default to 4 rather than 6, since that's the maximum precision that I typically see).

nividic:
Ok, I agree with you, it is better of have a fix_charge parameter to set the precision.

swails:
@nividic -- I've made the suggested change.  Let me know if this works for you.

Thanks!

nividic:
Ok, it works now. You can merge now.

Thanks a lot

Gaetano

swails:
OK, I added the `.save()` attributes to `ResidueTemplate` and `ResidueTemplateContainer`, so it should be quite simple to translate between MOL2 and OFF library files.  I'm going to go ahead and merge this.

PR #279, opened 2015-07-21
Allow URLs to be recognized and opened as standard files

### LINKED ISSUE #269: load from url
does parme support this? something like

```python
mol2 = pmd.dowload_mol2("http://q4md-forcefieldtools.org/REDDB/projects/W-73/tripos1.mol2")
```

swails:
This closes #269 

Some of the features added here:

- Filenames starting with `http://` or `https://` are recognized as URLs and treated accordingly
- Filenames starting with `file:///` are treated as local files, and `file://` is stripped away.
- Compressed remote files are supported and automatically decompressed through either `gzip.GzipFile` or `bz2.BZ2File`. In Python 2, this requires reading the *entire contents* of the remote file into memory before decompressing it, which can be memory intensive.  In Python 3, this is not necessary and so is not done.  Parsing remote compressed files is much better with Python 3 than Python 2, but the user-facing behavior is the same in both.
- Support for opening files in `append` mode is added. This had to be faked for BZ2 files, since `BZ2File` only supports `w` and `r`. This is done by reading the entire existing contents into memory, if it exists, and writing it back to a new file with the same name.


hainm:
Jason

What does 'genopen' stand for?

hainm:
Generator open?

swails:
"GENeral OPEN".  It's primarily for internal use.

hainm:
@swails 

is "url" stuff done yet? how can I load to `Structure` object? I've look at your test files but I bet I miss something.

swails:
I thought that it was :).  The intent was to simply allow you to put in a URL instead of a filename and have that automatically work.  I thought adding it to `genopen` was enough, but I was clearly wrong.

I've added the rest of the necessary code (and tests) that implement this in #281.  Thanks!

swails:
For the record, here is sample code that downloads a R.E.D. mol2 file and instantiates it as a `ResidueTemplate` and `Structure`:

```Python
In [1]: import parmed as pmd

In [2]: pmd.load_file('http://q4md-forcefieldtools.org/REDDB/projects/W-73/tripos1.mol2')
Out[2]: <ResidueTemplate DAN: 31 atoms; 33 bonds; head=None; tail=None>

In [3]: pmd.load_file('http://q4md-forcefieldtools.org/REDDB/projects/W-73/tripos1.mol2', structure=True)
Out[3]: <Structure 31 atoms; 1 residues; 33 bonds; NOT parametrized>
```

hainm:
cool. this is what I am expecting.

hainm:
what do you think by changing 'structure=True' to 'as_structure=True'?

swails:
I'm fine with `structure`.  If I was doing it over again, you'd probably persuade me to pick `as_structure`.  But I don't think the current name is misleading, it's a little shorter, and it's been that way for a little while now.

parmed/utils/io.py
```diff
@@ -13,44 +13,127 @@
...
-    code is Py2-Py3 compatible.
+    extension. Note, these files are opened in a way that *always* returns a
+    string. This is an important distinction in Python 3 where many file-like
+    objects return bytes instead of strings. This is detected and handled
+    properly so that the object returned from this function is always
+    string-based (so you cannot write or read bytes directly from a file opened
+    via ``genopen``).
+
+    This routine also recognizes URLs and will read remote files when given a
+    URL starting with either http:// or https://. Like with standard local file
```
hainm: add 'ftp' too?

swails: Can you open a separate issue for that?

PR #281, opened 2015-07-22
Implement remote file support in load_file

swails:
A couple notes here:

- Remote NetCDF files do not work -- the NetCDF implementations require a fd
  attribute, which urlopen-returned objects don't have (at least in Py2; Py3
  hasn't been checked)
- The optimized Amber topology file reader only supports local files, so urls
  force the use of the "slow" (i.e., pure python) reader. In this case, the
  parser will likely not be rate-limiting (the network will be).

Tests were added first in TDD-style.

### DIFF
--- a/parmed/amber/amberformat.py
+++ b/parmed/amber/amberformat.py
@@ -427,10 +427,11 @@ class AmberFormat(object):
         except ImportError:
             return self.rdparm_slow(fname)
 
-        if slow:
+        # The optimized parser only works on local files
+        if slow or fname.startswith('http://') or fname.startswith('https://'):
             return self.rdparm_slow(fname)
 
-        # We have the optimized version
+        # We have the optimized version and a local file
         try:
             ret = _rdparm.rdparm(fname)
         except TypeError:
--- a/parmed/amber/netcdffiles.py
+++ b/parmed/amber/netcdffiles.py
@@ -194,7 +194,13 @@ class NetCDFRestart(object):
         -------
         is_fmt : bool
             True if it is an Amber NetCDF restart file. False otherwise
+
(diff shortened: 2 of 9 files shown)


swails:
This really *does* close #269

hainm:
> Remote NetCDF files do not work -- the NetCDF implementations require a fd attribute, which urlopen-returned objects don't have (at least in Py2; Py3 hasn't been checked)

what's about saving it to /tmp folder and load? 

> The optimized Amber topology file reader only supports local files, so urls force the use of the "slow" (i.e., pure python) reader

same idea as above.

swails:
> what's about saving it to /tmp folder and load?

Too awkward for file format autodetection.  I could do it for the standard file constructors, but I won't until it's needed.  It may be that Python 3 *will* work... I should try it.  [1]

> same idea as above.

Unlikely to improve parsing performance.  The C++ parser is ~5x faster than the pure Python parser IIRC.  But if the download is about the same expense (or even more expensive) than the pure Python parsing, then the benefit of using the C++ optimized parser is substantially diminished.

Basically, if someone is willing to wait for a prmtop file to download, they won't care if the text parsing is a little slower :).

But this reminds me, I should test a gzipped and bzipped prmtop file -- I doubt the optimized parser handles those, either.

hainm:
yeah, should encourage people to use Py3. It has so many cool features.

python 3.5 even has async.

http://makina-corpus.com/blog/metier/2015/python-http-server-with-the-new-async-await-syntax

swails:
Yea, I lurked on the python-dev discussion about the PEP.  It's not on my radar to try yet since it's so new, but it could be cool.

PR #282, opened 2015-07-22
Fix parsing of compressed prmtops

swails:
The optimized parser does not work for compressed files. This would be a good
thing to add in the future, but will take me awhile (even if I pilfer cpptraj's
infrastructure for it).  Much easier to simply divert parsing those files to the
pure Python parser that already handles it natively.


swails:
I also added tests for it.

PR #284, opened 2015-07-22
Make SeriousParmWarning fatal by default in Python

### LINKED ISSUE #277: pmd.tools.printBonds(p, 1)
Jason,

what should I expect when using above command?

It's likely that there is no different between 
`pmd.tools.printBonds(p, 1)` and `pmd.tools.printBonds(p, 100)`

swails:
Fixes #277

### DIFF
--- a/parmed/scripts.py
+++ b/parmed/scripts.py
@@ -111,8 +111,8 @@ def clapp():
         print(splash)
     
     # Set our warning filter
-    if opt.strict:
-        warnings.filterwarnings('error', category=SeriousParmWarning)
+    if not opt.strict:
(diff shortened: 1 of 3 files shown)


swails:
The failure was a spurious Travis build error.  Those are really irritating.  I'm not sure if it's some weird OpenMM issue (since it always bombs when running an OpenMM test) that's caused by collisions with other build matrix elements going at the same time... perhaps it's just unlucky (and the new container system has more reliable runtimes and start times, so they line up better; or worse; than on the old architecture).  There aren't *actually* any problems, though.

PR #301, opened 2015-07-24
Always allow `structure` keyword in `load_file` and ignore it when it's not used

### LINKED ISSUE #290: better **kwargs for load_file
this is the reason why

```python
In [4]: pmd.load_file("http://ambermd.org/tutorials/basic/tutorial4/files/sustiva.mol2",
structure=True)
Out[4]: <Structure 30 atoms; 1 residues; 32 bonds; NOT parametrized>




In [5]: pmd.load_file("../tests/data/tz2.pdb", structure=True)
TypeError                                 Traceback (most recent call last)
<ipython-input-5-993ae273f07f> in <module>()
----> 1 pmd.load_file("../tests/data/tz2.pdb", structure=True)

/home/haichit/anaconda3/lib/python3.4/site-packages/parmed/formats/registry.py in 
load_file(filename, *args, **kwargs)
    129                             name, arg)
    130     if hasattr(cls, 'parse'):
--> 131         return cls.parse(filename, *args, **kwargs)
    132     elif hasattr(cls, 'open_old'):
    133         return cls.open_old(filename, *args, **kwargs)

TypeError: parse() got an unexpected keyword argument 'structure'
```

I would like to see 'structure=True' work for `load_file` with all filenames. If not, `pytraj` need to check if the file is mol2 or not (by checking if "ResidueTemplate"). "ResidueTemplate" does not have "save" method. (pytraj use `save` to save `Structure` to pdb file and reload).

swails:
This closes #290 (and other related threads; it's been brought up a couple times).

This is done in a way that should be future-proof against any parser that may support that argument in the future.  It looks at the argument names and discards `structure` from `**kwargs` IFF `structure` is not found in that argument list.

### DIFF
--- a/parmed/formats/registry.py
+++ b/parmed/formats/registry.py
@@ -127,10 +127,22 @@ def load_file(filename, *args, **kwargs):
         if not arg in kwargs:
             raise TypeError('%s constructor expects %s keyword argument' %
                             name, arg)
+    # Pass on the "structure" keyword IFF the target function accepts a target
+    # keyword. Otherwise, get rid of it.
     if hasattr(cls, 'parse'):
+        _prune_structure(cls.parse, kwargs)
         return cls.parse(filename, *args, **kwargs)
     elif hasattr(cls, 'open_old'):
+        _prune_structure(cls.open_old, kwargs)
         return cls.open_old(filename, *args, **kwargs)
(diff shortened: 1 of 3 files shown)


hainm:
+1. I will update `pytraj` code after you merge.

PR #303, opened 2015-07-26
Improvements to CIF and PDB parser/writer

### LINKED ISSUE #297: load_PDB: add 'resolution' for x-ray
```python
p = pmd.download_PDB("4GLC")
print (p.resolution)
```

swails:
The following additions were made to the PDB and CIF parsers:

- Parse X-Ray resolution from both CIF and PDB files with test cases (Closes #297)
- Fixes CIF parser when molecule has multiple models (e.g., PDB ID 2KOC), and adds test
- Adds ability to write PDB and CIF files with multiple `MODEL`s, with corresponding tests
- Simplifies coordinate handling in the writing routines by eliminating the now-vestigial numpy-free support. Just use `ndarray` for all of the coordinates, which significantly simplifies the code.

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -936,7 +936,13 @@ class CIFFile(object):
                                                auth.getRowList()])
             reflns = cont.getObj('reflns')
             if reflns is not None:
-                struct.resolution = float(reflns.getValue('d_resolution_high'))
+                res = reflns.getValue('d_resolution_high')
+                if res != '?':
+                    try:
+                        struct.resolution = float(res)
+                    except ValueError:
+                        warnings.warn('Could not convert resolution (%s) to '
+                                      'float' % res)
             cite = cont.getObj('citation_author')
             if cite is not None:
                 nameidx = cite.getAttributeIndex('name')
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -6,7 +6,6 @@ from __future__ import division, print_function, absolute_import
 
 from contextlib import closing
 import io
-import itertools
 import ftplib
 try:
     import gzip
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -12,11 +12,7 @@ try:
     import gzip
 except ImportError:
     gzip = None
-try:
-    import numpy as np
-    create_array = lambda x: np.array(x, dtype=np.float64)
-except ImportError:
-    create_array = lambda x: [float(v) for v in x]
+import numpy as np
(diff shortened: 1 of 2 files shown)


swails:
Failure was a rare round-off issue.  This can be merged.

hainm:
@swails 

how rare?

swails:
I don't know... maybe 5% of the time?  Hasn't been often enough for me to worry about changing it (and it's not always the same routine, I don't think).

If it happens on a merge, I just re-run that part of the build matrix and it goes away.  The OpenMM-related travis hangs are much more frequent (and I don't know how to get rid of those).

hainm:
Did you use random number in some cases?

swails:
Yes, I use random numbers fairly frequently.  But I don't think that's necessarily where it's failing.

Random numbers are great if you're just trying to make up data, because I've found that it increases the likelihood that you'll eventually catch a bug just by running the tests enough.

hainm:
but it will introduce bugs too. I used to use random number to make fake trajectory. But sometimes the trajectory coordinates are really bad, and got NaN result.

But it depends on application, might be good in your cases.

swails:
I am careful to make sure that the random numbers are picked within a plausible range for the real data to occur and that it I can't think of an instance where it would cause a false failure.

But having a test fail because the random number is a little too "loose" with where it can go is better than having the random number distribution too tight to reveal bugs.  The random numbers haven't caused me problems yet...

mojyt:
Since no RNG is truly random and should always produce the same stream of
numbers for a given seed it is perfectly legitimate to use them in tests
(and even desirable to make sure they work well). So as long as the seed is
specifiable everything should be ok. One of my gripes about ptraj way back
when was that you couldn't specify seeds for certain things like 'cluster
sieve random'; this is why the seed is always able to be user-specified in
cpptraj.

On Saturday, July 25, 2015, Jason Swails <notifications@github.com> wrote:

> I am careful to make sure that the random numbers are picked within a
> plausible range for the real data to occur and that it I can't think of an
> instance where it would cause a false failure.
>
> But having a test fail because the random number is a little too "loose"
> with where it can go is better than having the random number distribution
> too tight to reveal bugs. The random numbers haven't caused me problems
> yet...
>
> —
> <https://github.com/ParmEd/ParmEd/pull/303#issuecomment-124945857>.
>


-- 
-------------------------
Daniel R. Roe, PhD
Department of Medicinal Chemistry
University of Utah
30 South 2000 East, Room 307
Salt Lake City, UT 84112-5820
http://home.chpc.utah.edu/~cheatham/
(801) 587-9652
(801) 585-6208 (Fax)

PR #304, opened 2015-07-26
Fix Mol2 file writing when `Structure` does not have defined types

### LINKED ISSUE #287: cpptraj can not recognize residue from mol2 file saved by parmed
```python
In [1]: p = pmd.load_file("./data/tz2.pdb")

In [2]: p.residues
Out[2]:
ResidueList([
        <Residue SER[1]>
        <Residue TRP[2]>
        <Residue THR[3]>
        <Residue TRP[4]>
        <Residue GLU[5]>
        <Residue ASN[6]>
        <Residue GLY[7]>
        <Residue LYS[8]>
        <Residue TRP[9]>
        <Residue THR[10]>
        <Residue TRP[11]>
        <Residue LYS[12]>
        <Residue NME[13]>
])

In [3]: p.save("./tz2.mol2")

In [4]: top = pt.load("./tz2.mol2")

In [5]: top.residue_info()
#Res  Name First  Last Natom #Orig  #Mol
    1 ▒n▒     1   223   223     0     1
```

do @swails  and @mojyt have any comment about this?

'tz2.pdb' is here: https://github.com/pytraj/pytraj/blob/master/tests/data/tz2.pdb

swails:
Fixes #287

### DIFF
--- a/parmed/formats/mol2.py
+++ b/parmed/formats/mol2.py
@@ -444,7 +444,8 @@ class Mol2File(object):
                     except AttributeError:
                         z = 0
                     dest.write('%8d %-8s %10.4f %10.4f %10.4f %-8s %6d %-8s' % (
-                               j, atom.name, x, y, z, atom.type, i+1, res.name))
+                               j, atom.name, x, y, z,
+                               atom.type.strip() or atom.name, i+1, res.name))
                     if printchg:
                         dest.write(' %10.6f\n' % atom.charge)
                     else:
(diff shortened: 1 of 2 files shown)


swails:
Another random Travis failure... sigh.

hainm:
how random? 

I dont know about behind the scene in case of parmed. But in the case of pytraj, there is always a reason to fail. For example, sometimes pytraj causes segfault and travis just wait for a never-existing-output; so the test process was terminated with exit 1.

swails:
It happens on maybe 25-30% of all builds that use OpenMM as long as I have at least two builds using OpenMM.  Quite annoying, really, but the alternative is to either not test with OpenMM on either Python 2 or Python 3, and I'd rather get both and occasionally restart some builds.

Hopefully when the new OpenMM 6.3 conda build goes live, I can switch to that and have that solve my problems.  There have been some fixes to OpenMM between the last openmm-dev conda package and OpenMM 6.3, I think...  So hopefully that means this will get fixed.

PR #311, opened 2015-07-29
add Ftp file loading

hainm:


swails:
Ah cool.  If `urlopen` works for files on an FTP server, then this is a nice solution (the PDB/CIF downloader uses `ftplib`).

There are a couple other places where URLs are handled separately from local file names -- a quick `grep` shows `parmed/amber/netcdffiles.py` and `parmed/amber/amberformat.py`.  If you add the `ftp://` condition to the URL handlers there, I'd be happy to merge this :).

Thanks!

hainm:
Sure

Hai

hainm:
@swails 

add "ftplib" too?

swails:
> add "ftplib" too?

No -- `ftplib` is a Python module in the standard library that allows for interacting with FTP servers.

parmed/amber/netcdffiles.py
```diff
@@ -200,6 +200,7 @@ def id_format(filename):
         Remote NetCDF files cannot be loaded
         """
         if filename.startswith('http://') or filename.startswith('https://'):
+            or filename.startswith('ftp://') or filename.startswith('ftplib'):
```
swails: `ftplib` is a Python module, not a filename prefix.  Also, this is a `SyntaxError`.

parmed/amber/netcdffiles.py
```diff
@@ -543,7 +544,8 @@ def id_format(filename):
         -----
         Remote NetCDF files cannot be loaded
         """
-        if filename.startswith('http://') or filename.startswith('https://'):
+        if filename.startswith('http://') or filename.startswith('https://')
+            or filename.startswith('ftp://') or filename.startswith('ftplib'):
             return False
```
swails: Same as above (and still a syntax error)

hainm:
don't have 'ftp' test cases for amber and netcdf file yet (since i can not find one). :D

swails:
> don't have 'ftp' test cases for amber and netcdf file yet (since i can not find one). :D

That's fine -- it all uses `urlopen`, anyway.

parmed/amber/netcdffiles.py
hainm: yeah, already has new commits :P

swails:
Once Travis passes, I'll merge this.

PR #327, opened 2015-08-04
chamberparm's FORCE_FIELD_TYPE should now always contain CHARMM

KarlTDebiec:
Hi Jason,

This addresses an issue encountered when using parmed's chamber command with certain parameter files, including 'par_all36_cgenff.prm' and 'par_opls_aam.inp'. If I understand correctly, the use of the CHARMM force field and its unique terms with a given parm file is triggered by the presence of the string 'CHARMM' under the flag 'FORCE_FIELD_TYPE'. For most CHARMM force field files (e.g. CHARMM36), this will be already be present in the first line of the header of the parameter infile, which parmed copies to this flag. If it is not present in that line, parmed prepends it. However, this action is taken inside a loop over a list that may be empty. With this update, parmed now checks that this array contains at least an empty string.

Best,
Karl

### DIFF
--- a/parmed/amber/_chamberparm.py
+++ b/parmed/amber/_chamberparm.py
@@ -761,7 +761,6 @@ def ConvertFromPSF(struct, params, title=''):
         if 'CHARMM' not in pset: # needed to trigger "charmm_active"...
             pset = 'CHARMM: %s' % pset
         fftype.extend([len(params.parametersets), pset])
-    print(fftype)
     # Convert atom types back to integers if that's how they started
     if int_starting:
         for atom in struct.atoms:
--- a/parmed/amber/_chamberparm.py
+++ b/parmed/amber/_chamberparm.py
@@ -755,11 +755,13 @@ def ConvertFromPSF(struct, params, title=''):
             atom.type = str(atom.atom_type)
     parm = ChamberParm.from_structure(struct)
     parm.parm_data['FORCE_FIELD_TYPE'] = fftype = []
+    if params.parametersets == []:
+        params.parametersets.append('')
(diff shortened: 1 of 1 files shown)


KarlTDebiec:
I should add that the list ends up empty because of how `charmm/parameters.py` looks for the header:
```python
if parameterset is None and line.strip().startswith('*>>'):
                parameterset = line.strip()[1:78]
                continue
```
This catches: 
```
*>>>> CHARMM36 All-Hydrogen Parameter File for Proteins <<<<<<<<<<
*>>>>> Includes phi, psi cross term map (CMAP) correction <<<<<<<<
*>>>>>>>>>>>>>>>>>>>>>>>>>> Feb. 2012 <<<<<<<<<<<<<<<<<<<<<<<<<<<<
```
but misses:
```
*  --------------------------------------------------------------------------  *
*         CGenFF: Parameters for the Charmm General Force Field v. 2b8         *
*                    for Small Molecule Drug Design                            *
*  --------------------------------------------------------------------------  *
*
```
Checking for `*` instead of `*>>` may be preferable, but I thought a test before writing the flag was appropriate.

swails:
Looks good.

> Checking for `*` instead of `*>>` may be preferable, but I thought a test before writing the flag was appropriate.

The reason I didn't do that in the beginning is because in the CHARMM files, \* is a very common start to a lot of lines.  I'm really not too worried about what goes into the `FORCEFIELD` flag, as long as it doesn't prevent sander from running.

I'm going to go ahead and merge this.  If you find something that works better for you, you're more than welcome to submit a PR for it :).

Thanks!

PR #307, opened 2015-07-28
Implement CHARMM file writing support

swails:
Closes #305

PR #331, opened 2015-08-07
Support for cosine impropers in chamber

KarlTDebiec:
Hi Jason,

I've worked through what I think should be the last issue for getting OPLS-AA/M psf topologies converted for use with AMBER: support for cosine impropers in chamber. This needs a little more testing before it is ready to merge, which I wanted your advice on.

I've updated chamber's `par` parsing to check the periodicity, which is how harmonic (n=0) and cosine (n>0) impropers are differentiated in CHARMM. Terms identified as harmonic are stored in `improper_types`, while those identified as cosine are stored in `dihedral_types`. If I understand correctly this should be in line with how cosine impropers are handled from AMBER files.

Parsing the `psf` is a little more complicated, since it isn't initially known what style of impropers are present. I've rearranged the code to loop over the topology's impropers before the dihedrals; cosine impropers are added to the list of dihedrals as they are identified, and then handled alongside the proper dihedrals.

I've tested this with an Ala3 tetrapeptide, and the dynamics look reasonable. I've run the same system using NAMD with the same source psf and pdb; would you have any advice on the best/easiest way to loop over the frames of this trajectory and validate that AMBER gives the same energies?

Best,
Karl

### DIFF
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -234,16 +234,16 @@ class CharmmParameterSet(ParameterSet):
                 parameterset = line.strip()[1:78]
                 continue
             # Set section if this is a section header
-            if line.startswith('ATOMS'):
+            if line.startswith('ATOM'):
                 section = 'ATOMS'
                 continue
-            if line.startswith('BONDS'):
+            if line.startswith('BOND'):
                 section = 'BONDS'
                 continue
-            if line.startswith('ANGLES'):
+            if line.startswith('ANGLE'):
                 section = 'ANGLES'
                 continue
-            if line.startswith('DIHEDRALS'):
+            if line.startswith('DIHEDRAL'):
                 section = 'DIHEDRALS'
                 continue
             if line.startswith('IMPROPER'):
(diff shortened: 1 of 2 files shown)


swails:
I think this is colliding a little bit with #307.  In particular, I have a separate dict in `ParameterSet` for non-harmonic impropers (`improper_periodic_types`) that are defined using cosines instead of a harmonic function.

> Parsing the psf is a little more complicated, since it isn't initially known what style of impropers are present. I've rearranged the code to loop over the topology's impropers before the dihedrals; cosine impropers are added to the list of dihedrals as they are identified, and then handled alongside the proper dihedrals.

Let's add these to the `improper_periodic_types` dict instead.  That should make the assignment of improper torsions easier and eliminate the possibility of an improper periodic term overwriting (or being overwritten by) a proper periodic torsion.  It should be reasonably simple:

1. Look for harmonic improper term with exact match
2. Look for periodic improper term with exact match
3. Look for harmonic improper term with wildcards
4. Look for periodic improper term with wildcards

> I've run the same system using NAMD with the same source psf and pdb; would you have any advice on the best/easiest way to loop over the frames of this trajectory and validate that AMBER gives the same energies?

Either use "imin=5" with sander, or use the sander API in Python with ParmEd:

```Python
import parmed as pmd
import sander

traj = pmd.load_file('trajectory.nc')
parm = pmd.load_file('prmtop', 'inpcrd')

inp = sander.pme_input()
inp.cut = 10 # 10 A cutoff?
# ... inp struct members match mdin variable names
with sander.setup(parm, parm.coordinates, parm.box, inp):
    for i, crd in enumerate(traj.coordinates):
        sander.set_positions(crd)
        e, f = sander.energy_forces()
        print('Frame %4d: ETOT = %12.4f' % (i, e.tot))
```

You might have to loop over NAMD the same way, though.  At least with Amber, the mdout energies do not match the positions in the mdcrd file (since the coordinates are written *after* they are propagated, so they no longer match the positions used to compute the energies).  If NAMD works the same way, you'll need to do the same with NAMD.

KarlTDebiec:
Thanks! I'll check out https://github.com/ParmEd/ParmEd/pull/307 carefully.

KarlTDebiec:
Moving to build off the [charmm branch](https://github.com/swails/ParmEd/tree/charmm) rather than master.

PR #336, opened 2015-08-09
Add docs for load_file and Structure.save

### LINKED ISSUE #332: [website] load_file and save
it might be a good idea to update parmed website to point user to use `load_file` and `save` methods since those are parmed's api.

motivation: #322

swails:
Other doc improvements.

Closes #332

### DIFF
--- a/doc/amber.rst
+++ b/doc/amber.rst
@@ -1,5 +1,5 @@
 The :mod:`parmed.amber` package
-==================================
+===============================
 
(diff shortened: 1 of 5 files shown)


hainm:
I think several example make the doc clearer

http://parmed.github.io/ParmEd/html/readwrite.html

```python
import parmed as pmd
pmd.load_file('test.pdb').save('test.cif')
```

swails:
Good idea.  Added in #337

PR #343, opened 2015-08-12
Add parsing of Amber parameter, frcmod, and leaprc files

swails:
Convert `parmed/amber/parameters.py` into a proper subclass of `ParameterSet` and support parsing most of the Amber parameter file types (still missing Amber prep files, but those are horrible).


hainm:
Great

swails:
This fixes a bug in Amber OFF file writing for when reading residue templates with only a single atom (i.e., monatomic ions).

PR #349, opened 2015-08-14
Make multiplication operator on Structure commutative

### LINKED ISSUE #286: A word of caution here – the multiplication operator for integers is not implemented for Structure
Is there any reason not to implement it?

```python
>>> 20 * phenol
  File "<stdin>", line 1, in <module>
TypeError: unsupported operand type(s) for *: 'int' and 'AmberParm'
```

swails:
It would be silly to make `struct * 10` behave different than `10 * struct` -- not to mention confusing.  So just go ahead and implement `Structure.__rmul__` as the same as `Structure.__mul__`.  Add a test for it and update docs.

Closes #286


hainm:
right, mostly copy and paste but help users (me) no need to remember 'should I use 10 * parm or parm * 10).

Issue #272, opened 2015-07-19
Renaming residues?


jchodera:
Is there an idiom that will allow me to rename residues?

The obvious idiom
```
# Rename CYS to CYX                                                                                                                                                                            
for atom in receptor[':CYS']:
    atom.residue.name = 'CYX'
```
doesn't seem to cause any permanent changes.

hainm:
@jchodera: 

you can try
```python
for res in receptor.residues:
    if res.name == 'CYS':
        res.name = 'CYX'
```

jchodera:
That did the trick---thanks!

hainm:
I think `receptor[':CYS']` returns a copy, not a view of original `receptor`.

jchodera:
Ah, I see that is clearly explained [here](http://parmed.github.io/ParmEd/html/structure.html#structure-slicing-and-selections).  I wonder if there is a way to make this behavior less surprising....?

hainm:
@jchodera  I will leave the detail answer to @swails. But I think slicing an array to return a copy is not very suprising in Python world.

You can read "Advanced Indexing" when slicing numpy array.

http://docs.scipy.org/doc/numpy/reference/arrays.indexing.html

jchodera:
That's for the pointer!  That seems to be the key here:
> Advanced indexing always returns a copy of the data (contrast with basic slicing that returns a view).

swails:
Just to expand a little here. Slicing a list always returns a shallow copy of something in Python. In numpy, it's even shallower (unlike a list, even changes to classically immutable types propagate up to the original container because it is a view).

However, the Structure is not fully analogous to a simple container, since it contains many elements, which are all interconnected (like bonds to atoms, types to bonds, angles, etc.). Couple this with the fact that each item can really only belong to a single TrackedList and a view really is not a practical thing to implement.

So as @hainm pointed out, all slicing and indexing returns the equivalent of a deep copy, unless selecting a single integer, which just returns that atom.

swails:
And thank you @hainm for fielding this question. You gave a good answer.

PR #353, opened 2015-08-15
Add `Structure.view`

### LINKED ISSUE #278: Add a `view` indexer (WAS: get atom indices?)
Jason,

I am not clear how to get atom indices from parm selection.

In `pytraj`, I can do this
```python
top = pt.load_topology("my_file.pdb")
top.select("@CA").indices # atom indices
```

if not, may be parmed can add `select` method like `mdtraj` to return indices?

swails:
This makes it so you can use the selection/slicing syntax to obtain a view of the Structure data rather than a copy.  It will let you do something like this:

```Python
struct = pmd.load_file('something.prmtop')

# The following changes all of the charges in all atoms in residues 100 to 200
# of struct to 0, since this is a view and not a copy
for atom in struct.view[100:200,:]:
    atom.charge = 0
```

Closes #278 
cc #272 -- this would work for that code snippet as well

### DIFF
--- a/doc/structure.rst
+++ b/doc/structure.rst
@@ -347,15 +347,31 @@ There are three many ways to select from a :class:`Structure
    atoms.
 
 When selecting from a :class:`Structure <parmed.structure.Structure>`
-instance, the return value can be one of three things:
+instance, the return value can be one of two things:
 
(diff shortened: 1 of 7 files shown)


swails:
The failure is a rare round-off issue.  Good to go.

PR #354, opened 2015-08-15
Fix handling of setting empty array to Structure.coordinates

### LINKED ISSUE #285: need better error message p.coordinates = []
```python
In [10]: p.coordinates = []
IndexError                                Traceback (most recent call last)
<ipython-input-10-14356ae87aff> in <module>()
----> 1 p.coordinates = []

/home/haichit/anaconda3/lib/python3.4/site-packages/parmed/structure.py in coordinates(self, value)
   1566             coords = np.array(value, dtype=np.float64, copy=False, subok=True)
   1567             coords = coords.reshape((-1, len(self.atoms), 3))
-> 1568             for a, xyz in zip(self.atoms, coords[0]):
   1569                 a.xx, a.xy, a.xz = xyz
   1570             self._coordinates = coords

IndexError: index 0 is out of bounds for axis 0 with size 0
```

```python
In [12]: p.coordinates.shape
Out[12]: (438, 3)
```

I think you need to check shape of 'values' to have better error raise.

swails:
Fixes #285

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -1604,9 +1604,18 @@ class Structure(object):
             value = list(value)
             coords = np.array(value, dtype=np.float64, copy=False, subok=True)
             coords = coords.reshape((-1, len(self.atoms), 3))
-            for a, xyz in zip(self.atoms, coords[0]):
-                a.xx, a.xy, a.xz = xyz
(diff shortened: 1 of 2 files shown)


swails:
Just a round-off failure.

PR #361, opened 2015-08-18
Support sander+scipy structure minimization in the interpreter

swails:


parmed/tools/simulations/sanderapi.py
```diff
@@ -94,3 +98,39 @@ def energy(parm, args, output=sys.stdout):
...
+
+def minimize(parm, igb, saltcon, cutoff, tol, maxcyc):
+    """ Minimizes a snapshot. Use the existing System if it exists """
+    if not HAS_SANDER:
+        raise SimulationError('Could not import sander')
+    if not HAS_SCIPY:
+        raise SimulationError('Could not import scipy')
+
+    if parm.box is None:
+        if not igb in (0, 1, 2, 5, 6, 7, 8):
```
hainm: we should really make igb=3,4 model.

swails: I think those *used* to exist, but were deprecated and removed since they sucked.

PR #369, opened 2015-08-21
Slice residue

hainm:
Slice Residue with a tuple/list

```python
residue = pmd.load_file(get_fn('aminont12.lib'))['NALA']
residue[['CA', 'CB', 'C', 'N']]
residue[[0, 3, 5, 2]]
```

feel free to update the test to your style. :D

### DIFF
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -251,7 +251,10 @@ class ResidueTemplate(object):
                 if atom.name == idx:
                     return atom
             raise IndexError('Atom %s not found in %s' % (idx, self.name))
-        return self.atoms[idx]
+        elif isinstance(idx, (list, tuple)):
+            return [self[key] for key in  idx]
+        else:
+            return self.atoms[idx]
 
     def fix_charges(self, to=None, precision=4):
         """
--- a/test/test_parmed_modeller.py
+++ b/test/test_parmed_modeller.py
@@ -1046,7 +1046,7 @@ class TestSlice(unittest.TestCase):
     '''Test slicing ResidueTemplate'''
 
     def testSliceFromArrayLike(self):
-        """ Tests slicing by a tuple/list"""
+        """ Test slicing by a tuple/list"""
         residue = pmd.load_file(get_fn('aminont12.lib'))['NALA']
 
         names = ['CA', 'CB', 'C', 'N']
--- a/test/test_parmed_modeller.py
+++ b/test/test_parmed_modeller.py
@@ -10,6 +10,7 @@ try:
 except ImportError:
     pd = None
 import os
+import parmed as pmd
 from parmed import Atom, read_PDB
 from parmed.amber import AmberParm, AmberOFFLibrary
 from parmed.exceptions import AmberWarning, Mol2Error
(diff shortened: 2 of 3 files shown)


parmed/modeller/residue.py
```diff
@@ -251,7 +251,10 @@ def __getitem__(self, idx):
                 if atom.name == idx:
                     return atom
             raise IndexError('Atom %s not found in %s' % (idx, self.name))
-        return self.atoms[idx]
+        elif isinstance(idx, (list, tuple)):
+            return [self[key] for key in  idx]
```
swails: This gets a few things wrong... First, it should be `hasattr(idx, '__iter__')`, since it could be a generator or other iterable, like numpy array.  Second, it could be a mask array: `if len(idx) == len(self) and all(isinstance(x, bool) for x in idx)`, in which case the result should be `return [self[x] for x in idx if x]`

Also, the tuple could contain multiples of the same index.  The equivalent of what `Structure.__getitem__` does here is:

```Python
    return [self[x] for x in sorted(set(idx))]
```

swails:
I had a few comments, but for the most part this looks great!

swails:
When merged, this closes #369

hainm:
@swails 

What's your motivation to use iterable object (beside list/tuple) for this? to make api consistent?

and is this correct? (from your example)

```python
if len(self) and all(isinstance(x, bool) for x in idx):
    return [self[x] for x in idx if x]
```
(what does self[True] really return? Do you meant `return [self[i] for i, x in enumerate(idx) if x]`

hainm:
and if ``idx`` is a generator and if `all(isinstance(x, bool) for x in idx)` is exhausted, how can we use it again?

swails:
Fair enough... Go ahead and ignore the mask array for now.  That's probably an unlikely use-case anyway.  If we need to add support for it, we can later.  We still want to ignore duplicates, though, so using a sorted set should keep only unique elements.

hainm:
per ` return [self[x] for x in sorted(set(idx))]`

What if user want to keep the same order? res[[1, 6, 7, 3]]?

For example, if I want to correct atom name from res0[[1, 2,3,4]] with res1[[10,2,5]]

I agree with `set`.

swails:
Once you use `set`, you lose ordering.  Sets are inherently unordered, since indexing is hash-based.  So rather than let the ordering be arbitrary (based on platform), I figured make it always sorted.

If someone really cares about what order their atoms come in, they can use a listcomp.  What do you think?

hainm:
I prefer to keep it as it is. Because that's why I want to make this PR. 

I can understand `Structure` used sorted stuff because we don't want to have duplicated (same) atoms. But in this case residue template return a list. If user ever want to get the unique atom, can he/she use ``set`` (not sure if it works with Atom class).

Additionally, what do you think about the case 
```python
res[['N', 2]]
```
set(('N', 2)) return two objects but this does not mean `res[['N', 2]]` return two different Atoms.

Solution options
* let user responsible for this (allow mixing between int and string)
* only allow either list of int or list of string

swails:
OK, good by me then.

Thanks

PR #375, opened 2015-08-22
PARMED: add charge as a property for Structure and its view

hainm:
Motivation: charge is important. Should treat it differently.

I knew we had discussion about adding more property and you did not want to make the api complicated. But charge is important (that why people make R.E.D server). So we need to make it access as easy as possible in parmed.

Look at below code to see how easy we can assign charge without writing `for ...` loop?

```python
charges = [1., 2., 0.2]
parm.view[[0, 2, 3]].charge = charges

# get charge info?
parm.charge.sum()
parm[[0, 3, 7]].charge.sum()

(p0 + p1).charge

p3.charge = (p1['@CA'] +  p['@H=']).charge
```

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -1485,6 +1485,18 @@ class Structure(object):
                 for atom in self.atoms:
                     atom.type = int(atom.atom_type)
 
+    @property
+    def charge(self):
+        '''return charge as 1D numpy array'''
+        return np.array([atom.charge for atom in self.atoms])
+
+    @charge.setter
+    def charge(self, values):
+        if len(values) != len(self.atoms):
+            raise ValueError('length of values must be equal to natom')
+        for input_charge, atom in zip(values, self.atoms):
+            atom.charge = input_charge
+
     #===================================================
 
     @property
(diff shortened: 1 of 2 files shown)


hainm:
this is my proposal, so there is still missing stuff like proper doc or charge in Residue template.

hainm:
if you're cool with this, we probably want to add other properties too.

swails:
What happens with `struct.charges[10] = 0`?  Or `struct.charges[10:20] = [0 for i in range(20)]`? These convenience methods are equivalent to one-liners:

```Python
charges = struct.charges
# replaced by
charges = [atom.charge for atom in struct.atoms]

struct.charges = charges
# replaced by
for atom, charge in zip(struct.atoms, charges): atom.charge = charge
```

And my examples still work with views (and `ResidueTemplate`s as well).  You can also take slices of an atom list:

```Python
charges = [a.charges for a in struct.atoms[::2]]
```

Coordinates are special in that various operations in ParmEd requires the numpy performance and convenience for operating on coordinates (e.g., for some center-of-mass and center-of-geometry calculations), as well as but I really don't like that modifying `struct.coordinates` doesn't modify `struct` at all.  I don't want to add *more* of those.

And I still think that `load_dataframe` will satisfy the major need here...

Issue #270, opened 2015-07-13
from_dataframe?


hainm:
should have this method too?

```python
In [9]: p = pmd.load_file("./data/Tc5b.top")
In [10]: df = p.to_dataframe()
```

like
```python
df.charge.value = new_value
df.mass.values = new_mass
new_p = pmd.Structure.from_dataframe(df)
```

Anyway, this is only an idea flash in my mind.

swails:
I hesitated making it an alternate constructor because `to_dataframe` only exports atomic properties (with some attached residue properties).  So it doesn't store any connectivity/topology information, which means you can't use just a `DataFrame` to serialize a `Structure`.

I would naively think `Structure.from_dataframe(struct.to_dataframe())` would return a copy (deep or shallow, doesn't matter) of `struct`.  But in general, it won't even come close.

Another idea is `load_dataframe`, which can override atom properties from a `DataFrame` in an existing `Structure` instance.  So you could do something like:

```Python
import pandas as pd
df = struct.to_dataframe()
# Zero-out charge on all atoms that have a charge with a magnitude greater than 0.1 electrons
df.loc[df.charge.abs() > 0.1, 'charge'] = 0
# Write those charges back to struct
struct.load_dataframe(df) # This does not exist yet.
```

But I'm going to wait for someone to want to use this in their own work that will hopefully implement it themselves :).

hainm:
Actually I meant  `struct.load_dataframe(df)` like you wrote. Just loading what `to_dataframe` exported.

`Structure.from_dataframe(df)` is meaningless (we do have `Structure`).

swails:
Let's wait on this until someone has a compelling use-case.

swails:
This is the appropriate implementation for #298

swails:
Implemented by #380

hainm:
+1

PR #380, opened 2015-08-24
Add a load_dataframe method

swails:
This commit does a couple things. Major thing it does is add a load_dataframe
instance to Structure so that atomic properties can be set via a DataFrame (it
also works with dicts through duck typing).

This commit *also* adds a to_dataframe method to StructureView and a
load_dataframe method to StructureView


swails:
Closes #270 

@hainm -- this should hopefully address the need brought up in #375, at least partially.

Issue #394, opened 2015-08-27
wrong charge when loading file from pdb


hainm:
this is weird. I got different result if running different times

scripts:
```python
#!/usr/bin/env python

import parmed as pmd
pdb_id = '3p4a'
fname = pdb_id + '_chain_AB.pdb'

pdb = pmd.download_PDB(pdb_id)[['A', 'B'],:,:]['!:MG,SR']
print(sum(atom.charge for atom in pdb.atoms))

pdb.save(fname,
        overwrite=True)

parm = pmd.load_file(fname)
print([atom.charge for atom in parm.atoms])
print(sum(atom.charge for atom in parm.atoms))
```

error
```python
haichit@casegroup ~/test $ ./prepare_pdb.3p4a.py 
0.0
[...]

PR #404, opened 2015-08-29
Make PQR file detection stricter

swails:
PQR files are too similar to PDB files to properly distinguish between all variants of both.  So make it so PQR files are *only* detected if it doesn't appear to be a valid PDB file.


swails:
Fixes #394

PR #407, opened 2015-09-01
Modified nu residue names

hainm:
taken from 

3P4{A,B,C,D} and 1EHZ.

I will update if I see new ones (and the test is not very comprehensive)

scripts:
```python
import parmed as pmd

pdb_list = ['1EHZ',]

standard = set(('A', 'U', 'C', 'G', 'HOH'))

four_letters = ('A', 'U', 'C', 'G')

for pdbidb in pdb_list:
    t = pmd.download_PDB(pdbidb)
    resset = set(res.name.rstrip() for res in t.residues) - standard
    for c in four_letters:
        print (c, [x for x in resset if c in x])
```

### DIFF
--- a/test/test_parmed_residue.py
+++ b/test/test_parmed_residue.py
@@ -139,11 +139,11 @@ class TestNucleicAcidResidues(unittest.TestCase):
                     nnuc += 1
             return nnuc
 
-        pdb_with_nnures = [('1EHZ', 76), ('3p4a', 8), ('3p4b', 8),
-                           ('3p4c', 8), ('3p4d', 8)]
-        for pdbid, nres in pdb_with_nnures:
+        pdb_with_nnures = [('1EHZ', 76), ]
+        for pdbid, correct_nres in pdb_with_nnures:
             pdb = pmd.download_PDB(pdbid)
-            self.assertTrue(count_nures(pdb), nres)
+            nres = count_nures(pdb)
+            self.assertTrue(correct_nres == nres)
 
 if __name__ == '__main__':
     unittest.main()
--- a/test/test_parmed_residue.py
+++ b/test/test_parmed_residue.py
@@ -3,7 +3,9 @@ Tests the functionality in parmed.residue
 """
 
 import utils
+import parmed as pmd
 from parmed import residue
+from parmed.residue import RNAResidue, DNAResidue
 import unittest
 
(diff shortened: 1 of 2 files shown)


test/test_parmed_residue.py
```diff
@@ -128,5 +130,20 @@ def testTermini(self):
...
+            for res in parm.residues:
+                if RNAResidue.has(res.name) or DNAResidue.has(res.name):
+                    nnuc += 1
+            return nnuc
+
+        pdb_with_nnures = [('1EHZ', 76), ]
+        for pdbid, correct_nres in pdb_with_nnures:
+            pdb = pmd.download_PDB(pdbid)
+            nres = count_nures(pdb)
+            self.assertTrue(correct_nres == nres)
```
swails: Change this to:

```Python
self.assertEqual(correct_nres, nres)
```

It improves the failure message if it fails.  So instead of getting `AssertionError: False is not True`, you get `AssertionError: repr(correct_nres) is not equal to repr(nres)`.

hainm: I see.

hainm:
@swails 

any idea how to fix this error on travis? (happen with py3.4)

```OSError: Could not open https://github.com/ParmEd/ParmEd/raw/master/test/files/ala_ala_ala.psf: HTTP Error 429: Too Many Requests```

swails:
I'm rerunning the test.  If this becomes a persistent issue, I may have to rethink how automatic file type detection is done for websites.

hainm:
do you have any tip to turn off `Optimized readparm` when installing. It's a bit slow to compile while I only need to test a small change. I can turn off it in setup.py but later I need to ``checkout`` this file.

swails:
You can hack setup.py if you want.  But I made it so you can run the tests directly from the git repo directory so you don't need to "install" to run the tests.  That's what I usually do.

hainm:
> But I made it so you can run the tests directly from the git repo directory so you don't need to "install" to run the tests.

cool. this is what I need.

swails:
Looks good, thanks!

hainm:
> do you have any tip to turn off Optimized readparm when installing. It's a bit slow to compile while I only need to test a small change. I can turn off it in setup.py but later I need to checkout this file.

I just realize that we can use 'pip install -e .' for developer mode: so we just change the source code and use it (from any dir) without re-install).

PR #408, opened 2015-09-01
 add residue set and residue count for summary

hainm:
the summary's output in this PR is not very pretty for Structure having lots of restypes. But's ok for DNA/RNA. :D

### DIFF
--- a/parmed/tools/actions.py
+++ b/parmed/tools/actions.py
@@ -2963,10 +2963,9 @@ class summary(Action):
         _rcount = ','.join((sorted(_rcount.split(',')))) + '\n'
 
         residue_set = _reformat_long_sentence(_rset, 'Residue set:           ',
-                offset=None,
-                n_words=7)
+                                              offset=None, n_words=7)
         residue_count = _reformat_long_sentence(_rcount, 'Residue count:         ',
-                offset=-1, n_words=7)
+                                                offset=-1, n_words=7)
         retval += residue_set + residue_count
 
         if self.parm.box is not None and set(self.parm.box[3:]) == set([90]):
--- a/parmed/tools/actions.py
+++ b/parmed/tools/actions.py
@@ -2927,8 +2927,8 @@ class summary(Action):
                    len(self.parm.atoms), len(self.parm.residues))
         )
 
-        _rset = set(res.name for res in self.parm.residues).__str__() + '\n'
-        _rcount = Counter(res.name for res in self.parm.residues).__str__() + '\n'
+        _rset = str(set(res.name for res in self.parm.residues)) +  '\n'
+        _rcount = str(Counter(res.name for res in self.parm.residues)) +  '\n'
         _rcount = _rcount.replace('Counter({', '')
         _rcount = _rcount.replace('})', '')
         residue_set = (
--- a/parmed/tools/actions.py
+++ b/parmed/tools/actions.py
@@ -2928,9 +2928,14 @@ class summary(Action):
         )
 
         _rset = set(res.name for res in self.parm.residues).__str__() + '\n'
+        _rcount = Counter(res.name for res in self.parm.residues).__str__() + '\n'
+        _rcount = _rcount.replace('Counter({', '')
+        _rcount = _rcount.replace('})', '')
         residue_set = (
                  'Residue set:           ' + _rset)
-        retval += residue_set
+        residue_count = (
+                 'Residue count:         ' + _rcount)
+        retval += residue_set + residue_count
 
         if self.parm.box is not None and set(self.parm.box[3:]) == set([90]):
             a, b, c = self.parm.box[:3]
--- a/parmed/tools/actions.py
+++ b/parmed/tools/actions.py
@@ -2927,10 +2927,11 @@ class summary(Action):
                    len(self.parm.atoms), len(self.parm.residues))
         )
 
-        _rset = str(set(res.name for res in self.parm.residues)) +  '\n'
-        _rcount = str(Counter(res.name for res in self.parm.residues)) +  '\n'
-        _rcount = _rcount.replace('Counter({', '')
-        _rcount = _rcount.replace('})', '')
+        _rset = str(sorted(set(res.name for res in self.parm.residues))) +  '\n'
+        _rcount = str(Counter(res.name for res in self.parm.residues))
+        _rcount = _rcount.replace('Counter({', '').replace('})', '')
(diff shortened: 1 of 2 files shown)


parmed/tools/actions.py
```diff
@@ -2927,6 +2927,16 @@ def __repr__(self):
                    len(self.parm.atoms), len(self.parm.residues))
         )
 
+        _rset = set(res.name for res in self.parm.residues).__str__() + '\n'
+        _rcount = Counter(res.name for res in self.parm.residues).__str__() + '\n'
```
swails: why not `str(<blah>)` instead of `<blah>.__str__()`?  The former looks a bit more natural.

hainm: I will change to ``str( )``. I have habit to use ``.method`` to avoid using to many ``( )``, and for auto completion.

hainm:
2 passed + 1 failed (http issue).

test/saved_outputs.py
```diff
@@ -366,6 +368,8 @@
 Total mass (amu):      339672.2974
 Number of atoms:       56057
 Number of residues:    18015
+Residue set:           ['ALA', 'ARG', 'ASN', 'ASP', 'CYS', 'GLN', 'GLU', 'GLY', 'HSD', 'ILE', 'LEU', 'LYS', 'MET', 'PHE', 'PRO', 'SER', 'THR', 'TIP3', 'TRP', 'TYR', 'VAL']
+Residue count:          'ALA': 13, 'ARG': 9, 'ASN': 5, 'ASP': 14, 'CYS': 2, 'GLN': 4, 'GLU': 12, 'GLY': 10, 'HSD': 5, 'ILE': 12, 'LEU': 11, 'LYS': 6, 'MET': 5, 'PHE': 6, 'PRO': 10, 'SER': 9, 'THR': 6, 'TRP': 5, 'TYR': 4, 'VAL': 11,'TIP3': 17856
 System volume (ang^3): 615109.34
```
swails: Is there any way the format here can be improved with linebreaks?

swails:
The HTTP failures here probably arise from the repeated attempts to download the same file from the internet (it attempts to open it each time it tries a new format, then again when it has found the correct format).  From Github's perspective, this probably looks like a DOS attack, so they block it.

I'll address this soon by simply cutting down the number of attempted downloads somehow.

test/saved_outputs.py
hainm: yes. We can count the length of string and and split it (if > 80?).

I will submit another PR (soon).

swails: Why not just add it to this PR?

swails: And 80 characters sounds good (it would be nice if the continuation picked up at the same indentation level as where the residue list started on the previous line).  So something like:

```
Residue set:      ['ALA', 'ARG', '...'
                   'HSD', 'ILE', '...']
```

hainm: > And 80 characters sounds good (it would be nice if the continuation picked up at the same indentation level as where the residue list started on the previous line). So something like:

yes, this is what I am going to do

> Why not just add it to this PR?

English. should be 'commit'.

hainm:
ok. So I just pushed new commits. The best I can do right now.

ParmEd loads and processes files a bit slow so I am losing patient to try more (and prettier format).

For example, below code (adapted from ParmEd test) took > 1/2 minutes.

```python
"""
Tests for the various actions in ParmEd
"""
from __future__ import division, print_function

import utils
from parmed import periodic_table, gromacs, load_file
from parmed.amber import AmberParm, ChamberParm, AmoebaParm
from parmed.charmm import CharmmPsfFile
from parmed.exceptions import AmberWarning, CharmmWarning
from parmed.formats import PDBFile, CIFFile
from parmed.utils.six.moves import range, zip
from copy import copy
import os
import parmed.tools as PT
from parmed.tools import exceptions as exc
from parmed.tools import parmlist
import saved_outputs as saved
import sys
import unittest
import warnings

get_fn = utils.get_fn
get_saved_fn = utils.get_saved_fn
diff_files = utils.diff_files

gasparm = AmberParm(get_fn('trx.prmtop'))
solvparm = AmberParm(get_fn('solv.prmtop'))
gascham = ChamberParm(get_fn('ala_ala_ala.parm7'))
solvchamber = ChamberParm(get_fn('dhfr_cmap_pbc.parm7'))
amoebaparm = AmoebaParm(get_fn('nma.parm7'))

class Test(unittest.TestCase):
    def testSummary0(self):
        self.maxDiff = None
        """ Test summary action on AmberParm """
        parm = AmberParm(get_fn('things.parm7'), get_fn('things.rst7'))
        act = PT.summary(parm)
        print("SUMMARY")
        print(act)
        self.assertEqual(str(act), saved.SUMMARY)

    def testSummary1(self):
        self.maxDiff = None
        """ Test summary action for ChamberParm """
        parm = copy(solvchamber)
        parm.load_rst7(get_fn('dhfr_cmap_pbc.rst7'))
        act = PT.summary(parm)
        print("SUMMARYC1")
        print(act)
        print(saved.SUMMARYC1)
        self.assertTrue(utils.detailed_diff(str(act), saved.SUMMARYC1,
                                            relative_error=1e-6))
    def testSummary2(self):
        """ Test summary action for AmoebaParm """
        self.maxDiff = None
        parm = copy(amoebaparm)
        act = PT.summary(parm)
        print('SUMMARYA1')
        print(act)
        self.assertEqual(str(act), saved.SUMMARYA1)
        PT.loadRestrt(parm, get_fn('nma.rst7')).execute()
        act = PT.summary(parm)
        print("SUMMARYA2")
        print(act)
        self.assertEqual(str(act), saved.SUMMARYA2)

    def testSummary3(self):
        """ Tests the use of a PDB file with the summary action """
        self.maxDiff = None
        parm = load_file(get_fn('4lzt.pdb'))
        act = str(PT.summary(parm))
        print("PDB_SUMMARY")
        print(act)
        self.assertEqual(str(PT.summary(parm)), saved.PDB_SUMMARY)

if __name__ == '__main__':
    unittest.main()
```

hainm:
anyway, I am using lab cluster, so ...

swails:
The issues here look unrelated to the problems in Travis.  I'm guessing those came from OpenMM.  I'll look into it tomorrow.

swails:
This looks nice! Thanks!

swails:
Looks good.  I will merge this after #409, since that one contains both the OpenMM fix that's causing test failures here and the HTML blocking.

Thanks!

PR #409, opened 2015-09-02
Fix writing PDB and PQR files with multiple models

swails:
Old (bad) behavior was to write all models with the same coordinates.

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -722,7 +722,7 @@ class PDBFile(object):
                 else:
                     atoms = sorted(res.atoms, key=lambda atom: atom.number)
                 for atom in atoms:
-                    pa, others, (x, y, z) = print_atoms(atom, coords[0])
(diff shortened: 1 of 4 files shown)


parmed/formats/pdb.py
```diff
@@ -722,7 +722,7 @@ def print_atoms(atom, coords):
                 else:
                     atoms = sorted(res.atoms, key=lambda atom: atom.number)
                 for atom in atoms:
-                    pa, others, (x, y, z) = print_atoms(atom, coords[0])
+                    pa, others, (x, y, z) = print_atoms(atom, coord)
```
hainm: using 'print_atoms' seems not a good function name. :D I always think ``print`` is just for stdout.

swails: `print_atoms` is an internal function defined and used within a single scope.  When I came up with the name `print_atoms`, `print` was an adjective to me (it returns the printed atom), not a verb.  Its name matters little.

PR #410, opened 2015-09-03
New feature added to add12_6_4.py file.

pengfeili1:
Could better handle different situations. Adjust the code to easier to read.

### DIFF
--- a/parmed/tools/add1264.py
+++ b/parmed/tools/add1264.py
@@ -78,88 +78,87 @@ def params1264(parm, mask, c4file, watermodel, polfile, tunfactor):
 
     print("***********************************************************")
     # Determine which atom type was treated as the center metal ion
-    typelist = dict()
+    mettypdict = dict()
     for i in mask.Selected():
-        typeidx = parm.parm_data['ATOM_TYPE_INDEX'][i]
-        chgi = parm.parm_data['CHARGE'][i]
-        if typeidx in typelist: continue
-        typelist[typeidx] = (parm.atoms[i].atomic_number, int(chgi))
-        print("The selected metal ion is",
-              pt.Element[parm.atoms[i].atomic_number])
-    types = typelist.keys()
-    types.sort()
+        mettypind = parm.parm_data['ATOM_TYPE_INDEX'][i]
+        metchg = parm.parm_data['CHARGE'][i]
+        if mettypind in mettypdict: continue
+        mettypdict[mettypind] = (parm.atoms[i].atomic_number, int(metchg))
+        print "The selected metal ion is", \
+                pt.Element[parm.atoms[i].atomic_number]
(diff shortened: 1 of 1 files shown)


pengfeili1:
Hi Jason,

I have modified the code with renaming some variables, making them easier to read. Modifications have also been made to rise an error when there are two atom types have same VDW parameters but different polarizability parameters.

All the best,
Pengfei

hainm:
It is likely you need to add a test for raising.

parmed/tools/add1264.py
```diff
@@ -78,89 +78,88 @@ def params1264(parm, mask, c4file, watermodel, polfile, tunfactor):
...
-    # Determine the C4 term between the atom type of center metal ion 
-    # and every atom type in the prmtop file 
-
-    # 1. Get the list of AMBER_ATOM_TYPE and ATOM_TYPE_INDEX for all 
+        mettypind = parm.parm_data['ATOM_TYPE_INDEX'][i]
+        metchg = parm.parm_data['CHARGE'][i]
+        if mettypind in mettypdict: continue
+        mettypdict[mettypind] = (parm.atoms[i].atomic_number, int(metchg))
+        print "The selected metal ion is", \
+                pt.Element[parm.atoms[i].atomic_number]
```
swails: This needs to be changed to:

```Python
        print("The selected metal ion is",
              pt.Element[parm.atoms[i].atomic_number])
```

We support both Python 2 and Python 3 in the same code base, and this won't work with Python 3.  You can put `from __future__ import print_function` at the top to get the right behavior.

parmed/tools/add1264.py
```diff
@@ -78,89 +78,88 @@ def params1264(parm, mask, c4file, watermodel, polfile, tunfactor):
...
+    mettypinds = mettypdict.keys()
+    mettypinds.sort()
```
swails: This doesn't work for Python 3, either.  Unlike Python 2, Python 3 does not return a list with `dict.keys()` -- it returns a more efficient `dict_keys` iterator that consumes less memory.

Change these two lines into a single line:

```Python
    mettypinds = sorted(mettypdict.keys())
```

swails:
Ack!  Sorry -- I added tests for `add12_6_4`, but I didn't give them the right name so they're not run by default.  I've fixed that in #411, but what that will show is that this PR fails those tests (at least for Python 3).

I'll pull this PR into that PR and fix up the issues.  As @hainm said, if you have a test case for the failing test, then by all means go ahead and add it, but if not that's fine, too.

Thanks!

swails:
OK.  @pengfeili1 -- your commit is included in the #411 PR, so I'm going to go ahead and close this.  Let me know if it doesn't work for you for whatever reason.

I'll also go ahead and update the version in the Amber git repo with these changes.

PR #411, opened 2015-09-03
Fix up add12_6_4

swails:
This PR makes sure that the tests are run as part of the test suite.


swails:
This includes the commit in PR #410.

Closes #410

PR #412, opened 2015-09-03
Make StructureView iteration behave like Structure iteration

swails:
It iterates over atoms in the StructureView


hainm:
just want to give an example

```python
        sel1 = parm['@1']
        sel2 = parm[:1]
        self.assertIsInstance(sel1, pmd.Structure)
        self.assertIsInstance(sel2, pmd.Structure)
```

If using type checking in julia, don't need to write test ``IsInstance``

(I know that parm[0] return an Atom too).

swails:
> If using type checking in julia, don't need to write test IsInstance

Can you write a function that returns different types depending on the input type?  I don't see how you can get around type checking in a test if your return value can be one of two different types...

This is necessary because `Structure[<something>]` can return different types of objects, depending on what `<something>` is.  You *somehow* have to test that it's doing the right thing for each kind of `<something>`, whether you're using Julia or Python.

Either Julia doesn't allow me to do what I did in Python, or it does and you need the same tests.

PR #427, opened 2015-09-15
[WIP] Define a protocol for dealing with electrostatic constant

### LINKED ISSUE #426: Improve electrostatic constant treatment in ParmEd
Every program uses a *slightly* different value of the electrostatic force constant that it uses.  Currently, the charge is read in unmodified from the original file where they are provided as elementary charge units, or scaled according to the Amber electrostatic constant in the case of Amber topology files.

However, when converting from Gromacs to AMBER, or AMBER to OpenMM, or AMBER to Gromacs, the charges are not scaled by the ratios of the scaling constants used in each program, which means that each program uses a *slightly* different set of charges (same net charge for a neutral system).

The way to fix this is to give ParmEd its *own* electrostatic constant (that I'll probably take directly from NIST or the CRC) that it stores charges in internally, and then have the charges get scaled appropriately according to the electrostatic constant of the parent program for the output file format.  This should ensure that the charge representations inside each program is exactly the same following conversion (subject to the limitations of floating point precision when dividing large-ish numbers that are nearly equal).

swails:
To improve the agreement of interconverted charges, ParmEd now stores charges
internally scaled to its own definition of the electrostatic constant
(calculated from values in the CRC with the highest possible precision). Input
charges read from the different sources are scaled by the ratio of the constant
used in the parent program to the ParmEd internal constant, and are scaled
*back* upon writing.

This behavior, and the electrostatic constants used in each program supported by
ParmEd, are added to the documentation in a fairly detailed section.

Fixes #426


swails:
This is the wrong way to do it.  It corrupts the charges in the internally stored state.  This can wait until we have a decision made about what to do with Amber.

PR #430, opened 2015-09-21
Fix building on Macs with Anaconda or Miniconda and MacPorts GCC activated

swails:


setup.py
```diff
@@ -7,6 +7,14 @@
                      'correctly.\n')
     sys.exit(0)
 
+if sys.platform == 'darwin':
+    # You *need* to use clang and clang++ for ParmEd extensions on a Mac;
+    # Anaconda does annoying stuff that breaks this, since their distutils
+    # automatically tries to use "gcc", which would conflict with the MacPorts
+    # gcc... sigh.
+    os.environ['CXX'] = 'clang++'
```
hainm: is clang++ always available on Mac?

have you tried using Cython to call C/C++ function in python? It seems to me that cython supports different compilers well. (but I am not sure about ParmEd's case).

swails: > is clang++ always available on Mac?

Yes, as long as you have Xcode installed.

> have you tried using Cython to call C/C++ function in python? It seems to me that cython supports different compilers well. (but I am not sure about ParmEd's case; I should buy a Mac to try :D).

This doesn't dodge the problem at all.  Cython creates C++/C files that distutils still needs to compile.  And it will try to compile with `gcc` (or `g++`) when you use Anaconda or Miniconda Python.  If you have MacPorts gcc activated, it pulls gcc from MacPorts which doesn't have the special Apple-specific flags.

That said; no, I don't use Cython here.  The interface to C is simple enough that it's unnecessary (and I already have the pure C-API working fine).

hainm:
I see.

PR #439, opened 2015-09-25
Move segid to Residue

### LINKED ISSUE #438: Put segid on Residue instead of Atom
`segid` (or segment identifier) is really what CHARMM likes to call chains.  This should be a property on `Residue` rather than `Atom`.

swails:
Fixes #438


swails:
This is only erroring because the q4md website is down and the file downloads from that site are failing.  This is good to merge.

PR #440, opened 2015-09-29
Add XMLFile class

swails:
This object will automatically detect an OpenMM XML file and then deserialize it
into the original object. Works for State, System, Integrator, and ForceField.


swails:
Only failure is caused by R.E.D.'s website being down.

PR #442, opened 2015-09-30
Implements moleculetype combining semantics for GromacsTopologyFile

### LINKED ISSUE #435: GROMACS writing: Any way to merge molecules?
In some circumstances, I may want a GROMACS topology file containing multiple molecules to be written to a single [ molecule ] definition. Is there a way to get ParmEd to do this rather than splitting different molecules? For example, if I have a protein + ligand and I want to apply certain restraints between the protein and ligand, I might want to load the two into ParmEd separately but write them out as a single "molecule" (i.e. with the protein and ligand atoms appearing consecutively in the same [ atoms ] section). 

Thanks.

swails:
`GromacsTopologyFile.write(combine)` now accepts a list of lists, where the second level of list specifies molecule indices to combine into a single moleculetype.  These indices *must* be consecutive, since otherwise it will desynchronize from the coordinate file.

Closes #435

### DIFF
--- a/devtools/travis-ci/pyflakes_check.sh
+++ b/devtools/travis-ci/pyflakes_check.sh
@@ -17,7 +17,7 @@ notfound() {
 
 test -d parmed || notfound
 
-pyflakes parmed | \
+pyflakes parmed 2>&1 | \
     grep -v -e "from parmed.topologyobjects import \*" \
             -e "from parmed.modeller.residue import \*" \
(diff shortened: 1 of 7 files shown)


swails:
This PR also adds a test case (implemented first, as per TDD) and documentation.

hainm:
can you update website too?

swails:
I needed to fix the raw text blocks (which I did in 8939176).  Website is updated.

Issue #445, opened 2015-10-01
Merging molecules in gromacs topology sets atom masses to zero in the output  [ atomtypes ] section


nividic:
Hi there,

I was trying to merge molecules in the gromacs topology file by using:

from parmed import gromacs
gmx_top = gromacs.GromacsTopologyFile('complex.top')
gmx_top.write('test.top',[[0,1]])

However, I noted that if in the original topology file are present values for the atom masses in the [ atomtypes ] section, these are reported as zeros in the [ atomtypes ] section related to the output file i.e.

Input Topology:

[ atomtypes ]
; name at.num mass charge ptype sigma espilon
HC 1 0.00000 1.008000 A 0.264953 0.0656888
CT 6 0.00000 12.010000 A 0.339967 0.45773
C 6 0.00000 12.010000 A 0.339967 0.359824
....

Output Toplogy:

[ atomtypes ]
; name at.num mass charge ptype sigma epsilon
HC 1 0.00000 0.000000 A 0.264953 0.0656888
CT 6 0.00000 0.000000 A 0.339967 0.45773
C 6 0.00000 0.000000 A 0.339967 0.359824
.......

In my opinion it would be better to keep the original values as well. I have attached the link related to the topology file: 

https://drive.google.com/file/d/0B4Hyb-RxLkeGcVIxMjJrOVNscUE/view?usp=sharing

Thanks,

Gaetano

swails:
I found and fixed the issue.  The charges are still printed with a value of 0 since ParmEd doesn't currently store charge information in atom types.

See #448

swails:
This is fixed.

PR #448, opened 2015-10-06
Fix printout of masses in generated GROMACS top files

swails:
Accidentally swapped charge and mass. It was mainly cosmetic, as these masses
were not used by anything that I'm aware of, but it's still better to print out
compatible topologies.


swails:
Fixes #445

PR #451, opened 2015-10-07
Fix misassignment of `defaults` section for GromacsTopologyFile

### LINKED ISSUE #450: Splitting and re-merging a GROMACS topology results in incorrect FudgeLJ and FudgeQQ
FudgeLJ and FudgeQQ in GROMACS topologies are being re-set to incorrect (default?) values upon splitting and re-merging a GROMACS topology file. For example:

'''
system = parmed.load_file('159.top')
system.write('output1.top')

components = system.split()
outputsys = components[0][0] * len(components[0][1])
outputsys.write('output2.top')
'''

In this case, `output1.top` has the correct values (0.5 and 0.833 for fudgeLJ and fudgeQQ, respectively) but in output2.top, they are incorrect (1 and 1, respectively).

Input topology file for the above example: https://dl.dropboxusercontent.com/u/3409095/159.top

We've had to kill a number of things we were running because they apparently had incorrect scaling factors as a result of this issue.

@bannanc, this is the issue you mentioned.

swails:
Neither the `__copy__` nor the `__getitem__` functions for `GromacsTopologyFile` were implemented for that object (i.e., it was just calling the underlying function from the inherited `Structure`).  Since `Structure` stores the 1-4 scaling constants either in the dihedrals themselves (in the case of Amber), or in the explicit list of exceptions, ParmEd would never set the `defaults` attribute for the copy (meaning it would just get the default when copying a GromacsTopologyFile).

This broke the `GromacsTopologyFile` copy constructor when you just wanted to print out a new Gromacs topology file (and also broke the selector [], which is used by split under-the-hood).  The reason I didn't catch this before is that the underlying `Structure` still had the correct information -- just in the exclusion list.  So if you converted to another format (e.g., Amber), or tried to evaluate an energy with OpenMM, you would get the correct answer.  It was only when going *back* to `GromacsTopologyFile` that this didn't work.

Fixes #450

### DIFF
--- a/parmed/__init__.py
+++ b/parmed/__init__.py
@@ -3,7 +3,7 @@ The parmed package manipulates molecular structures and provides a way to go
 between standard and amber file formats, manipulate structures, etc.
 """
 
-__version__ = '2.0.3'
+__version__ = '2.0.4'
 __author__ = 'Jason Swails'
 
 __all__ = ['exceptions', 'periodic_table', 'residue', 'unit', 'utils',
--- a/parmed/__init__.py
+++ b/parmed/__init__.py
@@ -6,7 +6,7 @@ between standard and amber file formats, manipulate structures, etc.
 # Version format should be "major.minor.patch". For beta releases, attach
 # "-beta#" to the end. The beta number will be turned into another number in the
 # version tuple
-__version__ = '2.0.4-beta3'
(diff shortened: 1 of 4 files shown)


davidlmobley:
@bannanc - this is your issue. 

Thanks, Jason.

PR #467, opened 2015-10-24
Added tests to improve test coverage

swails:
Fixed a number of small bugs in the process

- Fix reading and processing of Amber topology files with added PDB data
- Fix detection of unsupported parameter types by Amber
- Fix parsing of old-style topology files
- Fix setMolecules solute_ions keyword
- Fix L-J parameter recalculation when negative indices present (rarely, if ever, used routine, but it is now tested for correctness)
- Fix 10-12 support in AmberParm with OpenMM integration. It had previously (erroneously) calculated both 12-6 *and* 12-10 interactions for any type pairs that had 12-10 parameters (by definition, only one is supposed to be calculated).  This was only a small difference in energy, but it now matches between sander and OpenMM much better.
- Fix prmtop reading with additional exclusions
- Fix deserialization of `Residue` (the `residue` attribute on each atom was not being set at all).
- Fix deserialization of `acceptor` and `donor` attributes in `Structure`
- Add `from_rawdata` method to `BeemanRestart` file class

Now achieved 99% coverage in `parmed/__init__.py`, 94% coverage in `parmed/amber/_amberparm.py`, 95% coverage in `parmed/amber/_chamberparm.py`, and 99% coverage in `parmed/amber/_tinkerparm.py`.


swails:
Adds mask syntax `@/El` to select all atoms with the element name `El` (I think this was the original intent given placeholder function names).  Also adds a couple improved error checks.

Version bump for the mask parser addition, and merge when Travis passes.

hainm:
what's the different between `@H=` and `@/H`?

swails:
> what's the different between @H= and @/H?

One selects based on name-matching with wild-card replacement and the other specifically picks out matching elements, *and nothing else*.

This is easier to see when you're trying to pick out carbon atoms: `@C=` picks out any atom whose name starts with C.  This is usually sufficient to get all carbon atoms, but you'll also pick up a lot of ions like chloride, calcium, cesium... anything that starts with C.

The element selection is cleaner if all you want to do is select a particular element.

Another example (though weaker), is if someone had named an atom `D` because it's deuterium (this is not typically how that would be done).  `@/H` gets that atom, `@H=` does not.

For hydrogen, specifically, there is little difference.  For carbon (and others, like Nitrogen and Sulfur that have the same problem to a lesser extent), it can have a big difference.

hainm:
ok. I see. How do you think '@/` to '@_'?

swails:
> ok. I see. How do you think '@/` to '@_'?

I didn't invent the `/` syntax -- it was already in the mask parser that I copied from ptraj.  That leads me to suspect it has its roots in Chimera, but I don't know for certain.  We're also more likely to use `_` to represent atom names than `/`, probably.

PR #479, opened 2015-11-16
read_PDB from text (lines of strings)

hainm:
what do you think? 

(I have this idea when playing with pandas.read_clipboard). 

Disclaim: I am not sure if this is very helpful, but fun to implement with few lines.


swails:
I think I'm -1 here for two main reasons.

* The input being a `str` is already applied to file names.  By overloading it here, you introduce the need to "guess" what the user input is.  Discriminating between file name and file contents based on whether the file exists or not is very harsh to typos:

```Python
In [1]: import parmed as pmd

In [2]: pmd.read_PDB('does_not_exist.pdb')
ValueError                                Traceback (most recent call last)
<ipython-input-2-1a2da0a92475> in <module>()
----> 1 pmd.read_PDB('does_not_exist.pdb')

/Users/swails/src/ParmEd/parmed/formats/pdb.py in parse(filename)
    611             all_coordinates.append(coordinates)
    612         struct._coordinates = np.array(all_coordinates).reshape(
--> 613                         (-1, len(struct.atoms), 3))
    614         return struct
    615 

ValueError: total size of new array must be unchanged

```
* You can accomplish exactly the same thing via `StringIO` (since `pmd.read_PDB` takes file-like in addition to file name).  Sure, it's a little extra typing, but it's not too bad:

```Python
import io

fobj = StringIO()
fobj.write(STRING_CONTENTS)
fobj.seek(0)
pmd.read_PDB(fobj)
```
Instead of
```Python
pmd.read_PDB(STRING_CONTENTS)
```
Sure it's an extra 3 lines of code, but it's really not much.

That's the general way to tackle this kind of thing with ParmEd (since file-like objects are taken in many places).  In cases where "file-like" only needs to iterate over lines in a file (like with the PDB parser), you can do this even *more* easily:

```Python
pmd.read_PDB(STRING_CONTENTS.split('\n'))
```

Since a list of strings is file-like *enough* for this routine (at least for now...).

swails:
Compare that to the current treatment of typos:

```Python
In [1]: import parmed as pmd

In [2]: pmd.read_PDB('does_not_exist.pdb')
FileNotFoundError                         Traceback (most recent call last)
<ipython-input-2-1a2da0a92475> in <module>()
----> 1 pmd.read_PDB('does_not_exist.pdb')

/Users/swails/src/ParmEd/parmed/formats/pdb.py in parse(filename)
    275         if isinstance(filename, string_types):
    276             own_handle = True
--> 277             fileobj = genopen(filename, 'r')
    278         else:
    279             own_handle = False

/Users/swails/src/ParmEd/parmed/utils/io.py in genopen(name, mode)
    143             return TextIOWrapper(open_url)
    144     else:
--> 145         return open(name, mode)

FileNotFoundError: [Errno 2] No such file or directory: 'does_not_exist.pdb'

```

This is a much clearer error message (the first one I was actually surprised resulted in an error at all -- I was really expecting just an empty `Structure`).

hainm:
ha ha. okie. I forgot about this case

`pmd.read_PDB('does_not_exist.pdb')`

PR #481, opened 2015-11-19
Fix atomic number assignment for Tinker XYZ files.

swails:


swails:
Travis got messed up in one of the 5 builds.

PR #489, opened 2015-12-03
Do not delete box when solvent removed by strip in AmberParm

swails:
This causes problems now that the parm is always rebuilt before writing.
Sometimes we want unit cell vectors even if we don't have solvent...


swails:
Failure is round-off.

hainm:
> Failure is round-off.

so ignore it?

swails:
yes -- it will get rerun upon merge.  If it fails there, I will rerun the test by hand.

PR #502, opened 2015-12-11
Begin working on XML parameter writing

swails:
This PR adds writing for atom types, residue templates, bond, and angle parameters.

Pinging @jchodera, @chayast, and @rafwiewiora since this should be useful in the Amber/CHARMM conversions.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -12,10 +12,11 @@ from parmed.formats.registry import FileFormatType
 from parmed.modeller.residue import ResidueTemplate
 from parmed.parameters import ParameterSet
 from parmed.periodic_table import Element, Mass
-from parmed.topologyobjects import NoUreyBradley
+#from parmed.topologyobjects import NoUreyBradley
 from parmed import unit as u
 from parmed.utils.io import genopen
 from parmed.utils.six import add_metaclass, string_types, iteritems
+from parmed.utils.six.moves import range
 
 @add_metaclass(FileFormatType)
 class OpenMMParameterSet(ParameterSet):
@@ -146,11 +147,10 @@ class OpenMMParameterSet(ParameterSet):
             self._write_omm_residues(dest)
             self._write_omm_bonds(dest)
             self._write_omm_angles(dest)
-#           self._write_omm_dihedrals(dest)
-#           self._write_omm_periodic_impropers(dest)
-#           self._write_omm_impropers(dest)
+            self._write_omm_dihedrals(dest)
+            self._write_omm_impropers(dest)
 #           self._write_omm_rb_torsions(dest)
-#           self._write_omm_cmaps(dest)
+            self._write_omm_cmaps(dest)
 #           self._write_omm_scripts(dest)
 #           self._write_omm_nonbonded(dest)
         finally:
@@ -168,9 +168,7 @@ class OpenMMParameterSet(ParameterSet):
 
     def _write_omm_atom_types(self, dest):
         dest.write(' <AtomTypes>\n')
-        ljtypes = dict()
-        counter = 0
-        for name, atom_type in iteritems(self.atom_types)
+        for name, atom_type in iteritems(self.atom_types):
             assert atom_type.atomic_number >= 0, 'Atomic number not set!'
             element = Element[atom_type.atomic_number]
             dest.write('  <Type name="%s" class="%d" element="%s" mass="%f"/>\n'
@@ -230,3 +228,90 @@ class OpenMMParameterSet(ParameterSet):
         dest.write(' </HarmonicAngleForce>\n')
 
     def _write_omm_dihedrals(self, dest):
+        if not self.dihedral_types: return
+        # In ParameterSet, dihedral_types is *always* of type DihedralTypeList.
+        # The from_structure method ensures that, even if the containing
+        # Structure has separate dihedral entries for each torsion
+        dest.write(' <PeriodicTorsionForce>\n')
+        diheds_done = set()
+        pconv = u.degree.conversion_factor_to(u.radians)
+        kconv = u.kilocalorie.conversion_factor_to(u.kilojoule)
+        def nowild(name):
+            return name if name != 'X' else ''
+        for (a1, a2, a3, a4), dihed in iteritems(self.dihedral_types):
+            if (a1, a2, a3, a4) in diheds_done: continue
+            diheds_done.add((a1, a2, a3, a4))
+            diheds_done.add((a4, a3, a2, a1))
+            dest.write('  <Proper type1="%s" type2="%s" type3="%s" type4="%s"'
+                       % (nowild(a1), a2, a3, nowild(a4)))
+            for i, term in enumerate(dihed):
+                i += 1
+                dest.write(' periodicity%d="%d" phase%d="%f" k%d="%f"' %
+                           (i, term.per, i, term.phase*pconv, i,
+                            term.phi_k*kconv))
+            dest.write('/>\n')
+        # Now do the periodic impropers. OpenMM expects the central atom to be
+        # listed first. ParameterSet goes out of its way to list it third
+        # (consistent with Amber) except in instances where order is random (as
+        # in CHARMM parameter files). But CHARMM parameter files don't have
+        # periodic impropers, so we don't have to worry about that here.
+        for (a2, a3, a1, a4), improp in iteritems(self.improper_periodic_types):
+            # Try to make the wild-cards in the middle
+            if a4 == 'X':
+                if a3 != 'X':
+                    a3, a4 = a4, a3
+                elif a2 != 'X':
+                    a2, a4 = a4, a2
+            dest.write('  <Improper type1="%s" type2="%s" type3="%s" '
+                       'type4="%s" periodicity1="%d" phase1="%f" k1="%f"/>\n' %
+                       (a1, nowild(a2), nowild(a3), nowild(a4), improp.per,
+                        improp.phase*pconv, improp.phi_k*kconv)
+            )
+        dest.write(' </PeriodicTorsionForce>\n')
+
+    def _write_omm_impropers(self, dest):
+        if not self.improper_types: return
+        dest.write(' <CustomTorsionForce energy="k*(theta-theta0)^2">\n')
+        dest.write('  <PerTorsionParameter name="k"/>\n')
+        dest.write('  <PerTorsionParameter name="theta0"/>\n')
+        kconv = u.kilocalorie.conversion_factor_to(u.kilojoule)
+        tconv = u.degree.conversion_factor_to(u.radian)
+        def nowild(name):
+            return name if name != 'X' else ''
+        for (a1, a2, a3, a4), improp in iteritems(self.improper_types):
+            dest.write('  <Improper type1="%s" type2="%s" type3="%s" type4="%s"'
+                       ' k="%f" theta0="%f"/>\n' %
+                       (nowild(a1), nowild(a2), nowild(a3), nowild(a4),
+                       improp.psi_k*kconv, improp.psi_eq*tconv)
+            )
+        dest.write(' </CustomTorsionForce>\n')
+
+    def _write_omm_cmaps(self, dest):
+        if not self.cmap_types: return
+        dest.write(' <CmapTorsionForce>\n')
+        maps = dict()
+        counter = 0
+        econv = u.kilocalorie.conversion_factor_to(u.kilojoule)
+        for _, cmap in iteritems(self.cmap_types):
+            if id(cmap) in maps: continue
+            maps[id(cmap)] = counter
+            counter += 1
+            dest.write('  <Map>\n')
+            grid = cmap.grid.switch_range().T
+            for i in range(cmap.resolution):
+                dest.write('  ')
+                base = i * cmap.resolution
+                for j in range(cmap.resolution):
+                    dest.write(' %f' % (grid[base+j]*econv))
+                dest.write('\n')
+            dest.write('  </Map>\n')
+        used_torsions = set()
+        for (a1, a2, a3, a4, a5), cmap in iteritems(self.cmap_types):
+            if (a1, a2, a3, a4, a5) in used_torsions: continue
+            used_torsions.add((a1, a2, a3, a4, a5))
+            used_torsions.add((a5, a4, a3, a2, a1))
+            dest.write('   <Torsion map="%d" type1="%s" type=2"%s" type3="%s" '
+                       'type4="%s" type5="%s"/>\n' %
+                       (maps[id(cmap)], a1, a2, a3, a4, a5)
+            )
+        dest.write(' </CmapTorsionForce>\n')
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -258,10 +258,13 @@ class OpenMMParameterSet(ParameterSet):
         for (a2, a3, a1, a4), improp in iteritems(self.improper_periodic_types):
             # Try to make the wild-cards in the middle
             if a4 == 'X':
-                if a3 != 'X':
-                    a3, a4 = a4, a3
-                elif a2 != 'X':
+                if a2 != 'X':
                     a2, a4 = a4, a2
+                elif a3 != 'X':
+                    a3, a4 = a4, a3
+            if a2 != 'X' and a3 == 'X':
+                # Single wild-card entries put the wild-card in position 2
+                a2, a3 = a3, a2
             dest.write('  <Improper type1="%s" type2="%s" type3="%s" '
                        'type4="%s" periodicity1="%d" phase1="%f" k1="%f"/>\n' %
                        (a1, nowild(a2), nowild(a3), nowild(a4), improp.per,
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -106,6 +106,16 @@ class OpenMMParameterSet(ParameterSet):
         new_params.parametersets = params.parametersets
         new_params._combining_rule = params.combining_rule
         new_params.residues = params.residues
+        # Assign Lennard-Jones atom classes
+        ljtypes = dict()
+        counter = 0
+        for name, atom_type in iteritems(new_params.atom_types):
+            key = (atom_type.rmin, atom_type.epsilon, atom_type.rmin_14,
+                    atom_type.epsilon_14)
+            if key not in ljtypes:
+                ljtypes[key] = counter
+                counter += 1
+            atom_type.number = ljtypes[key]
 
         return new_params
 
@@ -163,12 +173,9 @@ class OpenMMParameterSet(ParameterSet):
         for name, atom_type in iteritems(self.atom_types)
             assert atom_type.atomic_number >= 0, 'Atomic number not set!'
             element = Element[atom_type.atomic_number]
-            key = (atom_type.rmin, atom_type.epsilon)
-            if key not in ljtypes:
-                ljtypes[key] = counter
-                counter += 1
             dest.write('  <Type name="%s" class="%d" element="%s" mass="%f"/>\n'
-                % (name, ljtypes[key], element, atom_type.mass or Mass[element])
+                % (name, atom_type.number, element,
+                   atom_type.mass or Mass[element])
             )
         dest.write(' </AtomTypes>\n')
 
@@ -221,3 +228,5 @@ class OpenMMParameterSet(ParameterSet):
                        'angle="%s" k="%s"/>\n' %
                        (a1, a2, a3, angle.theteq*tconv, angle.k*kconv))
         dest.write(' </HarmonicAngleForce>\n')
+
+    def _write_omm_dihedrals(self, dest):
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -110,7 +110,7 @@ class OpenMMParameterSet(ParameterSet):
 
         return new_params
 
-    def write(self, dest, original_file='', reference=''):
+    def write(self, dest, provenance=None):
         """ Write the parameter set to an XML file for use with OpenMM
 
         Parameters
@@ -118,11 +118,17 @@ class OpenMMParameterSet(ParameterSet):
         dest : str or file-like
             The name of the file or the file-like object (with a ``write``
             attribute) to which the XML file will be written
-        original_file : str, optional
-            The original file name to put in the provenance information. Default
-            is empty string
-        reference = str, optional
-            The literature reference for this force field
+        provenance : dict, optional
+            If present, the XML file will be tagged with the available fields.
+            The keys of this dictionary are turned into the XML tags, and the
+            values become the contents of that tag. Default is no provenance
+
+        Notes
+        -----
+        The generated XML file will have the XML tag ``DateGenerated`` added to
+        the provenance information set to the current date. Therefore, you
+        should not provide this information in ``provenance`` (it will be
+        removed if it is provided).
         """
         if isinstance(dest, string_types):
             dest = genopen(dest, 'w')
@@ -132,7 +138,7 @@ class OpenMMParameterSet(ParameterSet):
 
         try:
             dest.write('<ForceField>\n')
-            self._write_omm_provenance(dest, original_file, reference)
+            self._write_omm_provenance(dest, provenance)
             self._write_omm_atom_types(dest)
             self._write_omm_residues(dest)
             self._write_omm_bonds(dest)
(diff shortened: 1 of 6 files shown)


jchodera:
Hooray!  Sorry for the confusion--I was under the impression @chayast and @rafwiewiora would be helping you write this.

parmed/openmm/parameters.py
```diff
@@ -122,18 +138,87 @@ def write(self, dest, provenance=None):
...
+        dest.write(' <Residues>\n')
+        for name, residue in iteritems(self.residues):
+            if not isinstance(residue, ResidueTemplate):
+                continue
+            dest.write('  <Residue name="%s">\n' % residue.name)
+            for atom in residue.atoms:
+                dest.write('   <Atom name="%s" type="%s" charge="%f"/>\n' %
+                           (atom.name, atom.atom_type.number, atom.charge))
+            for bond in residue.bonds:
+                dest.write('   <Bond atomName1="%s" atomName2="%s" />\n' %
```
jchodera: Hooray for using atom names!

parmed/openmm/parameters.py
```diff
@@ -122,18 +138,87 @@ def write(self, dest, provenance=None):
...
+        if not self.bond_types: return
+        dest.write(' <HarmonicBondForce>\n')
+        bonds_done = set()
+        lconv = u.angstroms.conversion_factor_to(u.nanometers)
+        kconv = u.kilocalorie.conversion_factor_to(u.kilojoule) / lconv**2 * 2
+        for (a1, a2), bond in iteritems(self.bond_types):
+            if (a1, a2) in bonds_done: continue
+            bonds_done.add((a1, a2))
+            bonds_done.add((a2, a1))
+            dest.write('  <Bond class1="%s" class2="%s", length="%f", k="%f"/>\n'
```
jchodera: We don't want to use `class` here---we want to use `type`, since we will be using the translation
* OpenMM `type` is AMBER atom type
* OpenMM `class` is AMBER Lennard-Jones class

parmed/openmm/parameters.py
```diff
@@ -122,18 +138,87 @@ def write(self, dest, provenance=None):
...
+        if not self.angle_types: return
+        dest.write(' <HarmonicAngleForce>\n')
+        angles_done = set()
+        tconv = u.degree.conversion_factor_to(u.radians)
+        kconv = u.kilocalorie.conversion_factor_to(u.kilojoule) * 2
+        for (a1, a2, a3), angle in iteritems(self.angle_types):
+            if (a1, a2, a3) in angles_done: continue
+            angles_done.add((a1, a2, a3))
+            angles_done.add((a3, a2, a1))
+            dest.write('  <Angle class1="%s" class2="%s" class3="%s" '
```
jchodera: We don't want to use `class` here---we want to use `type`, since we will be using the translation
* OpenMM `type` is AMBER atom type
* OpenMM `class` is AMBER Lennard-Jones class

rafwiewiora:
Sweet, thanks @swails! Well, clearly @swails couldn't wait anymore for us to get round to it :) Thanks a lot! Hopefully I can take this further on the weekend.

parmed/openmm/parameters.py
```diff
@@ -122,18 +138,87 @@ def write(self, dest, provenance=None):
...
+                   datetime.datetime.now().timetuple()[:3])
+        dest.write('  <Reference>%s</Reference>\n' % reference)
+        dest.write(' </Info>\n')
+
+    def _write_omm_atom_types(self, dest):
+        dest.write(' <AtomTypes>\n')
+        for (name, i), atom_type in iteritems(self.atom_types_tuple):
+            assert atom_type.atomic_number >= 0, 'Atomic number not set!'
+            element = Element[atom_type.atomic_number]
+            dest.write('  <Type name="%s" class="%d" element="%s" mass="%f"/>\n'
```
jchodera: We want to use the translation:
* OpenMM `type` is AMBER atom type
* OpenMM `class` is AMBER Lennard-Jones class 

Where does `atom_types_tuple` get set here?

parmed/openmm/parameters.py
```diff
@@ -101,10 +106,18 @@ def from_parameterset(cls, params, copy=False):
         new_params.parametersets = params.parametersets
         new_params._combining_rule = params.combining_rule
         new_params.residues = params.residues
+        # Make sure we have a complete set of atom types in the _tuple list
+        if len(params.atom_types_tuple) == 0:
+            i = 1
+            for name, type in iteritems(params.atom_types):
```
jchodera: I think we want to use the convention that OpenMM `type` is the AMBER atom type, while the OpenMM `class` is the AMBER Lennard-Jones reduced atom class (where several types map to one typename for the purposes of assigning LJ parameters).

swails:
> Hooray! Sorry for the confusion--I was under the impression @chayast and @rafwiewiora would be helping you write this.

I didn't finish it :).  I figured it would be easier for everyone if I got things started.

hainm:
> I didn't finish it :). I figured it would be easier for everyone if I got things started.

But it's clearly that you love writing code. :+1: +

parmed/openmm/parameters.py
swails: > I think we want to use the convention that OpenMM type is the AMBER atom type, while the OpenMM class is the AMBER Lennard-Jones reduced atom class (where several types map to one typename for the purposes of assigning LJ parameters).

Ah!  Right, I remember that conversation now.  That makes things easier.

swails:
@jchodera -- I made the relevant changes and expanded the writing for dihedrals, impropers, and cmaps.  I'll leave this open for comments until sometime tomorrow and then merge it so people can start playing with it.

For anyone that makes changes, I think the easiest way to work efficiently is to submit PRs once you've started work on something.  That way we know where not to duplicate efforts (that's why I submitted this PR before I had gotten the whole thing working).  Just remember to submit the PR to `ommparams`, not `master`.

jchodera:
Just looking through this again now...

parmed/openmm/parameters.py
```diff
@@ -101,10 +107,20 @@ def from_parameterset(cls, params, copy=False):
...
+                    atom_type.epsilon_14)
+            if key not in ljtypes:
+                ljtypes[key] = counter
+                counter += 1
+            atom_type.number = ljtypes[key]
 
         return new_params
 
-    def write(self, dest, provenance=None):
+    def write(self, dest, original_file='', reference=''):
```
jchodera: What if we added an optional `metadata` that could contain additional metadata to be added?  For example, we may want a human-readable description.

parmed/openmm/parameters.py
```diff
@@ -122,18 +141,180 @@ def write(self, dest, provenance=None):
...
+#           self._write_omm_scripts(dest)
+#           self._write_omm_nonbonded(dest)
         finally:
+            dest.write('</ForceField>\n')
             if own_handle:
                 dest.close()
+
+    def _write_omm_provenance(self, dest, original_file, reference):
+        dest.write(' <Info>\n')
+        dest.write('  <Source>%s</Source>\n' % original_file)
```
jchodera: We should write the `md5` hash of the original file as well, since the source filename is insufficient to uniquely identify the file version.

parmed/openmm/parameters.py
```diff
@@ -101,10 +107,20 @@ def from_parameterset(cls, params, copy=False):
...
         new_params._combining_rule = params.combining_rule
         new_params.residues = params.residues
+        # Assign Lennard-Jones atom classes
+        ljtypes = dict()
+        counter = 0
+        for name, atom_type in iteritems(new_params.atom_types):
+            key = (atom_type.rmin, atom_type.epsilon, atom_type.rmin_14,
+                    atom_type.epsilon_14)
+            if key not in ljtypes:
+                ljtypes[key] = counter
```
jchodera: What if we instead use the text name of the atom type?  For example, the [relevant section of the `parm96.dat` file](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/parm96.dat#L450-L451) specifies that a number of types map to `N` or `C`:
```
N   NA  N2  N*  NC  NB  NP  NO
C   C*  CA  CB  CC  CN  CM  CK  CQ  CW  CV  CR  CA  CX  CY  CD
```
while other LJ types are specified for their standard Amber type names.  It might be good to keep this mapping for `ljtypes`, rather than introducing a confusing new atom type numbering scheme that is used for OpenMM `class` assignment below.

parmed/openmm/parameters.py
```diff
@@ -122,18 +141,180 @@ def write(self, dest, provenance=None):
...
+        dest.write('  <Reference>%s</Reference>\n' % reference)
+        dest.write(' </Info>\n')
+
+    def _write_omm_atom_types(self, dest):
+        dest.write(' <AtomTypes>\n')
+        for name, atom_type in iteritems(self.atom_types):
+            assert atom_type.atomic_number >= 0, 'Atomic number not set!'
+            element = Element[atom_type.atomic_number]
+            dest.write('  <Type name="%s" class="%d" element="%s" mass="%f"/>\n'
+                % (name, atom_type.number, element,
```
jchodera: For `class`, I think we should use the reduced Amber atom type name scheme, rather than an arbitrary numbering scheme.  For example, atom type `NA` and `N` both map to LJ class type `N` in `parm96.dat`, while `N3` has its own LJ class type.

See https://github.com/ParmEd/ParmEd/pull/502/files#r47437846

parmed/openmm/parameters.py
```diff
@@ -122,18 +141,180 @@ def write(self, dest, provenance=None):
...
+
+    def _write_omm_residues(self, dest):
+        self.typeify_templates()
+        dest.write(' <Residues>\n')
+        for name, residue in iteritems(self.residues):
+            if not isinstance(residue, ResidueTemplate):
+                continue
+            dest.write('  <Residue name="%s">\n' % residue.name)
+            for atom in residue.atoms:
+                dest.write('   <Atom name="%s" type="%s" charge="%f"/>\n' %
```
jchodera: Do we have to worry about listing charges to the same precision as the input files?  I think the current scheme would ensure that full available precision was printed, which seems like what we want.

jchodera:
OK, finished reviewing this.  Looking great!

The only major thing is that I believe we want to use Amber reduced LJ type names for OpenMM atom `class`es, rather than arbitrary numbers.  For example, [in `parm96.dat`](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/parm96.dat#L450-L488), atom type NA and N both map to LJ class type N in parm96.dat, while N3 has its own LJ class type.

parmed/openmm/parameters.py
hainm: I think it's good to add `metadata` too.

jchodera: Maybe it should take either a `str` (pre-formatted XML block) or a `dict` (key/value pairs)?

swails: I only based the args on what was in amoeba09.xml that Peter generated.  Should I just take a `dict` and use the keys as XML tags?  Or just take it as kwargs?  My API ideas:

```Python
def write(self, dest, **metadata):
    for key, val in metadata.items():
        dest.write('  <%s>%s</%s>\n' % (key, val, key))
```
Or
```Python
```

parmed/openmm/parameters.py
swails: This can be passed in via the metadata API suggestion above.

parmed/openmm/parameters.py
swails: -1

This is an Amber-specific implementation detail, and ParmEd tries to avoid those that don't provide a tangible benefit.  IMO, there is only one reason to define L-J classes that are more general than atom types, and that's to compress the size of the L-J A- and B-coefficient tables so they don't blow GPU shared memory when we have NBFIXes (which some Amber FFs do, and more CHARMM FFs do).  If that was not a consideration, there is absolutely no advantage (at all) for using classes instead of types to define L-J parameters.  It just adds unnecessary complexity in the conversion script without any tangible benefit.

Also, it means the same code is optimal for both CHARMM and AMBER parameter file conversions.  IMO, this is a real win.

So my vote is either 

a) Compress as much as possible the way I'm doing now
b) Don't define classes at all, and just use type for everything

The disadvantages of this proposal:

1) Does not compress types as much as possible, which increases the possibility that NBFIX tables will overflow shared memory.
2) Requires Amber-specific implementation details that are parm.dat-file specific (there is no way to do this assignment in an frcmod file, for instance)

I don't see a real advantage in this scheme except that a tiny number of people (i.e., you, me, and maybe 2 or 3 other people) will see the parallels (and even here, where's the tangible benefit?)

parmed/openmm/parameters.py
swails: -1 for the same explanation above.  Also, in Amber prmtop files you get numeric indices (`ATOM_TYPE_INDEX` or whatever)...

parmed/openmm/parameters.py
swails: What is your proposed change?  I picked this because I didn't think it mattered how much precision I printed stuff to in the XML file...

swails:
> The only major thing is that I believe we want to use Amber reduced LJ type names for OpenMM atom classes, rather than arbitrary numbers. For example, in parm96.dat, atom type NA and N both map to LJ class type N in parm96.dat, while N3 has its own LJ class type.

Going from type being an index that iterated over each Amber atom type in each AA/NA residue environment (which is why `amber10.xml` has 2000 types) to using the Amber atom types with charges defined in residue templates (which is why `parm10.dat` has only 85) resulted in a *huge* complexity reduction.  It also significantly simplified the conversion script.  This proposal reduces the L-J types by 22.  My approach reduces it to 28 total (for ff14SB).  And unlike the first change, this would add complexity to the code base to support.

parmed/openmm/parameters.py
swails: > Maybe it should take either a str (pre-formatted XML block) or a dict (key/value pairs)?

The latter we can ensure gets properly XML-formatted.  The former we can't.  Ergo, I think the latter is far less error-prone and exposes a more user-friendly API.

jchodera: I like personally like having `metadata` be a dict in case we add other options later on:
```python
def write(self, dest, metadata):
    for key, val in metadata.items():
        dest.write('  <%s>%s</%s>\n' % (key, val, key))
```
but I certainly don't feel strongly about this.

> The latter we can ensure gets properly XML-formatted. 

Excellent point!

parmed/openmm/parameters.py
jchodera: OK.

swails:
One thing I'll say is that I did think about making the class definitions match how Amber compresses L-J types, and the other approach for the reasons I listed.

parmed/openmm/parameters.py
jchodera: Our primary concern is to translate the original parameter specification files (`*.dat` files for AMBER) into faithful representations of these parameter sets in `ffxml` files, ideally in a way that maximizes our ability to mix-and-match sets of these files.

I think there's value in ending up with a `<NonbondedForce>` block that essentially encodes the same information at the same level of detail as the [`parm96.dat` file does](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/parm96.dat#L450-L488).

It sounds like your current approach tries to compress even further than what is specified by these type maps:
```
N   NA  N2  N*  NC  NB  NP  NO
C   C*  CA  CB  CC  CN  CM  CK  CQ  CW  CV  CR  CA  CX  CY  CD
```
but using numeric types---is that correct?

What if we compromised, and used a sensible reference atom type name as the class name?  I think this will be much more comprehensible to anyone looking at these files in the future than class numbers.

swails: > Our primary concern is to translate the original parameter specification files (*.dat files for AMBER) into faithful representations of these parameter sets in ffxml files,

I think the primary concern should be to translate the original parameter specification files (\*.dat *and* \*.frcmod -- don't forget that all FFs are really combinations of these) into XML files that give equivalent energies.  Faithful representations are impossible where you need a `<script>` section, since no other program has an analogy to that.

> ideally in a way that maximizes our ability to mix-and-match sets of these files.

Good point.  My approach would not work here.  But just using types (and no classes) *would* ;).

That said, taking one of the atom type names should work nicely.  I'll probably just keep the first type corresponding each L-J parameter type as the class name.

jchodera: > That said, taking one of the atom type names should work nicely. I'll probably just keep the first type corresponding each L-J parameter type as the class name.

Maybe we should think a bit more deeply about the most obvious file we would want to mix-and-match: [`gaff.dat`](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/gaff.dat), perhaps mixed with another one of the AMBER protein forcefield types.

`gaff.dat` defines [separate LJ parameters for each atom type](https://github.com/choderalab/ambermini/blob/master/share/amber/dat/leap/parm/gaff.dat#L5891-L5962), even if they are duplicates.  This means, as you observe, that using compressed numerical groups wouldn't let us mix-and-match due to collisions, but either the compressed types written as corresponding atom type names or writing them all by atom name would work.

Is there any benefit in adding a flag to select which scheme we want?  Or should we just figure out the optimal way and stick with that?

swails: > gaff.dat defines separate LJ parameters for each atom type, even if they are duplicates. This means, as you observe, that using compressed numerical groups wouldn't let us mix-and-match due to collisions, but either the compressed types written as corresponding atom type names or writing them all by atom name would work.

True.  But now let's look at what else is allowed in Amber parameter files -- you can redefine L-J coefficients in an frcmod file for any atom type.  In our proposal, one of the atom types that shares L-J parameters with other types is defined as the "parent" of the L-J class.  If this atom type has its L-J parameters redefined in an frcmod file in some future force field (like the Garcia-Chen FF for nucleic acids), it could accidentally replace all of the L-J parameters for the whole group.

We're making this too complicated.  As I said, the only place where L-J type compression makes a difference is when you have NBFIXes and have to store lookup tables in shared memory.  But that is done via a `<script>` section where you can implement necessary compression in there.  And you would have to do this, anyway, since you don't want to include the parameters for atom types not in your system.

So I vote we either make class and type the same for these families of force fields, or just do away with class altogether (those two are identical, so I vote for the simplicity of the latter).

jchodera: > So I vote we either make class and type the same for these families of force fields, or just do away with class altogether (those two are identical, so I vote for the simplicity of the latter).

Agreed.  That seems most sensible for now.

swails:
OK.  Atom classes are gone.  The only thing we have left to do here is the nonbonded force.

There's one other idea I want to float, and that has to do with how the angle force is defined when Urey-Bradley terms are present.  Currently, what @peastman did to implement Urey-Bradley in the CHARMM FF was do angles via the standard `HarmonicAngleForce` and then put the Urey-Bradley assignment inside the same `<script>` block that did the L-J custom nonbonded force.

What I'd like to do when U-B terms are present, both to simplify the XML file and make it easier to read it back in to ParmEd, is to use a `CustomCompoundBondForce` and do both the angle and distance forces in the same term.

@peastman -- would this work in `ForceField` (and/or would it be easy enough to add support for)?

parmed/openmm/parameters.py
```diff
@@ -122,18 +131,179 @@ def write(self, dest, provenance=None):
...
+                        improp.phase*pconv, improp.phi_k*kconv)
+            )
+        dest.write(' </PeriodicTorsionForce>\n')
+
+    def _write_omm_impropers(self, dest):
+        if not self.improper_types: return
+        dest.write(' <CustomTorsionForce energy="k*(theta-theta0)^2">\n')
+        dest.write('  <PerTorsionParameter name="k"/>\n')
+        dest.write('  <PerTorsionParameter name="theta0"/>\n')
+        kconv = u.kilocalorie.conversion_factor_to(u.kilojoule)
```
hainm: `conversion_factor_to` to `conversion_to` or 'converted_to'

swails: `conversion_factor_to` is the correct API call.

jchodera:
> OK. Atom classes are gone. The only thing we have left to do here is the nonbonded force.

Great!  We just have to remember to add 
```XML
<UseAttributeFromResidue name="charge"/>
```
to the `<NonbondedForce>` block.

> There's one other idea I want to float, and that has to do with how the angle force is defined when Urey-Bradley terms are present. Currently, what @peastman did to implement Urey-Bradley in the CHARMM FF was do angles via the standard HarmonicAngleForce and then put the Urey-Bradley assignment inside the same <script> block that did the L-J custom nonbonded force.
> What I'd like to do when U-B terms are present, both to simplify the XML file and make it easier to read it back in to ParmEd, is to use a CustomCompoundBondForce and do both the angle and distance forces in the same term.
> @peastman -- would this work in ForceField (and/or would it be easy enough to add support for)?

[`ForceField`](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py) doesn't yet have a `CustomCompoundForceGenerator`, so we'd have to add one.  But this sounds like a very sensible approach, and the omission of a `CustomCompoundForceGenerator` seems like an oversight.

parmed/openmm/parameters.py
swails: Oh, you were suggestion a change.  `parmed.unit` is `simtk.unit` bundled (and the latter replaces this package if it's available).  So `unit` API changes are not up for discussion here (you'd have to take that to pandegroup/openmm, and backwards-incompatible changes in the public-facing API will almost certainly be rejected).

`conversion_factor_to` is better than `conversion_to` or `converted_to` IMO, anyway, since it tells you *exactly* what you get in a nice, concise function name.

swails:
> Great! We just have to remember to add
> 
> <UseAttributeFromResidue name="charge"/>
> to the <NonbondedForce> block.

Right.  I still had that in mind :)

parmed/openmm/parameters.py
hainm: :D 

I suggested just because of this

`kilocalorie.conversion_factor_to(u.kilojoule)`

vs

`kilocalorie.converted_to(u.kilojoule)`

swails: `conversion_factor_to` makes it obvious that you're getting a floating point constant (a factor) that will convert kilocalories to kilojoules, and it makes it clear that this is operating on units rather than quantities.  The latter spelling is not as clear.  That would actually make more sense on a `Quantity` -- something like

```Python
e = 10.0 * u.kilocalories
e.converted_to(u.kilojoules)
```

And even there, it's not clear if the returned value has units or not.  The spelling for that interface is

```Python
e = 10.0 * u.kilocalories
e.value_in_unit(u.kilojoules) # returns a unitless value
e.in_units_of(u.kilojoules) # returns a Quantity with units kilojoules
```

More than almost any other package I regularly use, the `simtk.unit` API is incredibly natural and intuitive.  Their methods have short, concise names that are very descriptive -- a difficult combination to find in many cases.  You'll be hard pressed to find method names that I like as much, let alone ones that I like better.

hainm: yeah, just because I did not understand the 'factor' word.

swails:
OK.  All that's left now is the nonbonded force.  I'll pose the question about handling the Urey-Bradley terms (basically the requirements are that `ForceField` would need to know how to handle `CustomCompoundBondForce` construction for angles with Urey-Bradley distances when building a `System`.

I'm going to merge this for now so we can all work on the same code base.

jchodera:
Awesome! Thanks!

jchodera:
One other question of relevance here: I had totally forgotten that we may want to split out GB parameters into a separate file so we can mix-and-match them.

Unfortunately, the way GB parameters are assigned is a bit...well, wonky.  That leaves us two choices:
* write a custom `<Script>` block to do the assignment based on examining the `Topology`
* use the OpenMM `class` as a sort of GB atom class

Any other ideas?

swails:
I presume you don't want to do what we have now where the GB models are their own separate force fields, right?

> Any other ideas?

We could add `solvation_radius` and `gb_screen` attributes to the atom types the same way the charge is done.  We could even go all out and do `mbondi_radius`, `mbondi2_radius`, `mbondi3_radius`, and `bondi_radius` attributes for each of the atom types so it would have all the info it needs for any GB model.

The main issue here is that these parameters are, in some cases (for C and H) assigned based on connectivity.  So we would have to pass the `ResidueTemplate` objects through the radius assigner (`parmed/tools/changeradii.py`) to get radii for each of the atom types in the residues, which is certainly doable (it would need small modifications to account for head and tail connections that are not registered bonds).

This would reduce the number of places that the GB radii are hard-coded (by not adding another one inside `parmed/openmm/parameters.py`).  But one thing this would mean is that atom types that are *not* part of the standard residue templates would either need to be given no radii, or "best-guess" radii.  Or they need to be deleted altogether, since there is really no straightforward way to use these atom types in OpenMM that I'm aware of.

> use the OpenMM class as a sort of GB atom class

I think I can extend the same arguments for why we should just use `type` instead of `class` as I did for the L-J parameters.

peastman:
> What I'd like to do when U-B terms are present, both to simplify the XML file and make it easier to read it back in to ParmEd, is to use a CustomCompoundBondForce and do both the angle and distance forces in the same term.

That would be fine, but you'll still need a `<script>` to do it.  ForceField can't automatically add a CustomCompoundBondForce, because it doesn't know where bonds should be added.

PR #509, opened 2015-12-18
Fix some CHARMM conversion/writing issues

swails:
- Use the same index (typically starting from 1) for both CHARMM PSF and CRD files
- Avoid using the `*` character in atom types, since it is the CHARMM wild-card
- Improve GROUP handling and make better default groups
- Make indices in PSF and CHARMM CRD start from 0


parmed/__init__.py
```diff
@@ -6,7 +6,7 @@
 # Version format should be "major.minor.patch". For beta releases, attach
 # "-beta#" to the end. The beta number will be turned into another number in the
 # version tuple
-__version__ = '2.0.12'
+__version__ = '2.1.0'
```
hainm: do you have any rule for version jump (0.12 to 1.0), or just do it?

swails: The `parmed.topologobjects.Group` API changed a little.  Ergo, bigger version bump.

hainm: sorry, I meant like jumping from '0.12' to '1.0' or '0.12' to '0.13'.

swails: Exactly.  A small thing, like fixing a bug or adding a small feature is a bump from `x.y.z` to `x.y.z+1`.  Minor API changes, like what I did here, is `x.y.z` to `x.y+1.0`.  Major changes increment `x`.

hainm: I see.

PR #517, opened 2015-12-30
Change atom typing scheme

swails:
The main purpose of this PR is to implement what we are discussing in pandegroup/openmm#1298.  Parameters are now defined by class, and atom types are named by `RESNAME-ATOMNAME`.

This PR also does two other things:

- Merges the master branch
- Fixes `<Residue>` writing (residues were missing a `</Residue>` tag ending each residue definition).


jchodera:
Cool!  Let's see if this works.

To be honest, I doubt the atom typenames are useful (especially since we will likely go with a ‘<script>‘ block for GB typing), but Peter seems adamant we do this.

We can always revisit later.

swails:
I edited your comment to turn the quote into a back-tick (otherwise the markdown hid everything after `<script>`, which rendered me *very* confused ;)).

> To be honest, I doubt the atom typenames are useful (especially since we will likely go with a <script> block for GB typing), but Peter seems adamant we do this.

I'm not sure exactly what you mean by this.  You mean the `name="residue-atom"` construct?  If so, this is not necessary to convert Amber and CHARMM parameter files.  But it *is* necessary to convert GROMACS and Tinker parameter files, since GROMACS allows you to override specific parameters, and Tinker uses the same system that OpenMM does (I recall Peter saying it was where he got the idea, but I can't find that conversation now).

It also allows you, in the future, to do any kind of typing you could possibly want.  So while the standard bonded and nonbonded parameters that we're starting with here don't need fine-grained types, converting Tinker FF files and some GROMACS FF files (e.g., ff99SB-ildn) does require this (and GB typing can be done immediately with this scheme).

jchodera:
Whoops! This phone keyboard has several kinds of backticks, none of which
seem to be what I want!

jchodera:
Thanks!

PR #518, opened 2016-01-03
Revert to using AMBER atom types in generating OpenMM ffxml atom types

jchodera:
Following the discussion in https://github.com/pandegroup/openmm/issues/1298, there is no way for us to break `ffxml` files into separate files for parameters, amino acids, and nucleic acids (and possibly other components) if atom types are based on `resname-atomname` due to the requirement that parameters can only refer to atomtypes that have been defined in the files that have already been read.  Even if the atom classes were set to be AMBER atom types, atom classes are simply expanded into the set of matching atom types that have been defined at the time the parameters are read, so this would not work without major changes to `ForceField`.

See [the OpenMM User Guide](http://docs.openmm.org/6.3.0/userguide/application.html#using-multiple-files), which states:
> If multiple XML files are specified when a ForceField is created, their definitions are combined as follows.
> A file may refer to atom types and classes that it defines, as well as those defined in previous files. It may not refer to ones defined in later files. This means that the order in which files are listed when calling the ForceField constructor is potentially significant.

Instead, if we define OpenMM atom types to be identical to AMBER atom types, we can define atom types and parameters in one file and leave residue definitions (amino acids, nucleic acids) for separate files.  This will also allow us to generate a `gaff.xml` that include GAFF atom types and parameters that can be used with the [residue template generator registration scheme](https://github.com/pandegroup/openmm/pull/1312) for incorporating Antechamber- or other parameterization schemes (which I can implement very soon after this).

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -166,14 +166,12 @@ class OpenMMParameterSet(ParameterSet):
 
     def _write_omm_atom_types(self, dest):
         dest.write(' <AtomTypes>\n')
-        for name, residue in iteritems(self.residues):
-            if not isinstance(residue, ResidueTemplate):
-                continue
-            for atom in residue.atoms:
-                element = Element[atom.atom_type.atomic_number]
-                dest.write('  <Type name="%s-%s" class="%s" element="%s" '
-                           'mass="%f"/>\n' % (name, atom.name, atom.type,
-                           element, atom.atom_type.mass))
+        for name, atom_type in iteritems(self.atom_types):
+            assert atom_type.atomic_number >= 0, 'Atomic number not set!'
+            element = Element[atom_type.atomic_number]
+            dest.write('  <Type name="%s" element="%s" mass="%f"/>\n'
+                       % (name, element, atom_type.mass or Mass[element])
+                       )
         dest.write(' </AtomTypes>\n')
 
     def _write_omm_residues(self, dest):
@@ -183,8 +181,8 @@ class OpenMMParameterSet(ParameterSet):
                 continue
             dest.write('  <Residue name="%s">\n' % residue.name)
             for atom in residue.atoms:
-                dest.write('   <Atom name="%s" type="%s-%s" charge="%f"/>\n' %
-                           (atom.name, name, atom.name, atom.charge))
+                dest.write('   <Atom name="%s" type="%s" charge="%f"/>\n' %
+                           (atom.name, atom.type, atom.charge))
             for bond in residue.bonds:
                 dest.write('   <Bond atomName1="%s" atomName2="%s" />\n' %
                            (bond.atom1.name, bond.atom2.name))
@@ -207,7 +205,7 @@ class OpenMMParameterSet(ParameterSet):
             if (a1, a2) in bonds_done: continue
             bonds_done.add((a1, a2))
             bonds_done.add((a2, a1))
-            dest.write('  <Bond class1="%s" class2="%s", length="%f", k="%f"/>\n'
+            dest.write('  <Bond type1="%s" type2="%s", length="%f", k="%f"/>\n'
                        % (a1, a2, bond.req*lconv, bond.k*kconv))
         dest.write(' </HarmonicBondForce>\n')
 
@@ -221,7 +219,7 @@ class OpenMMParameterSet(ParameterSet):
             if (a1, a2, a3) in angles_done: continue
             angles_done.add((a1, a2, a3))
             angles_done.add((a3, a2, a1))
-            dest.write('  <Angle class1="%s" class2="%s" class3="%s" '
+            dest.write('  <Angle type1="%s" type2="%s" type3="%s" '
                        'angle="%s" k="%s"/>\n' %
                        (a1, a2, a3, angle.theteq*tconv, angle.k*kconv))
         dest.write(' </HarmonicAngleForce>\n')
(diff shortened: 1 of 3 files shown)


parmed/openmm/parameters.py
```diff
@@ -312,8 +310,48 @@ def _write_omm_cmaps(self, dest):
...
+
+    def _write_omm_nonbonded(self, dest):
+        if self.combining_rule == 'geometric':
+            if len(self.nbfix_types) > 0:
+                raise NotImplementedError("Nonbonded forces with geometric combining rules have not been implemented yet.")
+            else:
+                raise NotImplementedError("Nonbonded forces with geometric combining rules have not been implemented yet.")
+
+        if len(self.nbfix_types) > 0:
+            raise NotImplementedError("Nonbonded forces with geometric combining rules have not been implemented yet.")
```
swails: Error message is misleading here.  The incompatibility is NBFIX, not geometric combining

parmed/openmm/parameters.py
```diff
@@ -312,8 +310,48 @@ def _write_omm_cmaps(self, dest):
...
+        # Compute conversion factors for writing in natrual OpenMM units.
+        length_conv = u.angstrom.conversion_factor_to(u.nanometer)
+        ene_conv = u.kilocalories.conversion_factor_to(u.kilojoules)
+
+        # TODO: We shouldn't hardcode these, but I don't know how they are stored.
+        coulomb14scale = 1.0 / 1.2
+        lj14scale = 0.5
+
+        # Write NonbondedForce records.
+        dest.write(' <NonbondedForce coulomb14scale="%f" lj14scale="%f">\n' % (coulomb14scale, lj14scale))
```
swails: After this you need to add the line

```Python
        dest.write('  <UseAttributeFromResidue name="charge">\n')
```

parmed/openmm/parameters.py
```diff
@@ -312,8 +310,48 @@ def _write_omm_cmaps(self, dest):
...
+        if len(self.nbfix_types) > 0:
+            raise NotImplementedError("Nonbonded forces with geometric combining rules have not been implemented yet.")
+
+        # Compute conversion factors for writing in natrual OpenMM units.
+        length_conv = u.angstrom.conversion_factor_to(u.nanometer)
+        ene_conv = u.kilocalories.conversion_factor_to(u.kilojoules)
+
+        # TODO: We shouldn't hardcode these, but I don't know how they are stored.
+        coulomb14scale = 1.0 / 1.2
+        lj14scale = 0.5
```
swails: These are stored in the `dihedral_types` list.  I view 1-4 electrostatic interactions as part of the torsion profile itself, and this approach allows you to specify variable 1-4 scaling (like what you get with GLYCAM-Amber combinations).  I'll modify this once it's merged.

jchodera:
Thanks for the comments!  Fixing now.

parmed/openmm/parameters.py
jchodera: Are there cases where the 1-4 scaling may differ for different interactions?  If so, we can presumably extend `ForceField.NonbondedGenerator` so that we could opt-out of the universal 14 scaling factors and instead specify our own exceptions.

jchodera:
I'd love to get more insight on how we can do more thorough testing.

One way would be to parameterize a few PDB files using `ForceField` and compare energies between OpenMM and AMBER.  Is there existing code we could utilize for that?

Another way would be to also add the ability for ParmEd to *load* `ffxml` files, and make sure that we can execute lossless round-trips of converting parameter sets to/from `ffxml`.

jchodera:
Should be good to merge.  The next step is to add more testing, I think!

swails:
ParmEd *can* load `ffxml` files, but not into a ParmEd `ParameterSet`, just into a `simtk.openmm.app.ForceField`:

```Python
Python 3.5.1 |Continuum Analytics, Inc.| (default, Dec  7 2015, 11:16:01) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import parmed as pmd
>>> pmd.load_file('amber99sbildn.xml')
<simtk.openmm.app.forcefield.ForceField object at 0x7fb9e72cfeb8>
>>> 
```

It would be really hard to parse a ffxml directly into a `ParameterSet`, because of the `<script>` sections -- which is where NBFIXes, Urey-Bradleys, geometric combining rules, etc. are all handled.  I'm *really* not prepared to read and analyze arbitrary Python source in order to try and build up a `ParameterSet`.  While it would be nice to round-trip, it can't be used at all for CHARMM, and only for the simplest of Amber force fields.

The best way that *I* can think of to test this is actually to start with standard Amber or CHARMM files (prmtop/inpcrd or psf/coord).  `OpenMMParameterSet` has a `from_structure` constructor which you can use to instantiate a `OpenMMParameterSet` from a `Structure`.  Then `ResidueTemplateContainer` has a `from_structure` constructor as well so you can construct residue templates from the defined bonds.  Build ffxml and PDB files from those structures and run them through OpenMM and compare to the energies you get directly from the Amber or CHARMM prmtop/psf file.  ParmEd has a bunch of helper functions that can simplify a lot of this for you, as well as functionality to decompose energy contributions just given a `System`:

```Python
In [1]: import parmed as pmd

In [2]: parm = pmd.load_file('ash.parm7', 'ash.rst7')

In [3]: parm
Out[3]: <AmberParm 25 atoms; 3 residues; 24 bonds; parametrized>

In [4]: pmd.openmm.energy_decomposition_system(parm, parm.createSystem())
Out[4]: 
[('HarmonicBondForce', 5.443559276452141),
 ('HarmonicAngleForce', 2.8765883826828977),
 ('PeriodicTorsionForce', 24.369689112675623),
 ('NonbondedForce', -30.23822122675511),
 ('CMMotionRemover', 0.0)]
```

swails:
Thanks!

jchodera:
> The best way that I can think of to test this is actually to start with standard Amber or CHARMM files (prmtop/inpcrd or psf/coord). OpenMMParameterSet has a from_structure constructor which you can use to instantiate a OpenMMParameterSet from a Structure. Then ResidueTemplateContainer has a from_structure constructor as well so you can construct residue templates from the defined bonds.

Conceptually, this sounds a little tricky: We're building a forcefield *from* a `prmtop` file, and then generating an `ffxml` file from this.  That's probably a good thing to test, but we really want to make sure we can convert AMBER forcefield files to OpenMM `ffxml` files and have both `LEaP` and `ForceField` produce the same systems (as much as possible).  

What if we instead either had the tests call `LEaP` to create some `parm7/rst7` files (or pregenerate some if you don't want to use `ambermini` as a testing prerequisite) and then used your `energy_decomposition_system` to compare energies `Force` by `Force` between the `parm7/rst7` and `ForceField` versions of the same system?

PR #519, opened 2016-01-04
Small tweaks to openmm ffxml writing

swails:
Move unsupported error catch to script function, since that's where it will
ultimately be implemented. Also fix 1-4 scaling factor determination so it is
not hard-coded for Amber (and works with any homogenous force field with an
appropriate error message for non-homogenous FFs, like Amber/GLYCAM).

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -363,7 +363,7 @@ class OpenMMParameterSet(ParameterSet):
                        (name, sigma, epsilon))
         dest.write(' </NonbondedForce>\n')
 
-    def _write_omm_scripts(self):
+    def _write_omm_scripts(self, dest):
         # Not currently implemented, so throw an exception if any unsupported
         # options are specified
         if self.combining_rule == 'geometric':
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -147,7 +147,7 @@ class OpenMMParameterSet(ParameterSet):
             self._write_omm_impropers(dest)
 #           self._write_omm_rb_torsions(dest)
             self._write_omm_cmaps(dest)
-#           self._write_omm_scripts(dest)
+            self._write_omm_scripts(dest)
             self._write_omm_nonbonded(dest)
         finally:
             dest.write('</ForceField>\n')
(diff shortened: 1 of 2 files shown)


jchodera:
Cool!

Does it really make sense to split nonbonded writing into _write_omm_scripts and _write_omm_nonbonded, since both of these together write the nonbonded stuff?

swails:
> Does it really make sense to split nonbonded writing into _write_omm_scripts and _write_omm_nonbonded, since both of these together write the nonbonded stuff?

IMO yes, because the scripts do more than just nonbonded stuff.  It's the general garbage cleaner that mops up all the loose ends and corner cases that `ForceField` itself is incapable of supporting "natively".  The things I need to add to `_write_omm_scripts` are:

- Support for Urey-Bradley terms
- NBFIX support
- Geometric combining rule support
- Mixed 1-4 scaling for dihedrals (e.g., for GLYCAM and Amber combinations)

jchodera:
Do we need to combine all these separate things into one function? We could
have the responsible forces each write a separate script block, can't we?

I just worry about things being less modular and easier to mess up if we
split the functionality for several forces across multiple functions.

swails:
I was probably going to break up each of those tasks into separate private functions, but all called from inside `_write_omm_script`.  I don't really want to write multiple `<script></script>` sections in the ffxml file for a couple reasons: I don't know if more than one per file is even supported, and I don't want there to be any ambiguity as to when each segment runs and any differences in context between when different `<script>` segments run.  But in order for only one script to be written to each ffxml file, you *need* to write all sections of it at the same time, which is why I think it makes sense to bundle that up in a single private function.

jchodera:
I think @peastman said that it was totally fine to have multiple script
sections.

We shouldn't be writing script sections that mess up context, ideally!

What if we first tried having each force be responsible for its own script
section, if needed, and then if that doesn't work tackle the monolithic
script writer? I just worry about things being quite complex if we have to
write a monolithic script writer.

swails:
> What if we first tried having each force be responsible for its own script
> section, if needed, and then if that doesn't work tackle the monolithic
> script writer? I just worry about things being quite complex if we have to
> write a monolithic script writer.

We can always shuffle around the function calls later.  But I'm going to merge this, since script writing hasn't started yet (and what is there now is a placeholder that does not impose any decisions currently).

jchodera:
OK, no problem!

jchodera:
What's next for us to do?  Maybe focus on those tests?

swails:
> What's next for us to do? Maybe focus on those tests?

That's a good next step, I think.  But only for the Amber force field, since CHARMM requires a script.  So take a handful of systems, run them through LEaP and OpenMM, and make sure the resulting systems give identical energies.

PR #525, opened 2016-01-06
Add some tests and fix a very minor mol2 parser bug for multi-residue structures

swails:


swails:
The failures are all caused by unreliable file downloads from Github on some of the builds.  I've noticed some lagging from GH recently, so this should hopefully resolve itself within time.  Merging despite the failed status.

PR #528, opened 2016-01-08
Fix sdist generation

rmcgibbo:
https://github.com/pandegroup/openmm/issues/680#issuecomment-170136497


rmcgibbo:
I'm not actually sure if you need the tests and examples and stuff in the sdist, just the `src` should be sufficient for the bug you saw with the headers.

swails:
Thanks!

PR #532, opened 2016-01-11
_write_omm_nonbonded fix key/value bug

rafwiewiora:
Was iterating through keys, but Dihedral Types lists are values in the `dihedral_types` dict,


swails:
How did this ever pass the test??  Thanks for fixing this.

PR #540, opened 2016-01-13
Urey Bradley term

ChayaSt:
I copied @jchodera's ```_write_omm_urey_bradley``` here.

### DIFF
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -330,7 +330,7 @@ class CharmmParameterSet(ParameterSet):
             if line.startswith('ANGLE') or line.startswith('THETA'):
                 section = 'ANGLES'
                 continue
-            if line.startswith('DIHEDRAL') or line.startswith('PHI') or line.startswith('DIHE'):
+            if line.startswith('DIHE') or line.startswith('PHI'): 
                 section = 'DIHEDRALS'
                 continue
             if line.startswith('IMPROPER') or line.startswith('IMPHI'):
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -330,7 +330,7 @@ class CharmmParameterSet(ParameterSet):
             if line.startswith('ANGLE') or line.startswith('THETA'):
                 section = 'ANGLES'
                 continue
-            if line.startswith('DIHEDRAL') or line.startswith('PHI'):
+            if line.startswith('DIHEDRAL') or line.startswith('PHI') or line.startswith('DIHE'):
                 section = 'DIHEDRALS'
                 continue
             if line.startswith('IMPROPER') or line.startswith('IMPHI'):
@@ -444,6 +444,7 @@ class CharmmParameterSet(ParameterSet):
                     theteq = conv(words[4], float, 'angle equilibrium value')
                 except IndexError:
                     raise CharmmError('Could not parse angles.')
+
                 angle_type = AngleType(k, theteq)
                 self.angle_types[(type1, type2, type3)] = angle_type
                 self.angle_types[(type3, type2, type1)] = angle_type
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -12,7 +12,7 @@ from parmed.formats.registry import FileFormatType
 from parmed.modeller.residue import ResidueTemplate
 from parmed.parameters import ParameterSet
 from parmed.periodic_table import Element
-#from parmed.topologyobjects import NoUreyBradley
+from parmed.topologyobjects import NoUreyBradley
 from parmed import unit as u
 from parmed.utils.io import genopen
 from parmed.utils.six import add_metaclass, string_types, iteritems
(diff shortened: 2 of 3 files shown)


swails:
Oh clever!  I thought we were going to have to put the Urey-Bradley terms inside the `<script>` section [like @peastman did for the CHARMM polarizable force field](https://raw.githubusercontent.com/pandegroup/openmm/master/wrappers/python/simtk/openmm/app/data/charmm_polar_2013.xml) (search for Urey-Bradley in the link).  This is a much better approach!

I also use the Amoeba force field all the time, so I know that the U-B terms are implemented as `HarmonicBondForce`, which is the correct behavior for the CHARMM force field (as is clear [here](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L4451))

jchodera:
Turns out the Amoeba Urey-Bradley term is the same as the CHARMM one (even up to the same weird definition of force constant).  It's a shame they called this `AmoebaUreyBradley`.

swails:
> Turns out the Amoeba Urey-Bradley term is the same as the CHARMM one

I know!  Never even occurred to me before, though.

> even up to the same weird definition of force constant

Not sure what this means...  It's implemented as a `HarmonicBondForce`, so it's just a simple `k(r-req)**2` functional form.  That's how all Urey-Bradley terms I'm aware of behave.

So now the `<script>` only needs to handle 1-4 scaling modifications, geometric combining rules, and NBFIXes AFAICT.  This is nice, since it does all seem to be related to nonbonded stuff.

jchodera:
> k(r-req)**2

A force constant is almost universally defined as "the constant that multiplies the deviation to get the force".  That is, in order to get
```
F = k (r-req)
```
where `k` is the "force constant", we need
```
U = (k/2) (r-req)**2
```

jchodera:
I guess I could call it a "half-force-constant"...

parmed/openmm/parameters.py
```diff
@@ -286,6 +287,23 @@ def nowild(name):
...
+        length_conv = u.angstroms.conversion_factor_to(u.nanometers)
+        _ambfrc = u.kilocalorie_per_mole/u.angstrom**2
+        _ommfrc = u.kilojoule_per_mole/u.nanometer**2
+        frc_conv = _ambfrc.conversion_factor_to(_ommfrc)
+        ureys_done = set()
+        for (a1, a2, a3), urey in iteritems(self.urey_bradley_types):
+            if (a1, a2, a3) in ureys_done: continue
+            if urey == NoUreyBradley: continue
+            dest.write('  <UreyBradley type1="%s" type2="%s" type3="%s" d="%f" k="%f"/>\n'
+                       % (a1, a2, a3, urey.req*length_conv, urey.k*frc_conv))
```
swails: Is `frc_conv` missing a factor of 2 here?  ParmEd stores bond type force constants according to the functional form `k(r-req)^2`, whereas OpenMM in most places uses `1/2*k*(r-req)^2`

swails:
Oh.  So ignore my last in-line comment then.

swails:
Looks good!

parmed/charmm/parameters.py
```diff
@@ -330,7 +330,7 @@ def read_parameter_file(self, pfile, comments=None):
             if line.startswith('ANGLE') or line.startswith('THETA'):
                 section = 'ANGLES'
                 continue
-            if line.startswith('DIHEDRAL') or line.startswith('PHI'):
+            if line.startswith('DIHEDRAL') or line.startswith('PHI') or line.startswith('DIHE'):
```
swails: Let's just change DIHEDRAL to DIHE. There's no reason to check for both.

parmed/openmm/parameters.py
```diff
@@ -372,6 +390,6 @@ def _write_omm_scripts(self, dest):
                                       'supported.')
         if len(self.nbfix_types) > 0:
             raise NotImplementedError('NBFIX not currently supported')
-        if self.urey_bradley_types:
-            raise NotImplementedError('Urey-Bradley angles not currently '
-                                      'supported.')
+        # if self.urey_bradley_types:
+        #     raise NotImplementedError('Urey-Bradley angles not currently '
+        #                               'supported.')
```
swails: Let's just delete these lines.  No reason to keep them commented out, since there's no reason to ever un-comment them back in.

swails:
After the two changes I described above, looks good to merge!

ChayaSt:
Done.

parmed/charmm/parameters.py
```diff
@@ -330,7 +330,7 @@ def read_parameter_file(self, pfile, comments=None):
             if line.startswith('ANGLE') or line.startswith('THETA'):
                 section = 'ANGLES'
                 continue
```
hainm: @swails is `continue` necessary in here (and others)?

swails: It's a long loop body (which is probably a sign that it should be refactored at some point). So that's safest. 

But yes, I think it's necessary here. Processing each section occurs in a separate set of if/else blocks than processing the headers, so when a header is processed it needs to skip the parameter parsing logic (since its on a header line, not a parameter line).

swails:
Great! Thanks!

PR #541, opened 2016-01-14
Respect `parametrize` flag when reading .top files

ctk3b:
@swails are you in the process of moving over all your testing functions names to snake? I can swap that out real quick too.

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -870,7 +870,8 @@ class GromacsTopologyFile(Structure):
             else:
                 raise GromacsError("Can't add %d %s molecules" % (num, molname))
         self.itps = itplist
-        self.parametrize()
+        if parametrize:
+            self.parametrize()
 
     #===================================================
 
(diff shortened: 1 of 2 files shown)


swails:
By snake you mean names_with_underscores? I've never heard that term before...

swails:
But if that's what you mean by snake then yes, I'm making those test function name changes. 

If you wanted to change the function names that'd be great, but certainly don't feel like you have to. I'd be happy to merge this now if you don't want to put more time into this at the moment. 

Thanks!

ctk3b:
Hmm for some reason I thought I learned that term from PEP8. Either way it does appear to be a [thing](http://www.wikiwand.com/en/Snake_case) :)

Should be good to merge.

swails:
Great! Thanks!

swails:
By the way, let me know when InterMol parameter type is done to where I should start trying to add it to ParmEd. I'm following your commits a little, just give me a ping there when it's wrapping up a little more. (Or is it already?)

ctk3b:
The shift to parametertypes is in. Gromacs and Lammps work pretty well. Desmond needs some work. 

It's probably worth trying to write a basic conversion function at this point.

ctk3b:
Let me know if you want me to give it a try first - I'm getting pretty comfortable with parmed now too. Might also be worth having a quick chat at some point in the next week(s) with @mrshirts

Issue #544, opened 2016-01-14
Reading charges in from Gromacs [ atomtypes ]


ctk3b:
Currently the `charge` field of Gromacs atomtypes [is ignored](https://github.com/ParmEd/ParmEd/blob/master/parmed/gromacs/gromacstop.py#L667). This information would be useful to me for working with OPLS but `charge` is currently not an attribute of `AtomType` objects.

Would you be opposed to supporting this? It's a bit of a weird issue since Gromacs ignores this information unless masses and charges are not specified in `[ atoms ]`. It's unclear to me what the safest way to support this is but the three scenarios I can see are:

1. Simply store the value in `AtomType` - anyone who wants to can manually set the `Atom`'s charge from its `AtomType`
2. Add a flag to `write` which suppresses writing of charges and masses in the `[ atoms ]` directive such that these values are pulled from `[ atomtypes ]` by Gromacs
3. Automatically suppress writing of charges and masses if a) charges are all zero and b) `atom.mass` == `atom.type.mass` (or some criteria like this that we deem are safe to assume)

I'm a bit wary of option 3 but I think the first two should be relatively safe and non-intrusive to current behavior.

swails:
Check out #545 -- I basically did for charge what was already being done for the L-J parameters.  Basically, it can be stored in AtomType, and that will be applied to any atom it is assigned to if that atom has not explicitly set its own parameters.

However, each atom can set its own specific L-J and charge parameters independent of its assigned atom type.  So now reading OPLS should Just Work.

When it comes to writing, what is the argument here for *not* writing charges to the `[ atoms ]` directly (and simply don't care about the charge that gets written to `[ atomtypes ]`?  I'd like to limit API complexity here, but also balance that against implementing useful features.  So if there's a good reason to implement your second point above for writing GROMACS topology files, I'd be fine with it.

I'd appreciate any comments you may have on #545, though (and whether that even works for you).

Thanks!

ctk3b:
The approach in #545 fits the bill entirely.

swails:
Great.  I tidied up that PR (hopefully fixing the failing tests), then I'll bump the version and merge.

PR #545, opened 2016-01-15
Adds a charge attribute to AtomType

swails:
The basic behavior here is that AtomType *can* have a charge, and if an `Atom` has an `atom_type` attribute, but never has a charge set, it will see the charge from its `atom_type`.  However, setting a charge on an `Atom` directly overrides its `atom_type`.

GROMACS parsers now assign the parsed charge to its atom type when reading `[ atom_types ]`.  This should address #544.

@ctk3b -- does this work for you?  What do you think of it?

### DIFF
--- a/parmed/__init__.py
+++ b/parmed/__init__.py
@@ -6,7 +6,7 @@ between standard and amber file formats, manipulate structures, etc.
 # Version format should be "major.minor.patch". For beta releases, attach
 # "-beta#" to the end. The beta number will be turned into another number in the
 # version tuple
-__version__ = '2.1.4'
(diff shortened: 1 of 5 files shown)


ctk3b:
I think this looks perfect - I might have overthought it slightly. As you mentioned in #544, I don't see any particular reason not to write these charges.

swails:
Excellent, I'll go ahead and merge this, then.

I think this is also a nice opportunity to upload to PyPI (which would conveniently make ParmEd pip-installable).

ctk3b:
Nice!

PR #548, opened 2016-01-15
Write charge to `[ atomtypes ]`

ctk3b:
Not really important but may as well after #545

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -1364,7 +1364,7 @@ class GromacsTopologyFile(Structure):
                 if print_atnum:
                     parfile.write('%8d ' % atom_type.atomic_number)
                 parfile.write('%10.5f  %10.6f  A %13.6g %13.6g\n' % (
-                              atom_type.mass, 0, atom_type.sigma/10,
+                              atom_type.mass, atom_type.charge, atom_type.sigma/10,
                               atom_type.epsilon*econv))
             parfile.write('\n')
             # Print all parameter types unless we asked for inline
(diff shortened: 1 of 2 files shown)


swails:
I'm a little concerned about setting a default `AtomType.charge` to 0.  The reason I picked `None` is because that indicates that the charge was never set on the `AtomType`.  If it defaults to 0, there's no way to tell if it was never set, or if it's simply a neutral atom.

I'm not sure this ambiguity is necessarily bad, though, since `Atom.atom_type` is always the fallback place to look for L-J and charge parameters...  I think I'm good to just merge this, though.

ctk3b:
Yeah it does slightly change behavior. Alternatively, I can check in the gromacs writer if it's `None` and if so just write `0.0`.

parmed/gromacs/gromacstop.py
```diff
@@ -1364,7 +1364,7 @@ def write(self, dest, combine=None, parameters='inline'):
                 if print_atnum:
                     parfile.write('%8d ' % atom_type.atomic_number)
                 parfile.write('%10.5f  %10.6f  A %13.6g %13.6g\n' % (
-                              atom_type.mass, 0, atom_type.sigma/10,
+                              atom_type.mass, atom_type.charge, atom_type.sigma/10,
```
swails: If you change `atom_type.charge` to `atom_type.charge or 0.0` that will handle the case where `atom_type.charge` is `None`

swails:
Although you run into the same issues if you don't set the L-J parameters.  I think this is fine as-is.  I'll go ahead and merge it.

PR #549, opened 2016-01-15
update write_pdb doc: move default value  

hainm:


parmed/formats/pdb.py
```diff
@@ -611,21 +611,21 @@ def write(struct, dest, renumber=True, coordinates=None, altlocs='all',
...
-            it was assigned previously. Default is True
+            it was assigned previously.
         coordinates : array-like of float, optional
             If provided, these coordinates will be written to the PDB file
             instead of the coordinates stored in the structure. These
             coordinates should line up with the atom order in the structure
             (not necessarily the order of the "original" PDB file if they
             differ)
-        altlocs : str, optional
+        altlocs : str, optional, defaul 'all'
```
swails: default

hainm: updated

swails:
Thanks!

PR #553, opened 2016-01-18
merge master

swails:
Merges master branch into ommparams


swails:
This PR now includes an optional `splitDihedral` argument to `Structure.createSystem` that will split dihedrals into proper and improper forces.

PR #562, opened 2016-01-20
Detect element when reading .mol2 files

ctk3b:
Same as what's done in the .gro reader. This is obviously not robust so let me know if this is a particular concern for this reader.

### DIFF
--- a/parmed/formats/mol2.py
+++ b/parmed/formats/mol2.py
@@ -9,6 +9,7 @@ from parmed.exceptions import Mol2Error
 from parmed.formats.registry import FileFormatType
 from parmed.modeller.residue import ResidueTemplate, ResidueTemplateContainer
 from parmed.residue import AminoAcidResidue, RNAResidue, DNAResidue
+from parmed.periodic_table import element_by_name, AtomicNum
 from parmed.structure import Structure
 from parmed.topologyobjects import Atom, Bond
 from parmed.utils.io import genopen
@@ -149,6 +150,8 @@ class Mol2File(object):
                     words = line.split()
                     id = int(words[0])
                     name = words[1]
+                    elem = element_by_name(name)
+                    atomic_number = AtomicNum[elem]
                     x = float(words[2])
                     y = float(words[3])
                     z = float(words[4])
@@ -171,7 +174,8 @@ class Mol2File(object):
                     if last_residue is None:
                         last_residue = (resid, resname)
                         restemp.name = resname
-                    atom = Atom(name=name, type=typ, number=id, charge=charge)
+                    atom = Atom(name=name, atomic_number=atomic_number,type=typ,
+                                number=id, charge=charge)
                     atom.xx, atom.xy, atom.xz = x, y, z
                     struct.add_atom(atom, resname, resid)
                     if last_residue != (resid, resname):
--- a/test/test_parmed_formats.py
+++ b/test/test_parmed_formats.py
@@ -133,6 +133,7 @@ class TestFileLoader(FileIOTestCase):
         for atom in mol2.atoms:
             self.assertEqual(atom.charge, 0)
             self.assertEqual(atom.residue.name, 'UNK')
+            self.assertNotEqual(atom.atomic_number, 0)
         # Check mol2 where last several columns do not exist in SUBSTRUCTURE
         mol2 = formats.load_file(get_fn('tripos4.mol2'), structure=True)
         self.assertIsInstance(mol2, Structure)


davidlmobley:
I don't have time to read the code in detail right now to figure out what you're doing, but it looks like you're getting the element from the atom name, which makes me very nervous. This will work for *some* types of `.mol2` files (specifically, Tripos atom types) but as far as I know, not others, unless I'm missing something. For example, "CA" (or "ca") and "Ca" are two very different things when dealing with AMBER or GAFF atom types. If you do this you're gonna need to make sure you're working with TRIPOS atom types or problems will result down the line.

davidlmobley:
Probably if this passes checks, you should update the tests to include a test which would address the issue I just raised.

davidlmobley:
Also, if this is what's done in the `.gro` reader then we may need to visit that as well...?

ctk3b:
> it looks like you're getting the element from the atom name, which makes me very nervous. 

Correct and I agree. Jason mentioned the issues in a [comment](https://github.com/mdtraj/mdtraj/issues/909#issuecomment-130400843) a while ago.

This, like the `.gro` reader, provides a VERY naive best guess. If you want to remain more conservative, I would agree with revisiting the decision made in that reader.

hainm:
this requires massive testing.

davidlmobley:
I am much more strongly opposed to using names to infer elements from `.mol2` files than from `.pdb` files. This is primarily because `.pdb` structures are most typically used for proteins or other large biomolecules where it's less likely to use "weird" atom names, whereas the main reason to use a `.mol2` file is for small molecules. And, at least within AMBER, the most commonly used `.mol2` files DO and ARE SUPPOSED TO have weird atom names (AMBER/GAFF names). For example, if you run `antechamber` on a `.mol2` file and don't tell it otherwise, it will assign non-Sybyl atom names. So if you build this in to the .mol2 reader you're building in a situation where the default behavior on a naive application of antechamber and ParmEd to a "typical" `.mol2` file will yield incorrect elements in ParmEd. *Very bad*, if you ask me.

Unless you have a better solution to the problem (i.e., check what atom types are being used) I really don't think the `.mol2` reader should be doing this, and I can't imagine the AMBER community would think this is a good idea either (since they're the main thing driving the use of non-Sybyl types in these `.mol2` files).

ctk3b:
Fair enough. I would vote for removing this logic from the `.gro` reader then as well.

davidlmobley:
That's fine for me. I don't think ANY systems my group simulates would
actually process correctly with this logic, though I would know to expect
problems when reading `.pdb` files and would NOT expect problems reading
`.mol2` files, especially from tools somewhat associated with the AMBER
package (as ParmEd is).

On Wed, Jan 20, 2016 at 11:50 AM, Christoph Klein <notifications@github.com>
wrote:

> Fair enough. I would vote for removing this logic from the .gro reader
> then as well.
>
> —
> <https://github.com/ParmEd/ParmEd/pull/562#issuecomment-173338873>.
>



-- 
David Mobley
dmobley@gmail.com
949-385-2436

swails:
Yea, it's unfortunate that some file formats seem to think that element identity is unimportant :).  Figuring out atoms by name is messy, and ParmEd's "guess element by name" functionality is *very* heavily-weighted toward biomolecular applications.  For instance, `CA` is always interpreted as a carbon (meaning calcium will never match).  Every amino acid has a CA atom, after all...  This heuristic could be improved, [like doing what OpenMM does](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/pdbfile.py#L127-L147), which is basically special-case calcium and assign CA to calcium if and only if the CA atom is a single-atom residue.  This type of logic maps better to a GRO file than a mol2 file.

In many cases, this assignment is better than nothing, though.  The current behavior is to assign the atomic number to -1 (atomic number of 0 means a virtual site) -- -1 means "I don't know".

I think the GRO file is PDB-like enough that the name guessing is a net positive.  If the GRO file is only used as a coordinate file, like via `pmd.load_file('gmx.top', xyz='gmx.gro')`, then the element information is taken from the more reliable GROMACS topology.  But the GRO parser will still more likely than not get the element correct.

However, as @davidlmobley pointed out, this heuristic is far less reliable for small-molecule stuff like mol2 files.  If we had a good heuristic to assign elements, it would be nice to apply that here.  But I don't think the current `element_from_name` satisfies the bill for mol2 files.

ctk3b:
Sounds good!

swails:
A basic use-case for the GRO file is to determine the approximate molar mass of a system, which has no mass information present.  For a large-ish biomolecule with a couple calcium and chloride ions (which are known to be misassigned), the error in the total mass will be a small percentage of the total mass of the system.  So you can still get useful information from it even if not all atoms have correctly-assigned elements.

The potential risk is if people *think* the information is reliable when it's not.  But I only have my own perspective on what's a worthwhile risk or acceptable tradeoff, so I need to hear other peoples' opinions on what is ideal behavior.

For instance, in this case, I'd say the mol2 element assignment by name is fine -- sure there's a likelihood it will be wrong that is higher than other formats, but if you know that in advance then you can account for that before deciding whether or not to use the element assignments.  But more people spoke out in opposition to that stance than in favor, which was helpful for me to hear.

PR #576, opened 2016-01-27
Use setuptools when available

swails:
This should play nicer with pip (distutils builds are deprecated when installing through pip).  This also lets me use entry points (rather than CL scripts), which will behave correctly on Windows (unlike distutils builds).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,4 +1,15 @@
-from distutils.core import setup, Extension
+try:
+    from setuptools import setup, Extension
+    kws = dict(entry_points={
+            'console_scripts' : ['parmed = parmed.scripts:clapp'],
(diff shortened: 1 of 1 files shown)


hainm:
you let the build fail if coverage decreases?

hainm:
and do you know how to display code coverage yet

(like this one: https://coveralls.io/files/1892190484)

swails:
I turned coverage on and got it at least running, but at the moment I'm just ignoring it :). There is one big thing that I currently don't like about it -- it doesn't combine the coverage reports from each build into a single, comprehensive report.

I'm still figuring out how it works and how to get it to do what I want.  But to answer your questions:

> you let the build fail if coverage decreases?

I have no idea :smile: 

> and do you know how to display code coverage yet?

No, but I don't want to display it yet until I get it where I want it.

PR #579, opened 2016-01-27
Warning when overwriting bonds and angles

ChayaSt:
I added a warning when overwriting bonds and angles to find different values for the same parameters. Is this something we should add? I can modify this so that users can have an option if they want to see the warning.


hainm:
>  I can modify this so that users can have an option if they want to see the warning.

I would go this way.

parmed/charmm/parameters.py
```diff
@@ -430,8 +430,17 @@ def read_parameter_file(self, pfile, comments=None):
...
                 key = (min(type1, type2), max(type1, type2))
                 bond_type = BondType(k, req)
-                self.bond_types[(type1, type2)] = bond_type
-                self.bond_types[(type2, type1)] = bond_type
+                if key in self.bond_types:
+                    # See if the existing bond type list has a different value and replaces it with a warning
+                    if self.bond_types[key] != bond_type:
+                        # Replace. Warn if they are differen
+                        warnings.warn('Replacing bond %r, %r with %r' %
+                                              (key, self.bond_types[key], bond_type))
```
swails: Warn with a `ParameterWarning` (which, if it's not already imported, is found in `parmed.exceptions`)

parmed/charmm/parameters.py
```diff
@@ -446,8 +455,17 @@ def read_parameter_file(self, pfile, comments=None):
...
 
                 angle_type = AngleType(k, theteq)
-                self.angle_types[(type1, type2, type3)] = angle_type
-                self.angle_types[(type3, type2, type1)] = angle_type
+                if (type1, type2, type3) in self.angle_types:
+                    # See if the existing angle type list has a different value and replaces it with a warning
+                    if self.angle_types[(type1, type2, type3)] != angle_type:
+                        # Replace. Warn if they are differen
+                        warnings.warn('Replacing angle %r, %r with %r' %
+                                              ([(type1, type2, type3)], self.angle_types[(type1, type2, type3)], angle_type))
```
swails: Warn with `ParameterWarning`.  Also, can you try cleaning up the formatting a bit so the lines don't run quite as long?

parmed/charmm/parameters.py
```diff
@@ -485,7 +503,7 @@ def read_parameter_file(self, pfile, comments=None):
                         if dtype.per == dihedral.per:
                             # Replace. Warn if they are different
                             if dtype != dihedral:
-                                warnings.warn('Replacing dihedral %r with %r' % 
+                                warnings.warn('Replacing dihedral %r with %r' %
```
swails: Same -- `ParameterWarning`

swails:
The way I usually turn of warnings is with `warnings.filterwarnings`.  Since CHARMM behavior in this case is undefined/undetermined, these warnings should be loud.  (It should not warn at all in the Amber variant, since that's how Amber works).

swails:
The tests had all failed because GH was down yesterday and the file download tests failed.  Rerunning now, and I'll merge when those are finished.

swails:
Do you have a test case you can add that triggers this *should* trigger this warning, just to make sure that it will?

ChayaSt:
Yes. I'll add those.

swails:
OK, ping me when you do (GH doesn't email me when commits get added to PRs)

ChayaSt:
Done.

swails:
Looks good.  Thanks.

PR #581, opened 2016-01-28
Add a bunch of tests to the openmm package.

swails:
Also fixes a couple small issues:

- passing an open file handle (instead of a file name) to XmlFile.parse failed
- atom type compression for L-J types was not implemented correctly, although
  the resulting systems were correct from a FF perspective. The only side-effect
  is that a Structure loaded from an OpenMM System might have had nonbonded type
  tables that were larger than they needed to be (and as a result might have had
  GPU performance implications for large systems).

### DIFF
--- a/parmed/openmm/topsystem.py
+++ b/parmed/openmm/topsystem.py
@@ -21,10 +21,6 @@ from parmed import unit as u
 from parmed.utils.decorators import needs_openmm
 from parmed.utils.six import iteritems, string_types
 from parmed.utils.six.moves import range
-try:
-    import simtk.openmm as mm
-except ImportError:
-    mm = None
 import warnings
 
 @needs_openmm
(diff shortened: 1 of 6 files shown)


parmed/openmm/topsystem.py
```diff
@@ -230,7 +227,7 @@ def _process_angle(struct, force):
 def _process_urey_bradley(struct, force):
     """ Adds Urey-Bradley parameters to the structure """
     if not struct.angles:
-        warnings.warn('Adding what seems to be Urey-Bradley terms before '
+        warnings.warn('Adding what seems to be Urey-Bradley terms before ' # pragma: no cover
```
hainm: do you ignore all warnings for coverage? if so, you can add the stuff to .coverc config:

https://github.com/hainm/ParmEd/blob/71e158f731c69a5007e9a507ad304a710cb059d4/test/.coveragerc

swails: No, I actually try to trigger all warnings (and most exceptions) in my tests.  This particular one is not important to trigger, though, so I decided to omit it rather than invent a test for it.

Issue #580, opened 2016-01-27
AmberParameterSet.from_leaprc fails to recognize lone pairs/extra points


swails:
[See the comment here](https://github.com/ParmEd/ParmEd/issues/486#issuecomment-175890223) in #486

rafwiewiora:
I thought saying 'fixes #number' in a PR would close this?

rafwiewiora:
Well good to close anyway!

Issue #486, opened 2015-12-03
Amber -> OpenMM force field conversion


swails:
See pandegroup/openmm#1281

rafwiewiora:
Some previous discussion for reference: https://github.com/pandegroup/openmm/issues/1184

jchodera:
We discussed this at length today, and came up with a strategy that we hope might work.  We wanted to discuss it here before they got started:
* @rafwiewiora is heading up the AMBER conversion; @ChayaSt is heading up the CHARMM conversion
* The first step is for each of them to load in their respective parameter sets of interest into ParmEd to make sure all files they want to convert are readable.  For @rafwiewiora this is currently reading the various sets of `.leaprc` files in AmberTools15 using `parmed.amber.parameters.from_leaprc`.  For @ChayaSt, this is the recent set of [CHARMM forcefield files from the Mackerell website](http://mackerell.umaryland.edu/charmm_ff.shtml).
* Once they have verified they can read their respective parameter sets, they will work together to implement a `ParameterSet` method called `create_omm_ffxml(filename)` (or whatever name you see fit) that, by analogy to `Structure.createSystem`, generates OpenMM-style `ffxml` files.
* The first section to be converted is the `<AtomTypes>` section, where we will take advantage of the recent extensions to the `ffxml` format.  For type `name`s, we will use the AMBER atom type names.  For type `class`es, we will use the AMBER Lennard-Jones atom classes.  This will reduce the number of atom types from thousands to ~60.  
* Next, we will include `<Residues>` definitions that specify per-atom `charge` attributes and make use of the new atom name capability in `<Bond>` and `<ExternalBond>` records
* Next, they will add `<Force>` terms, one at a time.  `<NonbondedForce>` will use the [`<UseAttributeFromResidue>` capability](https://github.com/pandegroup/openmm/pull/1182) to specify that `charge` will be taken from atom attributes.
* I've suggested that, in analogy with `Structure`'s decomposition of the `createSystem` function into `omm_bond_force`, `omm_angle_force`, etc., they break up the writing of force terms into `omm_bond_ffxml`, `omm_angle_ffxml`, etc.
* They will write tests that go along with this and test progressively.
* They will do this progressively in a PR so we can all comment on the progress.

Any other wisdoms to impart?

swails:
[...]

PR #582, opened 2016-01-29
addAtomTypes only updates atomic number info for types existing in .dat/frcmod // hard-code EP and LP

rafwiewiora:
As we discussed in https://github.com/ParmEd/ParmEd/issues/486 . 

Not sure if this is the best approach here, what do you think?

### DIFF
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -211,7 +211,7 @@ class AmberParameterSet(ParameterSet):
     #===================================================
 
     @classmethod
-    def from_leaprc(cls, fname):
+    def from_leaprc(cls, fname, addAtomTypes=True):
         """ Load a parameter set from a leaprc file
 
         Parameters
@@ -219,6 +219,10 @@ class AmberParameterSet(ParameterSet):
         fname : str or file-like
             Name of the file or open file-object from which a leaprc-style file
             will be read
+        addAtomTypes : bool
+            If False, the addAtomTypes section of the leaprc is ignored
+            (useful when one does not want to create extra AtomType's which may
+            be called in the leaprc, but not defined in the parameter files)
 
         Notes
         -----
@@ -275,14 +279,15 @@ class AmberParameterSet(ParameterSet):
                 i += 1
             atom_types_str = ''.join(chars).strip()
         for _, name, symb, hyb in _atomtypere.findall(atom_types_str):
-            if symb not in AtomicNum:
-                raise ParameterError('%s is not a recognized element' % symb)
-            if name in params.atom_types:
-                params.atom_types[name].atomic_number = AtomicNum[symb]
-            else:
-                params.atom_types[name] = \
-                        AtomType(name, len(params.atom_types)+1, Mass[symb],
-                                 AtomicNum[symb])
+            if addAtomTypes:
+                if symb not in AtomicNum:
+                    raise ParameterError('%s is not a recognized element' % symb)
+                if name in params.atom_types:
+                    params.atom_types[name].atomic_number = AtomicNum[symb]
+                else:
+                    params.atom_types[name] = \
+                            AtomType(name, len(params.atom_types)+1, Mass[symb],
+                                     AtomicNum[symb])
         # Now process the parameter files
         for fname in _loadparamsre.findall(text):
             params.load_parameters(_find_amber_file(fname))
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -242,13 +242,21 @@ class AmberParameterSet(ParameterSet):
         text = f.read()
         if own_handle: f.close()
         lowertext = text.lower() # commands are case-insensitive
+        # Now process the parameter files
+        for fname in _loadparamsre.findall(text):
+            params.load_parameters(_find_amber_file(fname))
+        # Now process the library file
+        for fname in _loadoffre.findall(text):
+            params.residues.update(
+                    AmberOFFLibrary.parse(_find_amber_file(fname))
+            )
+        # Now process the addAtomTypes
         try:
             idx = lowertext.index('addatomtypes')
         except ValueError:
             # Does not exist in this file
             atom_types_str = ''
         else:
-            # Now process the addAtomTypes
             i = idx + len('addatomtypes')
             while i < len(text) and text[i] != '{':
                 if text[i] not in '\r\n\t ':
@@ -279,19 +287,6 @@ class AmberParameterSet(ParameterSet):
                 raise ParameterError('%s is not a recognized element' % symb)
             if name in params.atom_types:
                 params.atom_types[name].atomic_number = AtomicNum[symb]
-            else:
-                params.atom_types[name] = \
-                        AtomType(name, len(params.atom_types)+1, Mass[symb],
-                                 AtomicNum[symb])
-
-        # Now process the parameter files
-        for fname in _loadparamsre.findall(text):
-            params.load_parameters(_find_amber_file(fname))
-        # Now process the library file
-        for fname in _loadoffre.findall(text):
-            params.residues.update(
-                    AmberOFFLibrary.parse(_find_amber_file(fname))
-            )
         return params
 
     #===================================================
@@ -540,6 +535,9 @@ class AmberParameterSet(ParameterSet):
             raise ParameterError('Error parsing MASS line. Not enough tokens')
         if words[0] in self.atom_types:
             self.atom_types[words[0]].mass = mass
+        elif words[0] in ('EP', 'LP'):
+            atype = AtomType(words[0], len(self.atom_types)+1, mass, 0)
+            self.atom_types[words[0]] = atype
         else:
             atype = AtomType(words[0], len(self.atom_types)+1, mass,
                              AtomicNum[element_by_mass(mass)])
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -278,8 +278,8 @@ class AmberParameterSet(ParameterSet):
                 chars.append(char)
                 i += 1
             atom_types_str = ''.join(chars).strip()
-        for _, name, symb, hyb in _atomtypere.findall(atom_types_str):
-            if addAtomTypes:
+        if addAtomTypes:
+            for _, name, symb, hyb in _atomtypere.findall(atom_types_str):
                 if symb not in AtomicNum:
                     raise ParameterError('%s is not a recognized element' % symb)
                 if name in params.atom_types:
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -211,7 +211,7 @@ class AmberParameterSet(ParameterSet):
     #===================================================
 
     @classmethod
-    def from_leaprc(cls, fname, addAtomTypes=True):
+    def from_leaprc(cls, fname):
         """ Load a parameter set from a leaprc file
 
         Parameters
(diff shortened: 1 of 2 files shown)


rafwiewiora:
Thinking about this more, I'm no longer sure this is necessary.. 

The use case here would be to have this set to False in the conversion script that will be shipped with OpenMM - if people want to convert a leaprc, we won't consider the addAtomTypes section to avoid potential Atom Type clashes. 

But then - if you're not going to manually modify that leaprc, you will run into trouble anyway most likely, because the leaprc will have ions and solvent parameters and that will cause clashes etc. So you might as well rely on deleting the addAtomTypes from the leaprc if you don't want them.

Well let's see what you think!

rafwiewiora:
Also this is assuming that the element recognition works well by mass - it seems to. But maybe it'd be better to not throw away the element information and when the .dat is loaded, go back and delete those atom types that were not declared in the .dat?

swails:
Can you check that `atomic_number` still gets set correctly on the `AtomType` objects in `atom_types`.  Because if it doesn't, I don't think we should ever ignore this completely.

swails:
Oh, I looked at a really old window where your last comment didn't pop up.  Since `atomic_number` seems to be set from the mass by default (thinking about it more, I think it *must* be, since I added parm.dat/frcmod parsing before leaprc parsing), then addAtomTypes really doesn't do anything.  I'd almost suggest just commenting out that whole segment and just identifying virtual sites by name (EP or LP, since Amber hard-codes for those two).

rafwiewiora:
Looks like it all works fine: as long as the recognition of the element by mass works ok, the atomic number flows from that: https://github.com/ParmEd/ParmEd/blob/ommparams/parmed/amber/parameters.py#L551-L552

One issue I see now is Mn vs Fe - ff99SBildn for example has Fe (residue FE) at mass 55.0, masses in https://github.com/ParmEd/ParmEd/blob/master/parmed/periodic_table.py are Mn: 54.93, Fe: 55.84 - and hence residue FE gets recognized as element Mn. Not surprising given these numbers, but if we kept addAtomTypes this would be definied correctly as Fe. What to do?

Ok, so let's combine this with fixing https://github.com/ParmEd/ParmEd/issues/580 and I'm going to (depending on this Fe problem):

* comment out the addAtomTypes section
* hard code EP and LP element recognition 

One more problem I have with the elements is these weird ion things, like IB ( 'big ion w/ waters' for vacuum (Na+, 6H2O)) - which gets recognized as Xenon because of the mass. I also see IM and IP in 99SBildn - but these are recognized as Cl and Na correctly. I guess there might be other ones like this in other FFs? How can we deal with this? I'm seeing big problems if someone wants to use say IB in OpenMM, which uses the element-based template recognition for residues etc.

drroe:
> One issue I see now is Mn vs Fe - ff99SBildn for example has Fe (residue FE) at mass 55.0, masses in https://github.com/ParmEd/ParmEd/blob/master/parmed/periodic_table.py are Mn: 54.93, Fe: 55.84 - and hence residue FE gets recognized as element Mn. Not surprising given these numbers, but if we kept addAtomTypes this would be definied correctly as Fe. What to do?

FYI what cpptraj does (based on how Sander QM/MM code recognizes elements) is to also use name information. So in your case the `F` in `FE` would let the code know it's probably iron and not manganese.

rafwiewiora:
> FYI what cpptraj does (based on how Sander QM/MM code recognizes elements) is to also use name information. So in your case the F in FE would let the code know it's probably iron and not manganese.

I like that! But that sounds like writing quite a few lines, I think it might be better for now to keep the addAtomTypes and only use that info to assign elements, where available, without adding the atom types that are not also defined in a .dat or frcmod.

parmed/amber/parameters.py
```diff
@@ -211,7 +211,7 @@ def __init__(self, *filenames):
     #===================================================
 
     @classmethod
-    def from_leaprc(cls, fname):
+    def from_leaprc(cls, fname, addAtomTypes=True):
```
hainm: Hi, Is there any specific reason to keep camelCase here?

rafwiewiora: I wasn't sure what you guys use here, camelCase is just strong in me from coding for OpenMM! Also that keeps the camelCase that Amber uses for this command. But anyway, this PR will not stay in this form, looks like it.

swails: Except for areas where the ParmEd and OpenMM APIs overlap (like `createSystem` and the like), ParmEd uses the `snake_naming_convention`.

rafwiewiora: Ok good to know!

swails:
> FYI what cpptraj does (based on how Sander QM/MM code recognizes elements) is to also use name information. So in your case the F in FE would let the code know it's probably iron and not manganese.

That's how I'd have done it if I wasn't a lazy purist :).  (I also knew that's how sander's QM/MM code *still* does it, even when `ATOMIC_NUMBER` is a thing in the prmtop now.)  But I didn't actually expect anything to get atomic masses *that wrong*.

swails:
> One issue I see now is Mn vs Fe - ff99SBildn for example has Fe (residue FE) at mass 55.0, masses in https://github.com/ParmEd/ParmEd/blob/master/parmed/periodic_table.py are Mn: 54.93, Fe: 55.84 - and hence residue FE gets recognized as element Mn. Not surprising given these numbers, but if we kept addAtomTypes this would be definied correctly as Fe. What to do?

Ugh.  Correct the stupid imprecise mass in Amber's files.

> One more problem I have with the elements is these weird ion things, like IB ( 'big ion w/ waters' for vacuum (Na+, 6H2O)) - which gets recognized as Xenon because of the mass. I also see IM and IP in 99SBildn - but these are recognized as Cl and Na correctly. I guess there might be other ones like this in other FFs? How can we deal with this? I'm seeing big problems if someone wants to use say IB in OpenMM, which uses the element-based template recognition for residues etc.

Just get rid of those atom types.  They're never used inside AMBER.  Their presence in the parm.dat files isn't really a big deal since they've been removed from the residue templates.  So it's just rubble left over that isn't hurting anyone, so it wasn't cleaned away.  Since they're not accessible from any of the Amber force fields we're trying to port, I would say they're not part of those force fields.

hainm:
> even when ATOMIC_NUMBER is a thing in the prmtop now.

But if this is not available in prmtop (generated by old tleap), qm/mm still make a guess based on name.

swails:
> But if this is not available in prmtop (generated by old tleap), qm/mm still make a guess based on name.

Yes, but the point is that sqm never uses ATOMIC_NUMBER, even if it's there.  And any prmtop that has been created in the last 5 years has ATOMIC_NUMBER present (and ParmEd has the `addAtomicNumber` command to add it if you needed to).

drroe:
> and ParmEd has the addAtomicNumber command to add it if you needed to

Cpptraj also adds atomic number information, so really sqm has no excuse :-)

swails:
You mean except the way-too-long excuse about having no time or money for any of a number of nefarious reasons?

rafwiewiora:
> Just get rid of those atom types. They're never used inside AMBER. Their presence in the parm.dat files isn't really a big deal since they've been removed from the residue templates. So it's just rubble left over that isn't hurting anyone, so it wasn't cleaned away. Since they're not accessible from any of the Amber force fields we're trying to port, I would say they're not part of those force fields.

Actually, there are still in the `.lib`, check out ions94.lib which is used with ff99SBildn. It has the IB residue and a CIO that uses the IP type....

rafwiewiora:
tagging @jchodera here for an input on these ions too, see:

> One more problem I have with the elements is these weird ion things, like IB ( 'big ion w/ waters' for vacuum (Na+, 6H2O)) - which gets recognized as Xenon because of the mass. I also see IM and IP in 99SBildn - but these are recognized as Cl and Na correctly. I guess there might be other ones like this in other FFs? How can we deal with this? I'm seeing big problems if someone wants to use say IB in OpenMM, which uses the element-based template recognition for residues etc.

and comment above quoting @swails 's reply. 

I don't think anyone would be using these in OpenMM (are these some crazy relict of the past btw?), but it might be hard picking things like this not to include, so I'm leaning towards just converting absolutely everything there is in the Amber files. But I need to sort out the element issue.

drroe:
> are these some crazy relict of the past btw?

Yes, I believe IM stands for "ion minus" and IP for "ion plus" (Cl- and Na+ respectively)

rafwiewiora:
> Yes, I believe IM stands for "ion minus" and IP for "ion plus" (Cl- and Na+ respectively)

Aha! Ok, so there's both an IM and a Cl type - IM is for the Cl- ion, Cl is from CHCl3, that makes sense. But then there's both an IP type and a Na type (described as Na+ ion in the .dat), they have the same parameters (they are just a copy), but the .lib still uses an IP, rather than Na for the Na+ ion. 

And then we have the CIO residue, which is again a Na+ ion using the IP type (though there's another Na+ residue that is the same. Wow....

swails:
These ion models are way out-of-date, though...  They were removed well before ff99SB was really considered deprecated (ff10 did away with these ions94.lib-style ions, and ff10 was identical to ff99SB for proteins).

The new ion parameters are *much* better than what you get with IM and IP.  I'm not sure how you guys want to handle this, but *I* will always use (and will always suggest others use) the newer water model-specific ion parameters.

jchodera:
I'm 100% in agreement.  These old ion parameters will [cause the ions to crystallize well below the solubility limit](http://www.ncbi.nlm.nih.gov/pmc/articles/PMC2652252/).  

We don't have a fully defined philosophy about this, but my understanding is that @peastman would prefer (1) old colde shouldn't break, and (2) we make it easy to do the right thing by default.  I would also prefer we can (3) use old parameters when needed for compatibility.  When it gets to choosing file naming conventions, we might want the default `ffxml` "bundles" to match up with the old names when possible but use updated parameter bundles, but with outdated parameters moved to something like `deprecated/` or `old/`.  We should also update the default ion names in `Modeller` as appropriate if those have changed.

rafwiewiora:
Ok, cool!

swails:
> (1) old colde shouldn't break,

This is an important concern, but I don't think the proposed changes here will affect that.  Not unless we propose *replacing* previously-converted XML files that are already in OpenMM, and I would recommend against that path (unless there's a convincing argument to do so?).

> (2) we make it easy to do the right thing by default.

To me, this is more important than the first point (although Peter may have slightly different priorities)

> (3) use old parameters when needed for compatibility.

You mean something along the lines of `ForceField('ff14sb.xml', 'really_crappy_ion_parameters.xml')`? :smiley:

Back to this PR, though -- are we still doing something with this, or do some of the newer PRs address what we're trying to fix here?

rafwiewiora:
Ok, I'm working on this now. Will:

* get rid of the changes as proposed in the first version of the PR
* hard-code EP and LP element recognition
* think about what to do with the Mn - Fe problem and how this ties into whether too comment out the whole addAtomTypes section

rafwiewiora:
Ok, I've changed this: moved addAtomTypes loading to after parameters and limited it only to overriding atomic number information for existing types, it can't add new ones. 

Also hard-coded 'EP' and 'LP' for atomic number 0. So the regex will continue not recognizing those in addAtomTypes ("" element), hence the hard-code only for parameter reading. 

How does it look?

rafwiewiora:
#587 supplements this with atomic number 0 --> "" element translation for ffxml writing

rafwiewiora:
And fixes #580

rafwiewiora:
Fixed failing test - previous K+ was added by addAtomTypes, but wasn't present in the dat or frcmod - so now is gone.

swails:
Looks good to me.  Is this ready to merge in your opinion?

rafwiewiora:
Yep, good to go in!

swails:
OK, I merged this into #588.  After Travis passes, I'll merge that in.  Then I'll merge #587 unless there are conflicts.  Good plan?  Or would you rather I swapped the merge order?

rafwiewiora:
Sounds like a good plan to me!

PR #583, opened 2016-01-29
Avoid writing empty sections in the ffxml / fix typeify_templates / change how the templates are typeified / only add ResidueTemplates to OpenMM residues

rafwiewiora:
Just adding these, for a) consistency with not writing other sections (i.e. just opening and closing tags) if there's nothing there b) to not have the empty tags to worry about when writing only residue template or only parameters files

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -168,6 +168,7 @@ class OpenMMParameterSet(ParameterSet):
         dest.write(' </Info>\n')
 
     def _write_omm_atom_types(self, dest):
+        if not self.atom_types: return
         dest.write(' <AtomTypes>\n')
         for name, atom_type in iteritems(self.atom_types):
             assert atom_type.atomic_number >= 0, 'Atomic number not set!'
@@ -178,6 +179,7 @@ class OpenMMParameterSet(ParameterSet):
         dest.write(' </AtomTypes>\n')
 
     def _write_omm_residues(self, dest):
+        if not self.residues: return
         dest.write(' <Residues>\n')
         for name, residue in iteritems(self.residues):
             if not isinstance(residue, ResidueTemplate):
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -18,6 +18,7 @@ from parmed.utils.io import genopen
 from parmed.utils.six import add_metaclass, string_types, iteritems
 from parmed.utils.six.moves import range
 import warnings
+from parmed.modeller import ResidueTemplate
 
 @add_metaclass(FileFormatType)
 class OpenMMParameterSet(ParameterSet):
@@ -107,9 +108,12 @@ class OpenMMParameterSet(ParameterSet):
         new_params.pair_types = params.pair_types
         new_params.parametersets = params.parametersets
         new_params._combining_rule = params.combining_rule
-        new_params.residues = params.residues
         new_params.default_scee = params.default_scee
         new_params.default_scnb = params.default_scnb
+        # add only ResidueTemplate instances (no ResidueTemplateContainers)
+        for name, residue in iteritems(params.residues):
+            if isinstance(residue, ResidueTemplate):
+                new_params.residues[name] = residue
 
         return new_params
 
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -18,7 +18,6 @@ from parmed.utils.io import genopen
 from parmed.utils.six import add_metaclass, string_types, iteritems
 from parmed.utils.six.moves import range
 import warnings
-from parmed.modeller import ResidueTemplate
 
 @add_metaclass(FileFormatType)
 class OpenMMParameterSet(ParameterSet):
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -137,7 +137,7 @@ class OpenMMParameterSet(ParameterSet):
             own_handle = True
         else:
             own_handle = False
-        self.typeify_templates()
+#       self.typeify_templates()
         try:
             dest.write('<ForceField>\n')
             self._write_omm_provenance(dest, provenance)
@@ -184,6 +184,11 @@ class OpenMMParameterSet(ParameterSet):
         for name, residue in iteritems(self.residues):
             if not isinstance(residue, ResidueTemplate):
                 continue
+            # if AtomTypes are available, this will throw an exception if the
+            # residue is claiming a Type that doesn't exist
+            if self.atom_types:
+                for atom in residue:
+                    atom.atom_type = self.atom_types[atom.type]
             dest.write('  <Residue name="%s">\n' % residue.name)
             for atom in residue.atoms:
                 dest.write('   <Atom name="%s" type="%s" charge="%s"/>\n' %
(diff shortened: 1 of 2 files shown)


rafwiewiora:
Now also added a fix to `typeify_templates`.

rafwiewiora:
Now also rearranged what the `typeify_templates` did into the `_write_omm_residues` function. This will now function as a test that the templates being written are not using AtomTypes that are not available in the parameters (which solves the problem with the fact that `solvents.lib` uses a CL type for CHCL3BOX, the type not existing in `parm99.dat`, but the solvent boxes were never written to the ffxml anyway). 

The typeification (this is hard to spell, is that even right?) will not happen if no Atom Types are loaded in - i.e. if I'm only converting a .lib file and I have previously ensured (will have a test in my conversion script) that all types claimed by templates are available.

@ChayaSt is this good enough for your stream files?

parmed/openmm/parameters.py
```diff
@@ -178,10 +179,16 @@ def _write_omm_atom_types(self, dest):
...
 
     def _write_omm_residues(self, dest):
+        if not self.residues: return
         dest.write(' <Residues>\n')
         for name, residue in iteritems(self.residues):
             if not isinstance(residue, ResidueTemplate):
                 continue
+            # if AtomTypes are available, this will throw an exception if the
+            # residue is claiming a Type that doesn't exist
+            if self.atom_types:
```
swails: Isn't this what `typeify_templates` does?  Why did you comment-out the call above, only to essentially re-implement it here?

swails: Ah, I see from the commit message.  Maybe we can make `typeify_templates` a little smarter about only "typeifying" what it can and not complaining when some types don't exist?

Also, maybe we should get rid of `ResidueTemplateContainer` instances (basically the solvent boxes) when calling `OpenMMParameterSet.from_parameterset` so that we only *keep* residues that we explicitly plan on writing to XML files.  What do you think about that?

rafwiewiora: > Ah, I see from the commit message. Maybe we can make typeify_templates a little smarter about only "typeifying" what it can and not complaining when some types don't exist?

The thing is I don't see the point of 'typeifying' other than it being a test for the existance of the atom types - because all it does is bind the AtomType objects to `atom.atom_type` but we subsequently don't use that at all - the writing function uses `atom.type`. So making it not complain when some types don't exist makes it essentially pointless, right?

> Also, maybe we should get rid of ResidueTemplateContainer instances (basically the solvent boxes) when calling OpenMMParameterSet.from_parameterset so that we only keep residues that we explicitly plan on writing to XML files. What do you think about that?

I don't see much difference between removing these at calling OpenMMParameterSet.from_parameterset or at writing residues to be fair - but maybe that would make the code a bit cleaner. Would there be any use case for having the Containers part of the OpenMMParameterSet in the future? 

How about we do:
* have the 'self.typeify_templates' call but add a 'if not self.atom_types: return` line to the `typeify_templates` function (so if no Atom Types available - assume user takes the risk of some atom types possibly being wrong, if any Atom Types available - assume user wants the check). But if feeding the `atom.atom_type` 's is useless like I think - maybe we should just replace that with an `assert` test? @ChayaSt how does this work for you? She has the stream files, and it seems like they might contain some Atom Types but not all? In this case this wouldn't work and we should maybe just get rid of typeify_templates completely and accept the risk? 

* and remove the `ResidueTemplateContainer` instances at calling `OpenMMParameterSet.from_parameterset`

How does this sound?

swails: > Would there be any use case for having the Containers part of the OpenMMParameterSet in the future?

No -- Containers are really only for solvent boxes or multi-residue templates (which I've only ever seen Amber use for solvent boxes).  OpenMM would never make use of that.

> I don't see much difference between removing these at calling OpenMMParameterSet.from_parameterset or at writing residues to be fair - but maybe that would make the code a bit cleaner.

This assumes that the only thing you will ever *want* to do is write out an XML file.  If the OpenMMParameterSet does not really understand or deal with solvent boxes, then it should never inherit one from any kind of constructor IMO.  If you filter it out at creation time, you never have to protect against their inclusion anywhere else.

> have the self.typeify_templates call but add a if not self.atom_types: return line to the typeify_templates function (so if no Atom Types available - assume user takes the risk of some atom types possibly being wrong, if any Atom Types available - assume user wants the check). But if feeding the atom.atom_type 's is useless like I think - maybe we should just replace that with an assert test? @ChayaSt how does this work for you? She has the stream files, and it seems like they might contain some Atom Types but not all? In this case this wouldn't work and we should maybe just get rid of typeify_templates completely and accept the risk?

OK, `typeify_templates` is just causing problems for now.  This was really added with an eye to the future where ParmEd can do its own parameter assignment (like leap and ForceField).  And `atom_type` provides the same kind of type/class dichotomy that OpenMM and Tinker supports (and GROMACS, too, although it is used a little less often).  But as you point out, it doesn't have much use now outside of error-checking for the use of uninitialized/undefined types.  And it would be good to support *just* reading residue templates and not parameters.

Let's keep the method around, but the changes here that get rid of the explicit call to the method are probably the best way to go for now.

> and remove the ResidueTemplateContainer instances at calling OpenMMParameterSet.from_parameterset

:+1: 

> How does this sound?

Good!

rafwiewiora: Awesome, I'm on it!

swails:
This guy looks done to me.  Is there any reason not to merge it yet?

parmed/openmm/parameters.py
ChayaSt: > @ChayaSt how does this work for you? She has the stream files, and it seems like they might contain some Atom Types but not all? In this case this wouldn't work and we should maybe just get rid of typeify_templates completely and accept the risk?

This is still giving me a key error with the stream files.

rafwiewiora:
It should be good now! I've added the:

* only add ResidueTemplate instances to residues

* `typeify_templates` only engaged when there are some AtomTypes, but only returns a warning if some residues can't find AtomTypes

Should be good to merge - unless some testing needed here?

parmed/openmm/parameters.py
```diff
@@ -137,7 +142,12 @@ def write(self, dest, provenance=None):
             own_handle = True
         else:
             own_handle = False
-        self.typeify_templates()
+        if self.atom_types:
+            try:
+                self.typeify_templates()
+            except:
```
swails: You should (almost) always avoid bare excepts, since it can mask real errors.  I think `KeyError` is the expected exception if things go wrong, yes?

rafwiewiora: That's right, fixed!

swails:
Travis is failing because of linter errors: https://travis-ci.org/ParmEd/ParmEd/jobs/106310318#L817

You can run this test by hand as well inside the ParmEd repo:

```
devtools/travis-ci/pyflakes_check.sh
```

pyflakes has helped me catch issues in the past...

rafwiewiora:
Aha! Re-import fixed.

rafwiewiora:
Maybe a quick: load in solvents.lib, make OpenMMParameterSet, assert number of residues (to check only ResidueTemplates are through), test writing kind of test?

swails:
All the existing tests *should* hit this code, so if you want to add a test to make sure the solvent boxes are excluded that would be fine (but not fully necessary).

rafwiewiora:
If you say it's unnecessary, then it's gotta be! Let's go with what we've got.

swails:
haha... then in it goes.

Thanks!

PR #585, opened 2016-01-31
allow reading of .lib files through AmberParameterSet() call

rafwiewiora:
As discussed in https://github.com/ParmEd/ParmEd/issues/486#issuecomment-177010286

Can now call `AmberParameterSet(file)` with .lib and .off files in addition to .dat and frcmod.

### DIFF
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -94,7 +94,13 @@ class AmberParameterSet(ParameterSet):
         is_fmt : bool
             True if it is an Amber-style parameter file. False otherwise.
         """
-        with closing(genopen(filename, 'r')) as f:
+        if isinstance(filename, string_types):
+            f = genopen(filename, 'r')
+            own_handle = True
+        else:
+            f = filename
+            own_handle = False
+        try:
             f.readline()
             line = f.readline()
             if not line.strip(): # Must be an frcmod file
@@ -191,6 +197,11 @@ class AmberParameterSet(ParameterSet):
                     return True
             else:
                 return True
+        finally:
+            if own_handle:
+                f.close()
+            else:
+                f.seek(0)
 
     #===================================================
 
@@ -200,13 +211,22 @@ class AmberParameterSet(ParameterSet):
         self.residues = dict()
         for filename in filenames:
             if isinstance(filename, string_types):
-                self.load_parameters(filename)
+                if self.id_format(filename): # dat or frcmod file
+                    self.load_parameters(filename)
+                else: # assume it's a lib or off file
+                    self.residues.update(AmberOFFLibrary.parse(filename))
             elif isinstance(filename, Sequence):
                 for fname in filename:
-                    self.load_parameters(fname)
+                    if self.id_format(filename):
+                        self.load_parameters(fname)
+                    else:
+                        self.residues.update(AmberOFFLibrary.parse(filename))
             else:
                 # Assume open file object
-                self.load_parameters(filename)
+                if self.id_format(filename):
+                    self.load_parameters(filename)
+                else:
+                    self.residues.update(AmberOFFLibrary.parse(filename))
 
     #===================================================
 
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -94,13 +94,7 @@ class AmberParameterSet(ParameterSet):
         is_fmt : bool
             True if it is an Amber-style parameter file. False otherwise.
         """
-        if isinstance(filename, string_types):
-            f = genopen(filename, 'r')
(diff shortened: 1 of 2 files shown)


parmed/amber/parameters.py
```diff
@@ -94,7 +94,13 @@ def id_format(filename):
         is_fmt : bool
             True if it is an Amber-style parameter file. False otherwise.
         """
-        with closing(genopen(filename, 'r')) as f:
+        if isinstance(filename, string_types):
```
swails: You can't do this in `id_format`, since it consumes part of the file trying to determine if the file is a particular format, and not all file-like objects that get sent to id_format are necessarily seekable (websites, for example).

parmed/amber/parameters.py
```diff
@@ -200,13 +211,22 @@ def __init__(self, *filenames):
         self.residues = dict()
         for filename in filenames:
             if isinstance(filename, string_types):
-                self.load_parameters(filename)
+                if self.id_format(filename): # dat or frcmod file
```
swails: I would do it the other way around.  Change this to:

```Python
if AmberOFFLibrary.id_format(filename):
    self.residues.update(AmberOFFLibrary.parse(filename)
else:
    self.load_parameters(filename)
```

The reason is that if you pass a wrong file type to it, then it will error as though it's being treated as an OFF library, which is a bit misleading.  Since we're special-casing AmberOFFLibrary, we should handle *that* case explicitly and pass everything else to the core parser.

swails: The above comment applies to the other instances here as well.

swails:
Looks good!  Can you add a test case for this as well?

parmed/amber/parameters.py
swails: Also, this hard-codes an assumption that `filename.tell()` returns 0 in the beginning.

rafwiewiora:
Alright, thanks for the corrections! So I don't know how to deal with the open file-objects then - how can you test the format? Even if I do a try: load OFF - except: dat/frcmod parses, it will still consume the first line if it's not a lib/off...

rafwiewiora:
Actually, how about this for `AmberOFFLibrary.id_format`:

``` python
        if isinstance(filename, string_types):
            with closing(genopen(filename, 'r')) as f:
                if AmberOFFLibrary._headerre.match(f.readline()):
                    return True
                return False
        else: # assume open-file like object
            x = filename.tell()
            if AmberOFFLibrary._headerre.match(filename.readline()):
                filename.seek(x)
                return True
            return False
```

swails:
I would treat an open file-like object as though it was a parameter file.  ParmEd does not support any kind of file-type detection on open file handles because I prioritized abstracting the definition of "file" over having `id_format` work on file handles.

Also, not all things that ParmEd treats as "file-like" objects support seek and tell.  We could support format detection on *only* file-like objects that do support seek and tell, then fail on file-likes that don't (like `urlopen`), but that should be done across all `FileFormatType` metaclasses.

Do you have a specific use-case for this right now, or are you just trying to cover all the bases for this PR?

rafwiewiora:
I see, thanks!

Yeah, I was just trying to cover all the bases, no particular use case in mind.

rafwiewiora:
Committed fixes, working on the test.

swails:
The test here can be *very* simple.  Basically just pass a couple parameter files and library files to the constructor and make sure that it adds to the `residues` dict as well as parameters.  The parameter and OFF file parsers are rather extensively tested elsewhere.

Perhaps make two tests, though, one where the arguments are all separate, like

```Python
params = AmberParameterSet('parm10.dat', 'frcmod.ff14SB', 'amino10.lib', 'aminoct10.lib')
```

and one where it's a single-argument list (which is supported because that's a natural form of how inspection-based methods would return the filenames, like parsing from a leaprc or globbing):

```Python
params = AmberParameterSet(['parm10.dat', 'frcmod.ff14SB'], ['amino10.lib', 'aminoct10.lib'])
```

Eventually test coverage will be a bigger part of the CI machinery, and test coverage for ParmEd as a whole is improving.

rafwiewiora:
You got it! Good to merge if this looks ok.

swails:
Looks good, thanks!

rafwiewiora:
Damn it, the perfectionist part of my mind noticed I had redundant `os.path.join` 's : 
``` python
os.path.join(get_fn('amino12.lib')),
os.path.join(get_fn('aminoct12.lib'))
```

We can fix that if you want?

swails:
Sure -- go ahead and open a new PR for that if you have time.

Thanks!

rafwiewiora:
alright!

PR #587, opened 2016-02-02
set empty string element for atomic number 0 (EP/LP)

rafwiewiora:
Supplement to #582 (going into different branches). 

That's just a whitespace removal on that one line.


rafwiewiora:
And fixed the test, the previous K+ was added from addAtomType only, not present in the .dat or frcmod.

swails:
OK, looks like there are no merge conflicts.  I'll go ahead and merge this, too.

rafwiewiora:
Thanks!

PR #590, opened 2016-02-04
Add args to: ignore ions/solvents at leaprc reading, only write out atom types that are used in templates in ffxml

rafwiewiora:
- [x] Add arguments for ignore ions and solvents in leaprc
- [x] Add argument for only write out those Atom Types to ffxml that are used in the templates (this will clean up all the ions, EP/LP etc.)

### DIFF
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -223,7 +223,7 @@ class AmberParameterSet(ParameterSet):
     #===================================================
 
     @classmethod
-    def from_leaprc(cls, fname, ions=True, solvents=True):
+    def from_leaprc(cls, fname):
         """ Load a parameter set from a leaprc file
 
         Parameters
@@ -231,13 +231,6 @@ class AmberParameterSet(ParameterSet):
         fname : str or file-like
             Name of the file or open file-object from which a leaprc-style file
             will be read
-        ions :  bool
-            If False, loadOffs of ion templates (atomic_ions.lib, ions94.lib, 
-            ions91.lib) are ignored. The names of ignored files are current as of 
-            Amber14/AmberTools15
-        solvents : bool
-            If False, loadOffs of the solvent templates (solvents.lib). The name 
-            of the ignored file is current as of Amber14/AmberTool15       
 
         Notes
         -----
@@ -267,13 +260,7 @@ class AmberParameterSet(ParameterSet):
         for fname in _loadparamsre.findall(text):
             params.load_parameters(_find_amber_file(fname))
         # Now process the library file
-        lib_files = _loadoffre.findall(text)
-        if not ions:
-            lib_files = [fname for fname in lib_files if fname not in 
-            ('atomic_ions.lib', 'ions94.lib', 'ions91.lib')]
-        if not solvents:
-            lib_files = [fname for fname in lib_files if fname != 'solvents.lib']    
-        for fname in lib_files:
+        for fname in _loadoffre.findall(text):
             params.residues.update(
                     AmberOFFLibrary.parse(_find_amber_file(fname))
             )
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -524,7 +524,8 @@ class AmberParameterSet(ParameterSet):
         # Now assign all of the equivalenced atoms
         for atyp, otyp in iteritems(equivalent_ljtypes):
             otyp = self.atom_types[otyp]
-            self.atom_types[atyp].set_lj_params(otyp.epsilon, otyp.rmin)
+            if atyp in self.atom_types:
+                self.atom_types[atyp].set_lj_params(otyp.epsilon, otyp.rmin)
         line = next(fiter).strip()
         if line == 'LJEDIT':
             rawline = next(fiter)
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -223,7 +223,7 @@ class AmberParameterSet(ParameterSet):
     #===================================================
 
     @classmethod
-    def from_leaprc(cls, fname):
+    def from_leaprc(cls, fname, ions=True, solvents=True):
         """ Load a parameter set from a leaprc file
 
         Parameters
@@ -231,6 +231,13 @@ class AmberParameterSet(ParameterSet):
         fname : str or file-like
             Name of the file or open file-object from which a leaprc-style file
             will be read
+        ions :  bool
+            If False, loadOffs of ion templates (atomic_ions.lib, ions94.lib, 
+            ions91.lib) are ignored. The names of ignored files are current as of 
+            Amber14/AmberTools15
+        solvents : bool
+            If False, loadOffs of the solvent templates (solvents.lib). The name 
+            of the ignored file is current as of Amber14/AmberTool15       
 
         Notes
         -----
@@ -260,7 +267,13 @@ class AmberParameterSet(ParameterSet):
         for fname in _loadparamsre.findall(text):
             params.load_parameters(_find_amber_file(fname))
         # Now process the library file
-        for fname in _loadoffre.findall(text):
+        lib_files = _loadoffre.findall(text)
+        if not ions:
+            lib_files = [fname for fname in lib_files if fname not in 
+            ('atomic_ions.lib', 'ions94.lib', 'ions91.lib')]
+        if not solvents:
+            lib_files = [fname for fname in lib_files if fname != 'solvents.lib']    
+        for fname in lib_files:
             params.residues.update(
                     AmberOFFLibrary.parse(_find_amber_file(fname))
             )
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -18,6 +18,7 @@ from parmed.utils.io import genopen
 from parmed.utils.six import add_metaclass, string_types, iteritems
 from parmed.utils.six.moves import range
 import warnings
+from parmed.exceptions import ParameterWarning
 
 @add_metaclass(FileFormatType)
 class OpenMMParameterSet(ParameterSet):
@@ -151,12 +152,15 @@ class OpenMMParameterSet(ParameterSet):
                 typeified = True
             except KeyError:
                 warnings.warn('Some residue templates are using unavailable '
-                              'AtomTypes')
+                              'AtomTypes', ParameterWarning)
         if not write_unused:
             if not typeified:
                 warnings.warn('Typification of the templates was not successful. '
-                              'Proceeding with write_unused=False is not advised')
-            skip_types = _find_unused_types()
+                              'Proceeding with write_unused=False is not advised',
+                               ParameterWarning)
+            skip_types = self._find_unused_types()
+        else:
+            skip_types = set()
         try:
             dest.write('<ForceField>\n')
             self._write_omm_provenance(dest, provenance)
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -40,7 +40,6 @@ class OpenMMParameterSet(ParameterSet):
     --------
     :class:`parmed.parameters.ParameterSet`
     """
-
     @staticmethod
     def id_format(filename):
         """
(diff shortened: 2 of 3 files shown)


rafwiewiora:
Ok done with the first take at this - what do you think?

rafwiewiora:
Ping for @jchodera and @ChayaSt

parmed/amber/parameters.py
```diff
@@ -223,7 +223,7 @@ def __init__(self, *filenames):
     #===================================================
 
     @classmethod
-    def from_leaprc(cls, fname):
+    def from_leaprc(cls, fname, ions=True, solvents=True):
```
swails: I'm not sure how I feel about this.  It's satisfying a particular use-case that is specific to the way in which we're currently thinking about converting the Amber parameter files to OpenMM XMLs.  It seems like this functionality fits more naturally in the OpenMM conversion script rather than in here.  At least to me.

rafwiewiora: Oh I completely agree, I feel super uncomfortable anytime I have to add an argument to a function just to satisfy something like this - but I thought it might just be on the side of 'useful for the future' so I put it in as a conversation starter. 

The thing is: how do I do this in a conversion script? Easiest way - copy this function and make these exact modifications to it (perhaps without the arguments, just hard-coded no solvents/ions. Perhaps the more neat-seaming way - pre-modify the string to remove those - but then I'm just gonna use the parsing from this function anyway to do that, so why not just copy it all. 

Honestly, it completely doesn't matter to me which way we do it. I think it's a bit cleaner if as much code is within ParmEd and then the conversion script is just a few lines. But I know that's not so clean to you. 

Perhaps we could meet half-way and have an 'ignore' list instead? For example by passing `from_leaprc(fname, ignore=['ions', 'solvents']` you would not load any file that contains 'ions' or 'solvents' in its name string. Similarly you could pass 'nucleic' or 'amino' to it, or perhaps 'parm' or 'frcmod' if you're feeling adventurous. Clearly that's a dev functionality mostly though. What do you think?

rafwiewiora:
Also what's up with the coveralls failing in the tests?

jchodera:
> The thing is: how do I do this in a conversion script?

Is there a simpler way to do it from the conversion script?
* Read `leaprc` contents line-by-line appending to a string
* Drop lines that load either `solvents.lib` or `ions.lib`
* Call `from_leaprc()` wrapping the string in a `StringIO`?

rafwiewiora:
Alright, I'm convinced, let's do it that way. I'm just gonna stay the .lib regex though!

parmed/openmm/parameters.py
```diff
@@ -128,7 +129,9 @@ def write(self, dest, provenance=None):
             If present, the XML file will be tagged with the available fields.
             The keys of this dictionary are turned into the XML tags, and the
             values become the contents of that tag. Default is no provenance
-
+        clean_params : bool
+            If True, any atom types which are not used in the residue templates
+            and any parameters using those atom types will not be written
```
swails: Need a blank line before `Notes`

swails:
The ignore keyword with matching regex is better, but I like @jchodera's suggestion the best.  If that works, that's what I'd prefer to do.

As for coveralls failing, I'm not sure.  If it happens again, I'll try to retrigger the build.

Can you add a test for this new function?

rafwiewiora:
Test added! And I've added two more things that were causing problems for me so far going forward with conversion:
* L-J equivalences in .dat for atom types that are not defined otherwise - caught on parmAM1.dat, `N   NA  N2  N*  NC  NB  NP  NO` NO type not defined

* added my type skipping to the NBFix `NotImplementedError` handling - e.g. parm14ipq.dat has NBFix for tip4pew OW

rafwiewiora:
@swails this good to merge you think? I can remove these mods from the last comment, if you want to tackle this separately together with the other issues I've found.

parmed/openmm/parameters.py
```diff
@@ -141,12 +145,18 @@ def write(self, dest, provenance=None):
...
+        typeified = False
         if self.atom_types:
             try:
                 self.typeify_templates()
+                typeified = True
             except KeyError:
                 warnings.warn('Some residue templates are using unavailable '
                               'AtomTypes')
+        if clean_params:
+            assert typeified, """Typification of the templates was not successful.
```
swails: Let's turn this into a warning.  Also, never use `assert` unless triggering the assert means a bug *in the library*.  The only times I use assert is when the design of ParmEd dictates that something must *always* be `True`.  For example, [here is an assertion that the combining rule must be either lorentz or geometric](https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/_amberparm.py#L806-L807), but [setting it to anything else is an explicit error](https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L1381-L1385).

This is a defensive assertion to make sure that if I someday add support for another combining rule but forget to update the `AmberParm` subclass, an assertion error will hopefully limit the severity of the bug.  This is also true wherever I raise an `AssertionError` (which is ignored in the coverage calculation since I never WANT to hit those lines).  But most critically is the fact that [what Python means by "optimized compilation" is to get rid of assertions](http://stackoverflow.com/questions/2830358/what-are-the-implications-of-running-python-with-the-optimize-flag).  So the moral is, if you want an error check to be guaranteed to run, don't use `assert`.  It's there to aid in development and debugging.

parmed/openmm/parameters.py
```diff
@@ -166,6 +176,13 @@ def write(self, dest, provenance=None):
             if own_handle:
                 dest.close()
 
+    def _clean_params(self):
+        keep_types = set()
+        for name, residue in iteritems(self.residues):
+            for atom in residue.atoms:
+                keep_types.add(atom.type)
+        self.skip_types = [typ for typ in self.atom_types if typ not in keep_types]
```
swails: Let's not cache this.  Let's have `_clean_params` (maybe rename it to `_find_unused_types`) simply return a set (which is a lot more efficient for lookups than a list).  So replace this last line with:

```Python
```

Set comps are supported in every version of Python ParmEd supports, so this is a fine construct to use.  This will protect us against caching a list of types that may wind up out-of-date if we add any residues to our `OpenMMParameterSet`.

Then call this function right before it's used.  You can avoid calling it in all of the `_write_omm*` methods by simply calling it in `write` and passing it as a parameter -- there is absolutely no importance put on ensuring backwards-compatibility of private methods, so go ahead and make that a required positional argument for all of the `_write_omm*` methods.

parmed/openmm/parameters.py
```diff
@@ -412,4 +438,7 @@ def _write_omm_scripts(self, dest):
             raise NotImplementedError('Geometric combining rule not currently '
                                       'supported.')
         if len(self.nbfix_types) > 0:
-            raise NotImplementedError('NBFIX not currently supported')
+            for (a1, a2), nbfix in iteritems(self.nbfix_types):
+                if any((a in self.skip_types for a in (a1, a2))): continue
```
swails: `continue` on a separate line

parmed/openmm/parameters.py
swails: Ah, I see now that the cache is simply discarded and regenerated each time `write` is called.  OK, there are two ways of doing this:

1. Have a `prune` method that goes through and removes any types and parameters that have no representation in the residue templates.
2. Do what we're doing (prune at write time, but make sure they're still in the parameter set).

My initial preference was 1, since the intuitive behavior (to me) was for `write` to simply dump its entire database.  But I think I've convinced myself that what you're doing here is better.  It's simpler from an implementation perspective *and* more flexible from a practical perspective.

So all I think I'd like to see are better names.  This isn't "cleaning" the parameter set (since nothing is getting removed or deleted) -- it's simply finding unused types.  So what do you think about changing the `clean_params=False` keyword to `write_unused=True` (where `write_unused` corresponds to both atom types *and* parameters including those types)?  And per my suggestion above, what about changing `_clean_params` to `_find_unused_types`?

Also, the way that you're currently using `skip_types` -- which I agree is the correct way to do it -- should not be persistent (i.e., don't store it as an instance attribute -- generate it once inside `write` and pass it to all of the functions that need to use it).

rafwiewiora:
Thanks for all the comments and the advice on `assert`!

Pruning the atom types and the parameters out of the ParameterSet was actually the first thing I started with, but then I could find no justification for that to happen - a) because why force user to regenerate the OpenMMParameterSet if they want to write both all-atom-types and cleaned-up-atom-types; b) The pruning would somewhat suggest that a pruned ParameterSet was the best thing to have, while I'm only implementing this for a particular `write` use-case - an ffxml protein-nucleic bundle that will ensure there are no redundant Atom Types that could cause overlaps in OpenMM. So yeah, glad you agree!

Alright, so I'm doing:

* change names - your suggestions are great

* don't catche - function returns set, pass to every function that needs it 

* continue on separate line - what's your 'philosophy' on this out of curiosity? I only started doing that for the ParmEd PR's when I noticed that you used the one line syntax quite often

rafwiewiora:
Oh and: 
* turn the assert into warning. What's a good way of forcing that any warning be translated into an exception within my conversion script? I want to do that so I absolutely don't miss anything potentially wrong.

swails:
> continue on separate line - what's your 'philosophy' on this out of curiosity? I only started doing that for the ParmEd PR's when I noticed that you used the one line syntax quite often

I use single-lines for one-off operations.  Something like `if condition: break` or `if condition: continue`.  If it has an `else`, I never do that.  I hate the look of the following:

```Python
if something: do_this()
else:
    do_that_other_really_long_name_thing()

# And
if something: do_this()
else: do_that()
```

I think what it boils down to is that when I can avoid it, I like each line of code to represent as few thoughts as possible.  So `if condition: break` or `if condition: continue` expresses a single thought -- the loop or iteration is terminated in a specific case.  If a conditional is complex enough to have "else" or "elif", it's too complex to break the if conditional into a single thought.  So it interrupts the way I read code (akin to how prolines "chink" a polypeptide :smiley:).

swails:
> turn the assert into warning. What's a good way of forcing that any warning be translated into an exception within my conversion script? I want to do that so I absolutely don't miss anything potentially wrong.

`warnings.filterwarnings`.  Look through the test cases to find where I do an `assertRaises` on a warning class.  You'll either see something like `warnings.filterwarnings('error', category=MyWarning)` right before that or in the `setUp` function of that test class.

I would also suggest giving the warning one of the predefined warning types in `parmed.exceptions` so you can selectively elevate the warnings you want to to errors.  Perhaps `ParameterWarning`?

rafwiewiora:
> (akin to how prolines "chink" a polypeptide ).

Ha, nice!

Thanks a lot!

swails:
OK -- let me know when you think this is ready to merge and I'll take another look (although I imagine I'll just need to merge it at that point)

rafwiewiora:
Alright, this should be ready. I went with calling the set `skip_types` as it was, I thought about changing that to `unused_types` in keeping with the rest of the name changes - but I think that doesn't necessarily say what the functions are going to do with the types, which is skip them.

rafwiewiora:
Nope, fixing. Also forgot about the predefined warning. Stay tuned.

rafwiewiora:
Pass. Good to have a look at @swails

swails:
Looks good! Thanks!

parmed/openmm/parameters.py
```diff
@@ -116,7 +117,7 @@ def from_parameterset(cls, params, copy=False):
 
         return new_params
 
-    def write(self, dest, provenance=None):
+    def write(self, dest, provenance=None, write_unused=True):
```
hainm: how is about changing `write_unused` to `include_unused` or something else.

it seems to be duplicated `def write(..., write_..)`

swails: I like your name better, but not enough to ask @rafwiewiora to open a new PR to fix it.  If he wants to, I'd merge it, but I think this name is acceptable.  So it's up to him.

hainm:
just have a comment about doc: I myself prefer to add brief doc to private method too. Python's duck typing is good to write but not really good to catch the flow. (this is from my experience of reading others' code).

For example:
```python
   def _write_omm_urey_bradley(self, dest, skip_types):
          if not self.urey_bradley_types: return None
```

swails:
I add docstrings to private methods when I think they're warranted (probably in more cases than not).  But since they're not rendered into the actual docs, I only do that if I think it adds to the general understanding of the code.  On private methods, they're really no better (and in some ways slightly worse) than simple comments.  But these methods are short enough, with names that are descriptive enough, that I don't think they add much.

parmed/openmm/parameters.py
rafwiewiora: In this case, I'm going to keep the `write_unused`, because I like it more. :)

Issue #593, opened 2016-02-05
Summary of problems with loading all Amber FFs 


rafwiewiora:
A few groups of problems here:

1. blank lines in .lib files: all_aminoAM1.lib, all_aminoAM3.lib (called by leaprc.ffAM1 and leaprc.ffPM3), both line 1373. Causes ValueError here: https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/offlib.py#L168

2. cannot understand a line - bond or angle lines:

 all_modrna08.frcmod (leaprc.modrna08)
 ```
parmed.exceptions.ParameterError: Could not understand BOND line [C -CM 449.9   1.461   same as c -c2]
```

 parm91X.ua.dat (leaprc.ff84, leaprc.ma.ff84) and parm91X.dat (leaprc.ff86):
 ```
parmed.exceptions.ParameterError: Could not understand ANGLE line [OW-HW  553.0    0.9572]
```

3. HEAD atom is not connect0:
 ```
Traceback (most recent call last):
  File "try_one_leaprc.py", line 27, in <module>
    params = pmd.amber.AmberParameterSet.from_leaprc(leaprc)
  File "/Users/rafalpwiewiora/anaconda/lib/python2.7/site-packages/ParmEd-2.2.0-py2.7-macosx-10.5-x86_64.egg/parmed/amber/parameters.py", line 265, in from_leaprc
    AmberOFFLibrary.parse(_find_amber_file(fname))
  File "/Users/rafalpwiewiora/anaconda/lib/python2.7/site-packages/ParmEd-2.2.0-py2.7-macosx-10.5-x86_64.egg/parmed/amber/offlib.py", line 140, in parse
    residues[name] = AmberOFFLibrary._parse_residue(fileobj, name)
  File "/Users/rafalpwiewiora/anaconda/lib/python2.7/site-packages/ParmEd-2.2.0-py2.7-macosx-10.5-x86_64.egg/parmed/amber/offlib.py", line 331, in _parse_residue
    raise RuntimeError('HEAD atom is not connect0')
RuntimeError: HEAD atom is not connect0
```

 for lipid11.lib (leaprc.lipid11), lipid14.lib (leaprc.lipid14), cph_nucleic_caps.lib (leaprc.constph)

4. Access to lib/oldff, missing from `find_amber_file`: https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/parameters.py#L44 

[...]

PR #594, opened 2016-02-05
Add a bunch of tests and fix a couple issues

swails:
This PR is opened to fix the issues raised in #593, but also adds a bunch of tests I've been sitting on for the actions.py module.

TODO:

- [x] Issue 1 (blank lines in lib files)
- [x] Issue 2 (Wonky whitespace delimiters in frcmod/dat files)
- [x] Issue 3 (Weird connectivity)
- [x] Issue 4 (expose `oldff` somehow)
- [x] Fix reading GLYCAM parameter files [NEW]


swails:
OK, the only thing left to do here is expose `lib/oldff` if this is actually needed.  But is it?  There's nothing I see in that folder that would be worth using as far as I can tell.  The newest thing there is ff91, and that's before the scaling factors were set to 1.2 and 2.0.  The current FF lineage started at FF94 (that was the original Cornell et al. work).  FF91 was a completely different beast.

And actually, everything FF91 and older uses a 10-12 potential on some atom pairs, and that FF is only supported in OpenMM through ParmEd's AmberParm (`AmberPrmtopFile` in OpenMM chokes on it -- not even sure if it raises an exception to be honest, but it certainly doesn't implement the 10-12 potential).

jchodera:
We can add the 10-12 potential as a `CustomNonbondedForce`.  How is that represented in `AmberParameterSet`?

swails:
> We can add the 10-12 potential as a CustomNonbondedForce.

I know we can ([that's what `AmberParm` does](https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/_amberparm.py#L996-L1021) when it's present).

> How is that represented in AmberParameterSet?

[It's not](https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/parameters.py#L494-L496), and I have no plans to add it.  That term hasn't been used in decades with no plans to add it back.  If users *must* have it in OpenMM, they can parametrize through LEaP and I think that's good enough.  You'll never convince @peastman or Vijay that FF91 or anything older is worth bundling :)

swails:
OK, this is ready to go.  Let me know if anything looks weird or doesn't quite work for your needs.

rafwiewiora:
> OK, the only thing left to do here is expose lib/oldff if this is actually needed. But is it? There's nothing I see in that folder that would be worth using as far as I can tell. The newest thing there is ff91

Not quite, from leaprc.ff02 (and that's an FF that's even listed in the 'main' table of forcefields to use in LEaP on page 31 of the newest manual):
```
loadOff all_nucleic02.lib
loadOff all_amino02.lib
loadOff all_aminoct02.lib
loadOff all_aminont02.lib
```

And those files are in lib/oldff - my money is on someone made a mistake copying? I see no reason to have ff02 files and oldff and ff99 files not in oldff....

rafwiewiora:
Sorry, I commented on a parm/oldff line in one of your commits wrongly rather then in the PR (saying I didn't see that directory), but now I see here you've commented it out! (I deleted that comment, just saying in case you have a weird notification not pointing anywhere existing...)

rafwiewiora:
> OK, this is ready to go. Let me know if anything looks weird or doesn't quite work for your needs.

I guess you've tested against all the cases that were causing problems that I reported? Then no further testing / needs I can do!

swails:
That's an old polarizable FF.  Admittedly newer than 91, but ParmEd doesn't know how to handle polarizable FFs (except some limited AMOEBA support).  There's still nothing in `lib/oldff` that ParmEd currently supports *or* that we'd want to distribute with OpenMM.  Nobody in Amber currently uses it to my knowledge.

I *did* add a `search_oldff` argument in the `from_leaprc` constructor, but you should never set it to `True` in any of your conversions.

rafwiewiora:
> That's an old polarizable FF. Admittedly newer than 91, but ParmEd doesn't know how to handle polarizable FFs (except some limited AMOEBA support).

Oh I see!

> I did add a search_oldff argument in the from_leaprc constructor, but you should never set it to True in any of your conversions.

Alright, that's good to have just in case.

swails:
The parm91Xua.dat or whatever file (with angle definitions using only 2 atom types) weren't tested (nor was supported added).  But everything else in #593 was addressed (lipid11.lib was not explicitly tested, but lipid14.lib is now part of the test suite).

Still doesn't support loadMol2, but I have to adjust the leaprc processing a little bit to support that (since in that case, order of loadOFF and loadMol2 files matter).  I'll go ahead and merge this and add the loadMol2 command support as a separate feature enhancement request.

PR #597, opened 2016-02-09
OpenMM provenance tweaks

rafwiewiora:
Not tested yet, just want to show you the idea first.

* Write multiple elements with the same tag - e.g. two references, by having `provenance = {'Reference' : ['reference1', 'reference2']`}

* Write elements with attributes - the Source element, by having:
``` python
provenance = {'Source' : {'Source' : 'leaprc.ff14SB', 'md5hash' : '...', 'sourcePackage' : 'AmberTools', 'sourcePackageVersion' : 15}
```

i.e. pass nested dict, one of the keys has to be identical to the element tag, that becomes the content, rest of keys feed attributes.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -127,8 +127,24 @@ class OpenMMParameterSet(ParameterSet):
             attribute) to which the XML file will be written
         provenance : dict, optional
             If present, the XML file will be tagged with the available fields.
-            The keys of this dictionary are turned into the XML tags, and the
-            values become the contents of that tag. Default is no provenance
+            Keys of the dictionary become XML element tags, the values of the
+            dictionary must be instances of any of:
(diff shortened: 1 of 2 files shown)


rafwiewiora:
Actually just realized this nested dict needs to be an OrderedDict(), gonna change that.

rafwiewiora:
> Actually just realized this nested dict needs to be an OrderedDict(), gonna change that.

Yeah no, I can just pass that from the conversion script ok. 

@swails I'm done here with code, Travis pass, coveralls complains - but it won't show me the coverage on the code, can't get the source file. I guess it won't assertions of the KeyError and TypeError I added?

swails:
ok!

hainm:
@swails 

I think you should update coverage.io setting

```
COVERAGE THRESHOLD FOR FAILURE

COVERAGE DECREASE THRESHOLD FOR FAILURE
```
picture1

PR #607, opened 2016-02-16
Fix "Inappropriate ioctl for device" in .top writer

ctk3b:
I encountered this error when running a jupyter notebook on a remote server and trying to save a `.top` file.

There seems to be a fundamental issue with `os.getlogin()` (as alluded to in the Python [docs](https://docs.python.org/2/library/os.html?highlight=getlogin#os.getlogin)) and as far as I can tell this is the best work around for posix systems.

See [here](http://code.activestate.com/lists/python-list/288845/) for some more info. If you are interested in a better cross-platform approach, I also found [this](http://stackoverflow.com/questions/842059/is-there-a-portable-way-to-get-the-current-username-in-python/842096#842096) suggesting the `getpass` module.

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -10,6 +10,15 @@ import copy
 from datetime import datetime
 import math
 import os
+import pwd
+import re
+try:
+    from string import letters
+except ImportError:
+    from string import ascii_letters as letters
+import sys
+import warnings
+
 from parmed.constants import TINY, DEG_TO_RAD
 from parmed.exceptions import GromacsError, GromacsWarning, ParameterError
 from parmed.formats.registry import FileFormatType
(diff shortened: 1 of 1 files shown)


swails:
Good enough for me!

hainm:
is there any way to add a test for this?

swails:
It's already tested. It replaces the approach that didn't work in a weird configuration with one that does. 

I don't know that it's possible to get the previous code to fail on Travis. But the new code works in the same places the old code did, which is good enough for me.

hainm:
> which is good enough for me.

okie.

ctk3b:
I'm sure we could dig a bit deeper to find out exactly under which configurations this occurs and try to replicate or mimic those on Travis. However, since this issue doesn't appear to come up very often it's probably not worth the effort at this point.

PR #610, opened 2016-02-18
absolute value for Epsilon

ChayaSt:
Since CharMM uses the negative value for epsilon only to signal that it's indeed an epsilon, we should write out the absolute value for epsilon.


swails:
Yep, good catch!

PR #611, opened 2016-02-18
Give an option of skipping setuptools, since it doesn't work nicely w…

swails:
…ith how

AmberTools sometimes installs Python modules.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -15,12 +15,12 @@ matrix:
     - { os: linux, env: PYTHON_VERSION=3.4 }
     - { os: linux, env: PYTHON_VERSION=3.5 MINIMAL_PACKAGES=yes }
     - { os: linux, env: PYTHON_VERSION=pypy }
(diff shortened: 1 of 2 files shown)


hainm:
What is your reason to switch to setuptools? Parmed worked fine before and it does not require external package for install.

swails:
Setup tools allows the use of entry points for automatic script generation and it knows how to play nicely with Windows. The disutils based install does not create Windows runnable scripts. 

Also, pip complains if you don't use setuptools since distutils doesn't create a package that pip knows how to uninstall. And supporting pip is important to me.

hainm:
just tested on amber repo. works fine with me.

swails:
Thanks!

hainm:
I just realize that if we `source amber.sh` before installing ParmEd with setuptools, thing will be fine; because $AMBERHOME/lib/ would be added to PYTHONPATH. :D

hainm:
**updated**: nvm about above comment.

PR #612, opened 2016-02-18
skip dihedral writing only if no propers AND periodic impropers

rafwiewiora:
Sorry, correcting own mistake from past PR!


swails:
One of the coveralls uploads failed. No big deal

swails:
Ah! Yes, this is necessary :)

PR #620, opened 2016-02-23
Fix unknown_functional setting in deserialized Structure instances

### LINKED ISSUE #618: deepcopy(Structure()) behavior and pickling methods
@ChayaSt noticed that, if you `copy.deepcopy` a `Structure` instance, `.createSystem()` on the copy fails because the `.unknown_functional` attribute does not exist: https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L1808

Of course I got obsessed over this and had to work out what was going on. My understanding here is that, as `Structure` doesn't have a `__deepcopy__` method, `copy.deepcopy(Structure())` will fall back to using the `__getstate__` and `__setstate__` pickling methods. And those happen to be missing the `unknown_functional` attribute, as well as the `space_group`. 

For the copying I guess you could just say - just use `copy.copy` because the `__copy__` method provides deep copying behavior. But for pickling this is definitely a bug it seems, you're getting a broken Structure after pickling-unpickling.

swails:
Fixes #618 

@ChayaSt and @rafwiewiora -- can you check that this fixes the deepcopy issues you were seeing?

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -3461,6 +3461,7 @@ class Structure(object):
                        _box=self._box,
                        nrexcl=self.nrexcl,
                        _combining_rule=self._combining_rule,
+                       unknown_functional=self.unknown_functional,
         )
         def idx(thing):
             if thing is None: return None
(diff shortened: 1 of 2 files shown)


rafwiewiora:
@ChayaSt can you check your script's working now? 

@swails just wondering if you missed where I said in the issue that `space_group` was missing too or is that not worth adding?

swails:
I missed it, sorry. I'll issue a PR tonight.

rafwiewiora:
Alright!

ChayaSt:
It's working now!
Thanks @rafwiewiora  for tracking this down and @swails for fixing it!

PR #623, opened 2016-02-24
do not write double external bonds if head == tail

rafwiewiora:
Any advantages / disadvantages to using `!=` or `is not`?

This should fix https://github.com/choderalab/openmm/issues/10

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -260,7 +260,7 @@ class OpenMMParameterSet(ParameterSet):
             if residue.head is not None:
                 dest.write('   <ExternalBond atomName="%s"/>\n' %
                            residue.head.name)
-            if residue.tail is not None and residue.tail != residue.head:
+            if residue.tail is not None and residue.tail is not residue.head:
                 dest.write('   <ExternalBond atomName="%s"/>\n' %
                            residue.tail.name)
             dest.write('  </Residue>\n')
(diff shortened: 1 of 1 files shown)


swails:
Compare atoms via `is not`.  Atom comparisons binary operators are there for comparing where they are in the sequence.  If you want to see if the head and tail are the same atom, check that they are exactly the same *instance* (ergo, use `is`).

rafwiewiora:
Ok, good to know! Changed.

swails:
Good question, by the way.  There is an important and sometimes subtle difference between `==` and `is`, and there is almost always a correct one to use.  The rule is `==` for equivalence (see if two objects are equal -- and each object gets to decide what "equal" means) and `is` for identity (objects *don't* get to decide -- two names either are or are not references to the same object).

swails:
So I'm not sure that this is necessarily the behavior we want.  Keep in mind that for these lipid residues, these are all fragments that can be bonded in either direction, and unlike with proteins there is no convention for the ordering.  Won't this just not add *any* external bonds?

rafwiewiora:
> So I'm not sure that this is necessarily the behavior we want. Keep in mind that for these lipid residues, these are all fragments that can be bonded in either direction, and unlike with proteins there is no convention for the ordering. Won't this just not add any external bonds?

It will just add one external bond, rather than double them up. The head section still adds one bond. 

Correct me if I'm wrong - OpenMM doesn't have this sense of head vs tail or directionality, right? It just needs to know that atom can participate in an external bond.

swails:
> It will just add one external bond, rather than double them up.

Oh yea, duh.  Thanks :smiley:

> Correct me if I'm wrong - OpenMM doesn't have this sense of head vs tail or directionality, right? It just needs to know that atom can participate in an external bond.

I'm not sure -- I thought it had a head and a tail.  But we would need to check the code or with @peastman on that.

swails:
I think this is a good change.

rafwiewiora:
great, thanks!

PR #625, opened 2016-03-01
Don't add self exceptions to bond graph exceptions

jchodera:
We were getting warnings like this for molecules containing three-membered rings:
```
/Users/choderaj/miniconda/lib/python2.7/site-packages/parmed/openmm/topsystem.py:428: OpenMMWarning: Detected incomplete exceptions. Not supported.
  OpenMMWarning)
```
This change fixes the construction of [`bond_graph_exceptions`](https://github.com/ParmEd/ParmEd/blob/master/parmed/openmm/topsystem.py#L392-L398) to exclude entries where `parmed` thinks an atom should be excluded from itself.

### DIFF
--- a/parmed/openmm/topsystem.py
+++ b/parmed/openmm/topsystem.py
@@ -392,11 +392,11 @@ def _process_nonbonded(struct, force):
     bond_graph_exceptions = defaultdict(set)
     for atom in struct.atoms:
         for a2 in atom.bond_partners:
-            if atom != a2:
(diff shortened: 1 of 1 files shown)


swails:
You're right.  Good catch!  Can you change `!=` to `is not`?

jchodera:
Done.

swails:
Perfect.  Once Travis and coveralls complete, I'll merge this.  Thanks!

Issue #627, opened 2016-03-02
OpenMMParameterSet writing - remove templates with missing atom types?


rafwiewiora:
Before I've introduced the `write_unused` argument to remove those atom types, which are not being used in the templates existing in the ParameterSet. 

Now, the ions conversion has a somewhat reverse of this issue - `atomic_ions.lib` have a lot of stuff that does not get parameters from the appropriate frcmod, so I would like to remove those templates from the ParameterSet, which cannot be typified. 

I'm going to deal with this on my script level for now just be checking whether any `self.atom_types[atom.type]` returns KeyError and then removing that residue. But I wondered if it'd be useful to add this to ParmEd too? Perhaps we can now actually make it a pruning call that will do both of the jobs (remove types without templates & templates without types)?

swails:
This should be doable at the ParmEd level. This functionality is being developed to serve this purpose (writing ffxml files for OpenMM), so it should do all the tasks that are necessary for that. 

Should we do this automatically in the write function, or maybe only when pruning unused types? Or should we add another argument to toggle that behavior?

I don't want to introduce a new flag where we will only ever (currently) use a single value. If we find ourselves in that situation, that should just become the behavior.

rafwiewiora:
I agree about a new flag - that'll be too much at this point. But I also wouldn't make 'remove all templates without types' the only behavior possible - you might want to just convert a template into ffxml - for example for use with the ffxml-GAFF. I did a imatinib.lib --> imatinib.xml conversion that way for my GAFF testing - I guess you can do the same mol2 --> ffxml using ParmEd? So we need the writing functionality to take templates without atom types too. 

I'm thinking a small change here: https://github.com/ParmEd/ParmEd/blob/master/parmed/openmm/parameters.py#L165-L171

The if `self.atom_types` already gives us 'protection' for when you want to just convert a template, but I also don't want to kill the general 'whatever goes in - goes out' functionality, even personally I want to be able to use the write function without any cleaning within it for debugging / converting some more unusual stuff etc. - this will be useful for the future. But I am pretty sure that whenever I set `write_unused=False` I am doing a clean FF conversion - so how about we expand the meaning of `write_unused=False` to both a) delete unused atom types AND b) delete 'unused' templates? 

Then I would just, in addition to the warning if KeyError is raised on line 169, also do a removal of those templates there `if not write_unused`. Need to change how the typification is done for this though, would you rather:
* have `typeify_templates()` return a set of residues that raised a KeyError?
[...]

PR #628, opened 2016-03-02
Expand meaning of write_unused to removing templates using unavailable atom types

rafwiewiora:
This for #627.

### DIFF
--- a/devtools/travis-ci/install.sh
+++ b/devtools/travis-ci/install.sh
@@ -25,9 +25,10 @@ else # Otherwise, CPython... go through conda
 
     if [ -z "$MINIMAL_PACKAGES" ]; then
         conda create -y -n myenv python=$PYTHON_VERSION \
-            numpy scipy pandas nose openmm pyflakes=1.0.0 coverage nose-timer \
+            numpy scipy pandas nose openmm coverage nose-timer \
             python-coveralls ambermini
         conda update -y -n myenv --all
+        conda install -y -n myenv pyflakes=1.0.0
     else
         # Do not install the full numpy/scipy stack
         conda create -y -n myenv python=$PYTHON_VERSION numpy nose pyflakes=1.0.0 \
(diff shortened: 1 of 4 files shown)


rafwiewiora:
Ok, first commit was a bad idea - it was removing those residues from ParameterSet rather than just not write, forget that. 

I just did it analogically to the type skipping in second commit.

rafwiewiora:
2 Pyflakes failures look unrelated to this?

swails:
Oh come on pyflakes!  It's an upstream change that broke the checks.  That change truly sucks.  I think the problem here is that pyflakes is being upgraded to 1.1.  Can you try to pin it at 1.0.0?  It's installed via pip for pypy and conda for everything else.  Just modify install.sh to pin the versions.  For pip, this would look like

```
pypy -m pip install pyflakes==1.0.0
```

For conda it would look like

```
conda install pyflakes=1.0.0
```

Hopefully that fixes it.

swails:
Also, can you add a test for this?

rafwiewiora:
Ok, I'm on it!

rafwiewiora:
Test added, pyflakes fixed to 1.0.0 (I didn't want to list all packages in the update call again instead of --all so I moved pyflakes to a separate line there). 

Had to add atomic_ions.lib to the test/files, that ok?

swails:
Looks good! Thanks for all the work. I'll merge after Travis passes

rafwiewiora:
Sorry didn't manage to write this before you merged - at a tech conference with no wifi and no signal...

I had to filter warnings in the test function: `warnings.filterwarnings('ignore', category=exceptions.ParameterWarning)` because otherwise it was failing the tests there - I don't understand this fully: are the warnings filtered to error in the tests? And should my test filter it back to 'default' or 'always' or does it get reset between functions?

swails:
When I want to test that a warning is properly triggered, I always put in a `filterwarnings('error', category=WhateverWarning)`, do an `assertRaises`, and then ignore them to test what happens when the warning is ignored.  So what you did is fine.

swails:
I really should move to using `nose`, that has an explicit `assert_warns` method (I don't think `unittest` for Python 2.7 has an `assertWarns` -- I've tried it before IIRC).

rafwiewiora:
Alright, thanks!

PR #630, opened 2016-03-04
Add printing of the C4 parameters

pengfeili1:
Add printing of the C4 parameters, even it is verbose but could help users to check whether the results are correct. Jason, you can try to create an additional variable in the command line of add12_6_4 to turn on/off the verbose output. I suggested to turn it on in default due to even it is verbose it is better to be careful here: since messing up the C4 terms could mess up the whole simulations afterwards.


parmed/tools/add1264.py
```diff
@@ -157,6 +157,17 @@ def params1264(parm, mask, c4file, watermodel, polfile, tunfactor):
                     # There are two C4 terms need to add together between two
                     # different ions
                     result[idx] += c4 / WATER_POL * pol * tunfactor
+
+    # Print the C4 parameters
+    print('Print the C4 parameters (shown with only one decimal place) : ')
+    for i in range(0, ntypes):
```
swails: The 0 in range is redundant, so you can replace this with just range(ntypes). This is what is done everywhere else in ParmEd.

swails:
This printout looks the same as what printLJMatrix provides. This change is not API-friendly (nobody likes library calls to pollute stdout), and seems not to print out *all* atom types that are assigned (since one LJ type usually corresponds to multiple atom types). 

If people are interested in this information, I think they should use printLJMatrix instead. Can you check that command and make sure it satisfies this need?

pengfeili1:
Hi Jason, the change could print out all atom types that are assigned. But I think your printLJMatrix is better, I checked it and found it satisfies this need.

swails:
Excellent.  I'll go ahead and close this PR then.

hainm:
@pengfeili1 
can you replace `for i in range(0, ntypes):` to `for i in range(ntypes):` to make the style consistent in ParmEd?

pengfeili1:
@hainm Thank you Hai. Jason has also suggested me that. Due to printLJMatrix support similar function, this change will not be applied to the code.

PR #632, opened 2016-03-06
Improve testing coverage by installing pysander

swails:
Since there's a conda package for this now, might as well add it to
improve test coverage :)


hainm:
yeah

PR #633, opened 2016-03-07
[WIP] LennardJonesFoce

ChayaSt:
This PR writes out `LennardJonesForce` if NBFIX exists and turns off `NonbondedForce` by zeroing the epsilon as discussed here: https://github.com/pandegroup/openmm/pull/1399#issuecomment-188586338 . 

We should probably add an option for future Charmm ffxml files (for new parameters not in the existing forcefield) to only write out L-J forces this way because most ions used in Charmm use NBFIX. Right now it only checks if there are NBFIXs in the current parameter set.


parmed/openmm/parameters.py
```diff
@@ -466,6 +467,10 @@ def _write_omm_nonbonded(self, dest, skip_types):
                 sigma = 1.0
                 epsilon = 0.0
 
+            if self.nbfix_types:
+                # turn off L-J. Will use LennardJonesForce to use CostumNonbondedForce to compute L-J interactions
+                sigma = 1.0
+                epsilon = 0.0
```
swails: Perfect

swails:
This looks great, @ChayaSt!

swails:
Only thing is can you check on the Travis failure?  One of the tests is failing and I don't know why.

ChayaSt:
My bad. I didn't realize you already had a ```toppar_water_ion.str``` file in the test files and added a newer version which has 11 NBFIXs rather than 6. I'll undo that commit.

swails:
Is this still a WIP?  Let me know if/when you want me to merge it.

ChayaSt:
I think we should wait until the generator is merged. I might still need to add changes here. Also, I want to add an option to write out ffxml files this way even if there's not NBFIX in the file. So if I want to add a str file for a newly parameterized molecule I can write out the ffxml file with the ```<LennardJonesForce>``` tag

swails:
cc pandegroup/openmm#1453.  We're getting close to where we can revisit merging this in :)

swails:
Any news on the status of this?

swails:
Any news on this?  Do we still have plans to merge this in at some point?

jchodera:
@ChayaSt : Can you tackle this on Monday?

ChayaSt:
Yes, I'll take care of this on Monday.

ChayaSt:
I added an option to write out ```<LennardJonesForce>``` and turn of ```sigma``` and ```epsilon``` for the charmm files that don't have ```NBFIX```. This ensures that future charmm files that are converted will be compatible with this design. You can merge this PR.

jchodera:
Thanks!

parmed/openmm/parameters.py
```diff
@@ -151,6 +151,12 @@ def write(self, dest, provenance=None, write_unused=True):
             templates remaining and parameters including those atom types will
             not be written. A ParameterWarning is issued if any such residues are
             found in a).
+        ljforce : bool
+            If True will turn off L-J and use the LennardJonesForce to use
+            CostumNonbondedForce to compute L-J interactions. Default is False. It
+            should be set to True when converting a charmm force field file that doesn't
+            have NBFIX in the file so that the ffxml conversion is compatible with the
+            main charmm36.xml file
```
swails: Add a note that when pair-specific L-J modifications are present (e.g., NBFIX in CHARMM), this behavior is always done and this flag is ignored.

ChayaSt:
Fixed.

swails:
Thanks!  It took awhile (opened in March), but I'll go ahead and merge this.

swails:
Nice work!

Issue #629, opened 2016-03-02
OpenMMParameterSet - repeats of residues


rafwiewiora:
Similar thing again - should it be in ParmEd or in my script?

atomic_ions.lib has some identical residue templates with different names - e.g. `K` and `K+` - now these both get written to the ffxml. This won't cause a bug in OpenMM - but it's kinda this 'undocumented behavior' of ForceField so we shouldn't do this - not even quite sure how to do this yet, one way would be too look for identical residue signatures in OpenMM - but that'd be post-writing the ffxml editing perhaps.

swails:
Why is this unallowed in OpenMM -- where does it cause issues?  The residues have different names.  Last time I checked, OpenMM still matched residues based on name (it just didn't match atoms in each residue based on name).  This is done in Amber to support both non-PDB and PDB standard names (that are common).

jchodera:
OpenMM does *not* match residues based on name---it uses a graph matching algorithm to match based on the connectivity of elements. Residues with identical element graphs are not exactly supported---I'm not sure if that currently triggers an exception or not, but the behavior of which residue gets matched in that case is definitely undefined.

rafwiewiora:
Current situation:

* no matching on names - matching is done based on 'residue signature' (which is just a string of elements in the residue and number of occurrences of the elements, e.g. `C5H10`) + identical bonding. 

* if you load in more than one identical residues: they get registered in the templates dictionary under the same template signature key:

 ``` python
def registerResidueTemplate(self, template):
        """Register a new residue template."""
        self._templates[template.name] = template
        signature = _createResidueSignature([atom.element for atom in template.atoms])
        if signature in self._templateSignatures:
            self._templateSignatures[signature].append(template)
        else:
            self._templateSignatures[signature] = [template]
```

 That's where the behavior becomes 'undefined' because you've just introduced an order in that list that depends on the order of loading in these templates. And a bunch of loops in the ForceField `break` when they have found something - all of this `undefined` behavior really comes from the lack of mechanism to check for duplicates, which I guess partially comes from a decision to increase efficiency by not completing loops. 

 So going further, when you come to match your templates with topology:
 ``` python
def _getResidueTemplateMatches(self, res, bondedToAtom):
        template = None
        matches = None
        signature = _createResidueSignature([atom.element for atom in res.atoms()])
        if signature in self._templateSignatures:
            for t in self._templateSignatures[signature]:
                matches = _matchResidue(res, t, bondedToAtom)
[...]

PR #634, opened 2016-03-07
add write_ident_res for monoatomic ions

rafwiewiora:
This is my proposal for https://github.com/ParmEd/ParmEd/issues/629

Feel free to reject completely of course - this is just for faster illustration that talking in issues. 

It's not the most elegant thing in the world - but I think it belongs here. Happy to move over to the conversion script too.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -117,7 +117,7 @@ class OpenMMParameterSet(ParameterSet):
 
         return new_params
 
-    def write(self, dest, provenance=None, write_unused=True):
+    def write(self, dest, provenance=None, write_unused=True, write_ident_res=True):
         """ Write the parameter set to an XML file for use with OpenMM
 
         Parameters
@@ -151,6 +151,12 @@ class OpenMMParameterSet(ParameterSet):
             templates remaining and parameters including those atom types will
             not be written. A ParameterWarning is issued if any such residues are
             found in a).
+        write_ident_res: bool
+            If False, if multiple identical residues are present, only the first
+            occurence will be written. This is implemented solely for cleaning
+            up monoatomic ions. 'Identical' is not defined for other types
+            of residues - setting this to False will only affect monoatomic residues.
+            Do not use for anything else!
 
         Notes
         -----
@@ -174,6 +180,8 @@ class OpenMMParameterSet(ParameterSet):
         else:
             skip_residues = set()
             skip_types = set()
+        if not write_ident_res:
+            skip_residues = skip_residues.union(self._find_identical_residues())
         if self.atom_types:
             try:
                 self.typeify_templates()
@@ -213,6 +221,19 @@ class OpenMMParameterSet(ParameterSet):
                 for atom in residue.atoms:
                     keep_types.add(atom.type)
         return {typ for typ in self.atom_types if typ not in keep_types}
+
+    def _find_identical_residues(self):
+        # this is meant for monoatomic ions ONLY
+        skip_ident_residues = set()
+        monoatomic_types = set()
+        for name, residue in iteritems(self.residues):
+            if len(residue.atoms) == 1:
+                typ = residue.atom[0].type
+                if typ in monoatomic_types:
+                    skip_ident_residues.add(name)
+                else:
+                    monoatomic_types.add(residue.atoms[0].type)
+        return skip_ident_residues
 
     def _write_omm_provenance(self, dest, provenance):
         dest.write(' <Info>\n')
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -228,11 +228,11 @@ class OpenMMParameterSet(ParameterSet):
         monoatomic_types = set()
         for name, residue in iteritems(self.residues):
             if len(residue.atoms) == 1:
-                typ = residue.atom[0].type
+                typ = residue.atoms[0].type
                 if typ in monoatomic_types:
                     skip_ident_residues.add(name)
                 else:
-                    monoatomic_types.add(residue.atoms[0].type)
+                    monoatomic_types.add(typ)
         return skip_ident_residues
 
     def _write_omm_provenance(self, dest, provenance):


rafwiewiora:
I could just hardcode in the conversion script / here this for sodium, chloride and potassium ions - these are the only ones affected. Is NA or Na+ the new PDB standard?

parmed/openmm/parameters.py
```diff
@@ -214,6 +222,19 @@ def _find_unused_types(self, skip_residues):
                     keep_types.add(atom.type)
         return {typ for typ in self.atom_types if typ not in keep_types}
 
+    def _find_identical_residues(self):
+        # this is meant for monoatomic ions ONLY
+        skip_ident_residues = set()
+        monoatomic_types = set()
+        for name, residue in iteritems(self.residues):
+            if len(residue.atoms) == 1:
+                typ = residue.atoms[0].type
```
swails: Should we add `type` or `atomic_number` to this set?  Since OpenMM does element matching to find duplicate residues, I think the latter is a better choice here.  What do you think?

rafwiewiora: Fantastic point! You reminded me I need to code this for OpenMM - the problem with by-element recognition is what about Fe2+ and Fe3+? I'm perhaps gonna change OpenMM to accept charges for the template matching to resolve that there. But that's also why we have to go either by type here or by (element, charge) - I'd just go with type for simplicity. If you had two monoatomics with the same element, same charge but different types then you're doomed anyway.

swails:
I've given this a little thought, and I think I know what I want to do here.  The end goal is implement a unique hash function for `ResidueTemplate` such that the hashes are equal for residues that OpenMM considers "equal".  Then it would be very easy to filter out the identical residues.

For the time being, let's use this hash function:

```Python
class ResidueTemplate(object):

    ...

    def __hash__(self):
        if len(self) == 1:
            return hash(self[0].atomic_number)
        # TODO implement hash for polyatomic residues
        return id(self)
```

This will implement what we want in a way that will work for monoatomic ions.  Then all we have to do to fully generalize the code is finish the `__hash__` implementation.

Also, my understanding (based on what I learned from you and John) is that OpenMM chokes on residue templates that have identical topologies (in that it results in undefined behavior and is, by the spec, a bug).  So the algorithm I think we should use here is to keep a `set` of residue templates we've written, and if a residue is already in that set, don't write it out.  Don't give users the ability to write out an illegal XML file.  The `__hash__` implementation on `ResidueTemplate` (and `PatchTemplate`) will ensure duplicate residues don't get written.

What do you think?

rafwiewiora:
I like it! But see my line comment re: atomic_number use. Perhaps we should hold off on this until I make the OpenMM changes and we'll make the hash exactly as the criterion that will be used in OpenMM for matching?

swails:
> I like it! But see my line comment re: atomic_number use. Perhaps we should hold off on this until I make the OpenMM changes and we'll make the hash exactly as the criterion that will be used in OpenMM for matching? (I'll do that tomorrow I think)

100% agree.

> I'm good with having the three problematic ions hard-coded in the conversion script for now.
>  Show all checks

That's fine if you want to.  But I'm good with merging in the hash approach as soon as you have it finished as well.  Up to you.

rafwiewiora:
> That's fine if you want to. But I'm good with merging in the hash approach as soon as you have it finished as well. Up to you.

Yeah I meant I will PR the ParmEd hash as soon as I have the OpenMM approach done, so coming soon! Using the hardcode in conversion script only to play around until that's done.

swails:
> Yeah I meant I will PR the ParmEd hash as soon as I have the OpenMM approach done, so coming soon! Using the hardcode in conversion script only to play around until that's done.

Ah, great.  Once we get all of this machinery in place for both Amber and CHARMM processing, we should do a major version bump so you can pin the version in your conversion script.  Something like:

```Python
import parmed
import sys

if pmd.version < (3, 0):
    sys.exit('You need to upgrade to at least ParmEd version 3.0')
```

rafwiewiora:
> Ah, great. Once we get all of this machinery in place for both Amber and CHARMM processing, we should do a major version bump so you can pin the version in your conversion script. Something like

Nice!

rafwiewiora:
OpenMM changes done in https://github.com/pandegroup/openmm/pull/1373 so we're good to go further here. 

For clarity I will define two terms: 'identically matching residues' and 'identical residues'. 

*Identically matching residues* have:
* same elemental signature (same number of atoms of x element)
* same bond graph

*Identical residues*:
* are identically matching
* use the same atom types

We must:
* not allow identical residues under any circumstances, as this will lead to a 'Multiple matching templates' exception in OpenMM on `createSystem`. 
* allow identically matching residues that are not identical - only use case so far is ions of different oxidation states - e.g. Fe2+ and Fe3+. This requires OpenMM's user to specify which template to use on `createSystem`.

(OpenMM user could also specify which template to use in case multiple *identical residues* are present, but that's not justifiable burden - we should make the deletions).

So for the case of monoatomic ions we need to identify on:
* atomic number
* atom type

So should we do this via `__hash__` that uses atomic number as the hash + atom type check in the `OpenMMParameterSet` code?

rafwiewiora:
As far as then generalizing this is concerned I'm thinking about porting over from OpenMM:
* the elemental signature generating function
* the bond graph matching function

Now either we keep the hash for `ResidueTemplate`, the two identifiers above + a new atom type identifier function go into `OpenMMParameterSet` == definition of *identical residues*. Or perhaps we could make an `OpenMMResidueTemplate` subclass of `ResidueTemplate` with the `__eq__` defined this way?

rafwiewiora:
BTW I was just thinking about the `__hash__` in terms of the internal use of the hash value of stuff like dictionary keys and set members. 

Wouldn't defining the `ResidueTemplate` `__hash__` through atomic number mess things up in that two `ResidueTemplates` that are *identically matching* using the above nomenclature would have the same hash value and hence will be unusable as two separate dictionary keys and will not be considered different in a set, even though we want *identically matching* but not *identical* to be different in those ways? 

I might be completely confused here, but perhaps defining an `__eq__` instead of a hash to include atom type and charge will be wiser here. (but then we need to subclass into `OpenMMResidueTemplate`, I imagine you don't want such `__eq__` Parmed wide?)

rafwiewiora:
Last thing I promise, sorry for so much reading. This is a slightly separate thing but let's keep it all here. 

The overloading of templates, also defined in pandegroup/openmm#1373. I went we such design:
``` xml
<Residue name="Ag" overload="1"/>
```
Bigger overload number (`overloadLevel` internally in OpenMM) - that one wins.

I need to be able to set these for only some of the residues in an `OpenMMParameterSet` for my ions - the 2+ ions in 'overloading sets' will get `overload="1"`, while 3+ and 4+ ions in those sets will keep their default of zero - i.e. the `overload` will not be printed in their tags. 

So `ResidueTemplate` needs to carry a `overloadLevel` attribute. Now does that strengthen the argument to make a `OpenMMResidueTemplate` subclass or do you want me to add it to the `ResidueTemplate` only for use in OpenMM?

swails:
I really think we just want to add a `__hash__` function to `ResidueTemplate` to avoid writing duplicate templates (or a separate hasher that can be used like the below).

```Python
printed_residues = set()
for res, templ in iteritems(self.residues):
    if templhasher(templ) in printed_residues:
        # Duplicate
        continue
    printed_residues.add(templhasher(templ))
    # Print templ
```

Those are never used as keys in dicts -- it's currently names that are keys for templates, so having identical hashes will not cause problems where these are currently used.  But it might be safer overall to use something like the snippet above if its only function will be to track printed templates for OpenMM.

rafwiewiora:
OK, but that leaves these questions:

* if add hash that returns atomic number, how to add atom type & charge checking - I'm just gonna assume you're okay with a separate hashed defined in openmm/parameters.py that returns (atomic_number, atom_type, charge) based on your comment above? 

* *the overloading* - where to define a `overloadLevel` attribute

rafwiewiora:
Replacing this with fresh PR https://github.com/ParmEd/ParmEd/pull/639 . Closing.

swails:
The hash should return some unique hash function such that two residue templates whose hashes are equal *are* equal.  If that means each atom is identified via the atomic number, atom type, *and* charge, then the hashing function would have to reflect that.  The most straightforward hash IMO would be some form of tuple (for monoatomic ions, a tuple of atomic #, type, charge would do it).

Basically my strategy was going to be to have the hashing function reduce a residue template into some kind of well-defined tuple (of tuples) that was a consistent representation of the connected graph of bonds and atoms that defined the residue.  Then return the result of `hash(topologytuple)`.

rafwiewiora:
Well sound like what I went for in #639, check that out :)

PR #638, opened 2016-03-10
Fix behavior of rigidWater in OpenMM System creation

### LINKED ISSUE #624: default behavior of rigid waters is different than in OpenMM
In OpenMM, ```createSystem``` has rigid water set to ```True``` by default. While Parmed also has this set to ```True```, there is another option ``` flexibleConstraints=True``` which  adds the bonds and angles regardless of ```constraints``` and calculates those forces and energies while keeping the degrees of freedoms constraint. OpenMM doesn't do this by default. This creates discrepancies in the energies when comparing systems created in OpenMM vs ParmEd.
The output of this script
https://github.com/ChayaSt/openmm/blob/charmm36/devtools/forcefields/charmm/tests/test_ffconversion.py is:
```
pamed energies: [('HarmonicBondForce', 2.238239263973854), ('HarmonicAngleForce', 1.2560985770962343), ('NonbondedForce', -0.17723389838895195), ('CMMotionRemover', 0.0)]
omm eneriges [('HarmonicBondForce', 0.0), ('HarmonicAngleForce', 0.0), ('NonbondedForce', -0.17723389838895195), ('CMMotionRemover', 0.0)]
```
I tried setting ```flexibleConstraints=False```, but I still get the same output.

swails:
Hopefully fixes #624 

@ChayaSt -- can you check that this PR fixes what you were seeing?

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -32,7 +32,6 @@ from parmed.constants import DEG_TO_RAD, SMALL
 from parmed.exceptions import ParameterError
 from parmed.geometry import (box_lengths_and_angles_to_vectors,
         box_vectors_to_lengths_and_angles)
-from parmed.residue import SOLVENT_NAMES
 from parmed.topologyobjects import (AtomList, ResidueList, TrackedList,
         DihedralTypeList, Bond, Angle, Dihedral, UreyBradley, Improper, Cmap,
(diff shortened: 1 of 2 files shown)


swails:
@ChayaSt -- did this fix the behavior you were seeing in #624?

swails:
I'm actually going to go ahead and merge this.  I think it fixes #624, but it certainly fixes a bug.

PR #639, opened 2016-03-10
Do not write identical residues to OpenMM XML // Add support for omm_overloadLevel for templates

rafwiewiora:
Fresh PR for https://github.com/ParmEd/ParmEd/issues/629 + overload support for residue templates (implemented in https://github.com/pandegroup/openmm/pull/1373 and https://github.com/rafwiewiora/openmm/commit/4e9b65cf26f10820f4c46f4fc9dadc32e2264a61 specifically). 

Summary of the overloading for reference:
* residue templates in XML `overload` attribute like such:
``` xml
<Residue name="Ag" overload="1"/>
```
* default is zero, that does not have to be present in the XML
* internal variable called `overloadLevel` so I made it `omm_overloadLevel` in ParmEd
* higher `overloadLevel` wins if names are the same for 'identically matching' templates
* `residue.omm_overloadLevel` will be set for templates in `OpenMMParameterSet` by conversion script

### DIFF
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -70,6 +70,10 @@ class ResidueTemplate(object):
         atom of this residue when this residue is the last in a chain
     groups : list of list(:class:`Atom`)
         If set, each group is a list of Atom instances making up each group
+    omm_overloadLevel : integer
+        For use with OpenMM ResidueTemplates. If OpenMM ForceField is given multiple
+        identically-matching residue templates with the same names it choses
+        (overloads with) the one with the highest overloadLevel. Default is 0.
     """
 
     def __init__(self, name=''):
@@ -84,6 +88,7 @@ class ResidueTemplate(object):
         self.first_patch = None
         self.last_patch = None
         self.groups = []
+        self.omm_overloadLevel = 0
 
     def __repr__(self):
         if self.head is not None:
@@ -739,4 +744,3 @@ class ResidueTemplateContainer(list):
             AmberOFFLibrary.write(self.to_library(), fname, **kwargs)
         else:
             raise ValueError('Unrecognized format for ResidueTemplate save')
-
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -70,10 +70,11 @@ class ResidueTemplate(object):
         atom of this residue when this residue is the last in a chain
     groups : list of list(:class:`Atom`)
         If set, each group is a list of Atom instances making up each group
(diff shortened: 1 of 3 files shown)


parmed/modeller/residue.py
```diff
@@ -84,6 +88,7 @@ def __init__(self, name=''):
         self.first_patch = None
         self.last_patch = None
         self.groups = []
+        self.omm_overloadLevel = 0
```
swails: Let's just keep this named `overload_level`, or perhaps `priority` would be a better name for it?

parmed/openmm/parameters.py
```diff
@@ -214,6 +214,13 @@ def _find_unused_types(self, skip_residues):
                     keep_types.add(atom.type)
         return {typ for typ in self.atom_types if typ not in keep_types}
 
+    def _templhasher(self, residue):
```
swails: This never uses `self` -- either make it a `@staticmethod` or a module-level global function.

parmed/openmm/parameters.py
```diff
@@ -257,11 +264,18 @@ def _write_omm_atom_types(self, dest, skip_types):
 
     def _write_omm_residues(self, dest, skip_residues):
         if not self.residues: return
+        written_residues = set()
         dest.write(' <Residues>\n')
         for name, residue in iteritems(self.residues):
-            if not isinstance(residue, ResidueTemplate) or name in skip_residues:
+            if (not isinstance(residue, ResidueTemplate) or name in skip_residues
+                or self._templhasher(residue) in written_residues):
```
swails: `self._templhasher` might be expensive -- especially when generalized to larger residue templates.  Can you precompute this just inside the `for` loop to avoid calling this function many times?

swails:
Looks good!  I just made a couple suggestions.  Let me know what you think.

rafwiewiora:
Thanks for reviewing! 

* changing name to `overload_level` to keep it the same (except for camelCase --> snake_case) as it is in OpenMM, ok?
* decorating `_templhasher` with `@staticmethod`
* I don't quite understand the expensiveness comment - do you mean take it out of this if clause and put it into a separate one so that some residues don't call it because they're excluded on `name in skip_residues`? I'm gonna delete the `not isinstance(residue, ResidueTemplate)` btw - we're already allowing `ResidueTemplate`'s only on creation of the `OpenMMParameterSet`

rafwiewiora:
OK, all done, this should do it.

swails:
Looks good!

PR #645, opened 2016-03-15
parse amber netcdf having no box info

hainm:
fix 
```python
    637     def cell_lengths_angles(self):
--> 638         return np.hstack((self._ncfile.variables['cell_lengths'][:],
    639                           self._ncfile.variables['cell_angles'][:]))

KeyError: 'cell_lengths'
```


test/test_parmed_amber.py
```diff
@@ -1600,6 +1601,14 @@ def test_auto_detection(self):
             f.write('Only one line\n')
         self.assertFalse(asciicrd.AmberAsciiRestart.id_format(fn))
 
+    def test_parsing_netcdf_file_without_unitcells(self):
```
swails: This test needs a `unittest.skipIf(PYPY, 'Cannot run test under PYPY')`.  Or something like that -- the other NetCDF tests in test_parmed_amber.py have examples.

test/test_parmed_amber.py
```diff
@@ -1600,6 +1601,14 @@ def test_auto_detection(self):
...
+        """Parsing Amber netcdf file that does not have box info"""
+        fn, tn = get_fn('tz2.nc'), get_fn('tz2.parm7')
+        parm = pmd.load_file(tn, fn)
+        self.assertEqual(parm.get_coordinates().shape, (101, 223, 3))
+        self.assertEqual(len(parm.atoms), 223)
+
```
swails: What about adding `self.assertIs(parm.box, None)`?

swails:
Good catch and fix.  Since coverage ran, I suspect just the pypy tests failed.  Until numpypy supports the necessary features for `netcdf.py`, we need to skip NetCDF tests when using pypy.

hainm:
> I suspect just the pypy tests failed

yes, only pypy failed. I just updated
- skip if PYPY
- None test

hainm:
my hidden aim for this PR is to use ParmEd as one of the [backend] (https://github.com/arose/nglview/pull/44)  for protein viewer (given that pytraj does not work in Win while nglview and ParmEd work well). 

My broader aim is to help the development of ParmEd easier (like trying to quickly visualize the addition of N/C terminal in ParmEd (if exist))

arose:
> @arose: tag you here to share my thought.

Never used ParmEd but if it helps I am all for it

swails:
Thanks!

PR #646, opened 2016-03-16
Mass should be case-insensitive

ChayaSt:
I ran into a CharMM stream file where MASS is written as Mass. 
Here's the file:
https://github.com/ChayaSt/Charmm36_FF/blob/master/toppar/stream/misc/toppar_ions_won.str#L31-L95

### DIFF
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -321,25 +321,25 @@ class CharmmParameterSet(ParameterSet):
                 parameterset = line.strip()[1:78]
                 continue
             # Set section if this is a section header
-            if line.startswith('ATOM'):
+            if line.upper().startswith('ATOM'):
                 section = 'ATOMS'
                 continue
-            if line.startswith('BOND'):
+            if line.upper().startswith('BOND'):
                 section = 'BONDS'
                 continue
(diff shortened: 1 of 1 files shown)


swails:
Is this just true for MASS, or will the other sections need similar modifications?

ChayaSt:
I went through all the files and haven't found a different section that needs this modifications. However, I was wondering if we should just make all section titles case-insensitive? Since the files are obviously not consistent.

ChayaSt:
There is another issue I came across.

When you load many files in ```CharmmParameterSet```, you should be able to look up the patches in the   patches dict that's bound to the class if the patches were read in a different file. However, the way the code works now is that ```read_topology_file```  creates a new dict patches internal to the function and some patches are not found. I can change that too.

swails:
I think CHARMM conventions is to ignore case. So let's make all section IDs case insensitive.

I'm not sure that I see the difference in your patch handling and what was there before, but yours is arguably the better way to do things anyway.

swails:
Excellent!  This looks like it's good to go, yes?

ChayaSt:
Yes.

swails:
Excellent.  In it goes.

PR #648, opened 2016-03-22
Leave out element for dummy atom

ChayaSt:
If the atom type is a dummy atom, leave out element from attributes so that forcefield.py can parse the xml.


swails:
Thanks!

PR #649, opened 2016-03-23
Add warning when replacing residues

ChayaSt:
Turns out CharMM has many duplicate residues with different atom types across different force fields that should be compatible. I didn't check yet if the parameters are identical for these atom types, but I thought it's good to get a warning when they get overwritten. I still need to figure out how to handle this. If the parameters are all identical, it shouldn't matter. But if they're not, we might need to write out both and use  overload.

### DIFF
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -791,6 +791,9 @@ class CharmmParameterSet(ParameterSet):
                     # Get the residue definition
                     words = line.split()
                     resname = words[1].upper()
+                    if resname in (self.residues):
+                        warnings.warn('Replacing residue %r' %
+                                              (resname), ParameterWarning)
                     # Assign default patches
                     hpatches[resname] = hpatch
                     tpatches[resname] = tpatch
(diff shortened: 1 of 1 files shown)


ChayaSt:
In that line I am checking if the residue that is being parsed in the current file already exists in ```CharmmParameterSet```

hainm:
@jswails: what's about turn off coverage report un PR? Kind of distracting.

swails:
The failure here is fake.  Is this still needed?  And if so, good to merge?

hainm:
@ChayaSt 

```python
if resname in self.residues:
```
I am kind of confused. Is `resname` string and self.residues ResidueList? if yes for both, this is somewhat dangerous.

swails:
`self.residues` is a dict matching a string with either a `ResidueTemplate` or `ResidueTemplateContainer`.

hainm:
I see.

ChayaSt:
> Is this still needed? And if so, good to merge?

Yes.

Issue #651, opened 2016-03-29
segment gets changed when splitting PSF file


ajasja:
Hi!

I'm trying to split a PSF file.
I'm using version 2.1.10.

```python
import mdtraj as md
import parmed as pmd

t=md.load("TLR-flag.pdb",top='TLR-flag.psf')
p_ind = t.top.select("segname TLR")

cf=pmd.load_file('TLR-flag.psf')
cf[p_ind].save('test1.psf', vmd=True)
```

The segment name changes from `TLR` to `SYS`. Is this intended or a bug?

Examples are attached: [TLR.zip](https://github.com/ParmEd/ParmEd/files/193710/TLR.zip)

swails:
Ideally, ParmEd should not lose information.  I'll look into why this is happening.

hainm:
may be parmed can add 

```python
cf.take(atom_indices, keep_segid=True)
```

The slicing `cf[p_ind]` does not keep segid info.

ajasja:
@swails The funny thing is that if I save the whole psf (so `cf.save('tets.psf', vmd=True)`) then the segids are preserved. So [here](https://github.com/ParmEd/ParmEd/blob/master/parmed/formats/psf.py#L133) the `atom.residue.segid` is not available (but I don't know why)

@hainm I would expect the slicing to keep the segid info by default.

hainm:
> but I don't know why

no worry, @swails always know the answer. :D 

> I would expect the slicing to keep the segid info by default.

ah yes.

ajasja:
@swails Aha, got it:
When the atom is added [here](https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L903) the segid is not passed. (this also appears in several other places)

swails:
That's exactly what I was expecting it to be.  I'm fixing it now.

As for the `pwd` module -- that is Unix only.  I need to find an alternative for Windows.  Perhaps it would be helpful if I got appveyor running...

Thanks for the report!

Also, your hack seems entirely reasonable.  IMO, CHARMM segid and the CHAIN ID are the same thing, except CHARMM clearly didn't want to be limited by the 26-chain, 1-letter limit :).  I'll implement your suggested change when I have time, or you can file a PR yourself.  Let me know what you'd like to do.

swails:
OK.  I opened PRs to address both issues you raised here.  When they pass, I'll go ahead and merge them in.  If you can, give them a try and see if they work?

Thanks!

[...]

PR #653, opened 2016-03-30
Hopefully fix ParmEd on Windows

swails:
This removes some Unix-only functions being used in the header for GROMACS topology files.

This should hopefully fix the issues reported in #651 relating to `pwd`

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -10,7 +10,6 @@ import copy
 from datetime import datetime
 import math
 import os
-import pwd
 import re
 try:
     from string import letters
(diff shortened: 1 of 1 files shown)


ctk3b:
Might also be worth checking out the `getpass` approach mentioned in #607 

http://stackoverflow.com/questions/842059/is-there-a-portable-way-to-get-the-current-username-in-python/842096#842096

swails:
Ack.... I forgot that I was actually reversing your fix.. :(.  I'll hold off on merging this until I have a better approach.  Thanks for the reminder.

ajasja:
Hi @swails!
With this PR I can at least install ParmEd on windows using `python setup.py develop`
So at least something in that direction should be done. Since this is just needed for the header in some file, pwd could be used on unix, but the win32.GetUser() on windows.

Wrapping the import of pwd in exception handling should also probably work...

swails:
#655 replaces this PR.  That keeps the existing behavior intact while using workarounds for Windows.  The only reason I did this first was because I hadn't remembered that that code was there because @ctk3b was fixing a different issue (I thought I had made it overly complex myself).

There might be a more cross-platform approach than #655 (?), but that should be more than flexible enough for all users without introducing a regression in some fancy sessions.

PR #655, opened 2016-03-31
Hopefully fix Windows without getting rid of the fix for Unix OSes.

swails:
This replaces #653


ajasja:
Unfortunately `os.getlogin()` does not exist. Can I do a PR on top of your PR? 
I can fix this tomorrow. (either using an environment variable or win32). I'd go for the former approach since it's not critical that it's the correct user name.

hainm:
you're right about `getlogin`

https://docs.python.org/2/library/os.html#os.getlogin

hainm:
you're right about `getlogin`

https://docs.python.org/2/library/os.html#os.getlogin

PR #658, opened 2016-04-02
Fixed PSF parsing when negative residue numbers are present.

swails:
See https://simtk.org/forums/viewtopic.php?f=161&t=6512&p=16820#p16820 where Henry van den Bedem reported the same issue for OpenMM (and the PSF parser in OpenMM is an early incarnation of the version I originally wrote for ParmEd).


swails:
Ugh... Still coveralls that's failing for some stupid reason.

PR #660, opened 2016-04-04
update clean method to compeletely remove temp files

hainm:


hainm:
so in AMBER, just need to (cd parmed && $(PYTHON) setup.py clean) to remove temp files.

swails:
Looks reasonable to me.  Thanks!

setup.py
```diff
@@ -1,6 +1,11 @@
 import os
 import sys
 
+if sys.version_info < (2, 7):
+    sys.stderr.write('You must have at least Python 2.7 for ParmEd to work '
+                     'correctly.\n')
+    sys.exit(0)
```
swails: This needs to be non-zero.

hainm: updated.

PR #666, opened 2016-04-08
Make ParameterSet.urey_bradley_types keys consistent everywhere

### LINKED ISSUE #665: Parameter set from structure
When I try creating a ```ParameterSet``` from ```structure```, the urey bradley terms don't copy over correctly. There are only two atom types in the key (1,3 atoms) instead of all three atom types for the angle. This happens because ```structure.urey_bradleys``` only keeps the 1st and 3rd atom of the angle. 

The fix I'm suggesting is keep the middle atom in ```structure.urey_bradleys``` so that you can get the identical ```ParameterSet```  used to parameterize ```structure````

swails:
Some places used 2 atom types as keys, others used 3.  This updates the code to use 3 atom types everywhere.

Fixes #665.


swails:
This actually fixed a bug I didn't realize existed in the GROMACS topology file parser with the CHARMM force field.  In particular, because the Urey-Bradley terms only had 2 keys, it would sometimes match angles it shouldn't match since the middle atom was different (but not tracked).  After this change, the Urey-Bradley energies agree.  The original validation didn't consider CHARMM systems, which is how this slipped through.

So in some form, this needs to go in.

swails:
Is this ready to merge, then?

ChayaSt:
I think so. If I run into more problems we can reopen this.

PR #667, opened 2016-04-08
[WIP] Urey Bradley term fix

ChayaSt:
The changes in this PR adds the middle atom to the ```structure.urey_bradleys``` list.

This is not working yet because of [this](https://github.com/ParmEd/ParmEd/issues/665#issuecomment-207151978)

I'm not sure if these changes are the best way to handle this.

### DIFF
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -222,8 +222,8 @@ class CharmmParameterSet(ParameterSet):
         periodicity found if no matching DihedralTypeList is found.
         """
         return cls.from_parameterset(
-                ParameterSet.from_structure(
-                        struct, allow_unequal_duplicates=False)
+                ParameterSet.from_structure(struct,
+                                            allow_unequal_duplicates=False)
         )
 
     @classmethod
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -222,8 +222,8 @@ class CharmmParameterSet(ParameterSet):
         periodicity found if no matching DihedralTypeList is found.
         """
         return cls.from_parameterset(
-                ParameterSet.from_structure(struct,
-                                            allow_unequal_duplicates=False)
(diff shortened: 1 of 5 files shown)


swails:
I really don't think we should add `atom3` to `UreyBradley`.  There are *so* many places this would require changing that it would be a huge job.  Not to mention the fact that it represents an API break.  Furthermore, Urey-Bradley terms really *are* 2-body terms, and the current data structure highlights that fact.  Lastly, it's easy to get the central atom from `atom1` and `atom2` -- [see here](https://github.com/swails/ParmEd/blob/ureybrad/parmed/parameters.py#L478).

The main problem we're trying to solve here is how do we key and match Urey-Bradley terms.  The easiest and safest way to do that is with 3 atom types, and that was being done in the CHARMM code, but not the GROMACS code.

Give #666 a try and see if that works for you.  I also added what I hope is a fix for [the issue you already mentioned](https://github.com/ParmEd/ParmEd/issues/665#issuecomment-207151978).

ChayaSt:
> I really don't think we should add atom3 to UreyBradley. There are so many places this would require changing that it would be a huge job. Not to mention the fact that it represents an API break. Furthermore, Urey-Bradley terms really are 2-body terms, and the current data structure highlights that fact. Lastly, it's easy to get the central atom from atom1 and atom2 -- see here.

I agree with you. I merged your changes into my branch and will work with that.

swails:
I *just* pushed the fix to the missing Urey-Bradleys.  I'm not sure when you merged my branch in, but if it wasn't within the last 5 minutes, can you do it again?

swails:
Ack, I just pushed a fix to the missing U-B "fix".  I think it should work now.

ChayaSt:
Yes, it's working now. Thanks!

ChayaSt:
Closing this PR. https://github.com/ParmEd/ParmEd/pull/666 fixes this.

Issue #34, opened 2015-01-20
Automatic bond determination in PDB and mm/CIF parsing.


swails:
It would be nice to have some kind of automatic bond determination.  This should come, ideally, from residue template libraries if available, and then basic bond-by-distance if the non-standard residue is unavailable.

It might be nice to add the 'components.cif' file from the PDB somewhere to significantly expand the number of residues for which we have a known bonding pattern.

swails:
Also, the `CONECT` records should be read in PDBs

davidlmobley:
While this is fine for proteins, it will need to get a lot more complicated for ligands/hetero atoms or you may just not want to deal with it at all in that case. It's not uncommon for these to get refined with somewhat strained bond distances so inferring bond types for ligands can be tricky. The best solutions I've seen also use local geometry (i.e. bond angles and configuration, i.e. tetrahedral/etc.), and even those sometimes go wrong. The challenge is in part that you don't always know what compound you're dealing with (in terms of protonation state/bond types) and so sometimes you have to guess this based on the geometry you're provided.

swails:
My plan is to treat this the same way that LEaP does (and I presume Gromacs as well): bonds are determined from templates (the `ResidueTemplate` object defined in the `modeling` package).  I'll add templates for the standard residues (amino and nucleic acids) and assign bonds from those.

Ligand(/non-standard residue) bonds I can handle a number of ways:

- Assign from distances (with the caveats you mentioned)
- Do nothing (but respect `CONECT` records, which in RCSB-downloaded PDB files should specify all of the bonds)
- Optionally go to the wwPDB, download the chemical component CIF file, create a ResidueTemplate from that, and assign bonds that way.

I might also go ahead and survey the PDB to pull out the most common chemical components and create ResidueTemplates for those.

It won't work for everything, but it will work for a lot.

hainm:
+1 for guess_bond. cpptraj does has it.

```python
In [5]: pdb = io.loadpdb_rcsb("1o15")

In [6]: pdb
Out[6]:
<pytraj.Trajectory with 1 frames: <Topology with 2 mols, 34 residues, 1086 atoms, 
1170 bonds, PBC with box type = ortho>>
```

`parmed`
```python
In [2]: t = pmd.download_PDB("1o15")

In [3]: t
Out[3]: <Structure 1086 atoms; 34 residues; 0 bonds; PBC; NOT parametrized>
```

swails:
[...]

Issue #662, opened 2016-04-06
Angle determination from OpenMM topology


pgrinaway:
I'm wondering if it would be feasible to allow `parmed.openmm.load_topology(topology)` to identify angles as well--currently it seems to only do that if there is a `System` as well. I realize the presence of a `System` makes that more straightforward, since angle forces have been assigned, but if it's not too hard, then that would be a great feature to have.

swails:
This is on the list of desired functionality.  For me, it's behind #34, which has been a *long* time in coming.  Basically what I'd do is add a method on `Structure` like `fill_bonded_terms` to generate the angles and dihedrals from the bond graph (not impropers, though... nobody assigns those the same anyway :)).

Issue #355, opened 2015-08-17
[wishlist] add 'TER' keyword with `save` to pdb


hainm:
In [3]: p = pmd.download_PDB('3gx5')

In [4]: p
Out[4]: <Structure 2193 atoms; 207 residues; 0 bonds; PBC (orthogonal); NOT parametrized>

In [7]: p0 = p['(!:K)&(!:MG)&(!:SAM)']

In [8]: set([res.name for res in p0.residues])
Out[8]: {'A', 'C', 'G', 'HOH', 'U'}

In [9]: p0.save('3gx5_only_nuc_and_hoh.pdb')

`cpptraj` has 'TER' keyword when saving pdb file (life below)
```
ATOM   2051  O   HOH A 115       3.578  -2.703   8.282  1.00 56.04           O
TER    2052      HOH A 115
ATOM   2052  O   HOH A 116      24.458 -10.733  23.188  1.00 51.39           O
TER    2053      HOH A 116
ATOM   2053  O   HOH A 117       0.118 -19.394   5.378  1.00 79.30           O
```

hainm:
if not having 'TER': tleap will give wrong charge.

swails:
I know.  I really want this to wait on #34, though to make sure it's as close as I can get it to the actual PDB standard (basically equivalent to cpptraj's `terbyres` keyword).

hainm:
what do you think about adding option to add 'TER' keyword between specific residues?

hainm:
ok, I found myself `res.ter = True`

PR #668, opened 2016-04-10
Add template-based bond determination

swails:
This is the start of the implementation of #34 (!!), and is the first step in ultimately tackling #355 and #662 

Things to do:

- [x] Handle atoms and residues not in templates
- [x] Figure out how to stitch together polymeric residues
- [x] Process CONECT records in PDB files ~~and the equivalent in mmCIF files~~
- [x] Implement the distance-based bonding for atoms and residues not present in templates
- [x] Special-case disulfide bonds and sulfur linkers

EDIT: More to-dos

- [x] Automatic bond assignment for GRO files
- [x] Improve element matching for GRO files based on residue templates (!! nice!)

EDIT: Even more

- [x] Implement a pairlist to drastically improve efficiency of bond searching


parmed/modeller/standardtemplates.py
```diff
@@ -0,0 +1,9 @@
+""" Standard residue templates for biomolecular residues """
+
+import os
+from parmed.amber.offlib import AmberOFFLibrary as _parser
```
hainm: you used _parser to avoid dumping AmberOFFLibrary in this namespace?

swails: Yea, more or less. :smiley:

swails:
OK, this should be done and good to go.  I'm punting on the mmCIF CONECT stuff for now, since I don't know the spec currently and the bond-by-distance stuff takes care of most of what that would miss (like disulfides, sulfide-linking, and non-standard residue bonding patterns.

I'm good to merge this when it passes.

Closes #34

swails:
Actually, I found a couple more things to do.  Once I get to those, then this is good to merge.

swails:
Ack... I really need to implement a fast, numpy-vectorized, grid-based pairlist generator for the unassigned atoms whose bonds are assigned by distance.  Otherwise the bond assignment scales as O(N^2).

ctk3b:
[This](https://github.com/mdtraj/mdtraj/blob/master/mdtraj/core/topology.py#L1022) might help a little, if I correctly understand what you're trying to do right now, but there's certainly more room for improvement.

swails:
I considered that method, but I was worried about the potential memory and time requirements when systems get large. I was thinking something like a very simple grid/voxel-based approach.

The approach I'm thinking of should scale as O(N) in both memory and time with a reasonably small time prefactor (since the pairlist can be restricted to atoms within 2.5 angstroms of each other).  It should also be fast enough to do in pure Python...  But we'll see.  It would seem a shame to have to start requiring C extensions just for this.

swails:
It's slower than I'd like, but acceptable for now.  It'll make parsing huge PDB files with many unknown residue types (such as those for bilayers and carbohydrates) ~4x slower (this is for the 117K atom CHARMM segid overflow PDB test with >60K atoms not in standard templates).

The pairlist is only built for atoms whose bonds still need to be assigned after going through the templates, so this should perform very well in the majority of cases, and acceptable (for now) in pathological ones.

swails:
This should be good to go after the tests pass. Whew!

PR #672, opened 2016-04-14
Added a copy option to load_parameters

ChayaSt:
This PR adds the option not to create a deep copy of the ```ParameterSet``` when calling ```load_parameter```. This save substantial time as profiled here: https://github.com/choderalab/torsionfit/pull/25#issuecomment-209653537

### DIFF
--- a/parmed/charmm/psf.py
+++ b/parmed/charmm/psf.py
@@ -448,7 +448,7 @@ class CharmmPsfFile(Structure):
 
     #===================================================
 
-    def load_parameters(self, parmset):
+    def load_parameters(self, parmset, copy=True):
         """
         Loads parameters from a parameter set that was loaded via CHARMM RTF,
         PAR, and STR files.
@@ -457,6 +457,10 @@ class CharmmPsfFile(Structure):
         ----------
         parmset : :class:`CharmmParameterSet`
             List of all parameters
+
+        copy : bool
+            if False, parmset will not be copied. This will silently modify the references to types in the structure
+            type_list so use with caution. Default is True
 
         Notes
         -----
@@ -473,7 +477,8 @@ class CharmmPsfFile(Structure):
         ------
         ParameterError if any parameters cannot be found
         """
-        parmset = _copy(parmset)
+        if copy:
+            parmset = _copy(parmset)
         self.combining_rule = parmset.combining_rule
         # First load the atom types
         for atom in self.atoms:
--- a/parmed/charmm/psf.py
+++ b/parmed/charmm/psf.py
@@ -448,7 +448,7 @@ class CharmmPsfFile(Structure):
 
     #===================================================
 
-    def load_parameters(self, parmset, copy=True):
+    def load_parameters(self, parmset, copy_parameters=True):
         """
         Loads parameters from a parameter set that was loaded via CHARMM RTF,
         PAR, and STR files.
@@ -458,9 +458,31 @@ class CharmmPsfFile(Structure):
         parmset : :class:`CharmmParameterSet`
             List of all parameters
 
-        copy : bool
-            if False, parmset will not be copied. This will silently modify the references to types in the structure
-            type_list so use with caution. Default is True
+        copy_parameters : bool
+            if False, parmset will not be copied.
+
+            Warning:
+            -------
+            Not copying parmset will allow ParameterSet and Structure to share reference to types.
+            If you modify the original parameter set, the references in Structure list_types will be silently modified.
+            However, if you change any reference in the parameter set, then that reference will no longer be shared with
+            structure.
+
+            Example where the reference in ParameterSet is changed. This will NOT modify the paraemters in the psf.
+
+            psf.load_parameters(parmset, copy_parameters=False)
+            parmset.dihedral_type[('a1', 'a2', a3', a4)] = DihedralType(1, 2, 3)
+
+            This WILL change the parameter in the psf because the reference has not been changed in ParameterSet
+
+            psf.load_parameters(parmset, copy_parameters=False)
+
+            d = parmset.dihedral_types[('a1', 'a2', 'a3', 'a4')]
+            d.phi_k = 10
+            d.per = 2
+            d.phase = 180
+
+            Use with caution!
 
         Notes
         -----
@@ -477,7 +499,7 @@ class CharmmPsfFile(Structure):
         ------
         ParameterError if any parameters cannot be found
         """
-        if copy:
+        if copy_parameters:
             parmset = _copy(parmset)
         self.combining_rule = parmset.combining_rule
         # First load the atom types
--- a/parmed/charmm/psf.py
+++ b/parmed/charmm/psf.py
@@ -461,26 +461,34 @@ class CharmmPsfFile(Structure):
         copy_parameters : bool
             if False, parmset will not be copied.
 
-            Warning:
+            WARNING:
             -------
-            Not copying parmset will allow ParameterSet and Structure to share reference to types.
+            Not copying parmset will cause ParameterSet and Structure to share references to types.
             If you modify the original parameter set, the references in Structure list_types will be silently modified.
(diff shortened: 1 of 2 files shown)


parmed/charmm/psf.py
```diff
@@ -458,6 +458,10 @@ def load_parameters(self, parmset):
         parmset : :class:`CharmmParameterSet`
             List of all parameters
 
+        copy : bool
+            if False, parmset will not be copied. This will silently modify the references to types in the structure
```
hainm: "This will **silently** modify the references to types in the structure" --> sounds dangerous.

jchodera: Or useful!

parmed/charmm/psf.py
```diff
@@ -458,6 +458,10 @@ def load_parameters(self, parmset):
         parmset : :class:`CharmmParameterSet`
             List of all parameters
 
+        copy : bool
```
swails: `copy` is very vague.  Can you change this to `copy_parameters` or `copy_parameterset`?

parmed/charmm/psf.py
swails: I agree with both of you.  It's a perfect opportunity to shoot yourself in the foot.  For instance:

```Python
psf.load_parameters(parmset, copy_parameters=False)

parmset.dihedral_types[('a1', 'a2', 'a3', 'a4')] = DihedralType(1, 2, 3)
```
The above does *not* change the parameter in the `psf` instance.  What `copy_parameters=False` really does is *allow* the paramter set and the structure to share references, but it doesn't enforce it.  So if you change any of the references in the parameter set (i.e., via assignment), then that reference is no longer shared with the structure you loaded the parameters into.  However, the following:
```Python
psf.load_parameters(parmset, copy_parameters=False)

d = parmset.dihedral_types[('a1', 'a2', 'a3', 'a4')]
d.phi_k = 10
d.per = 2
d.phase = 180
```
This *does* change the parameter in `psf` (if it was assigned to one of the torsions, of course).  So this method is very dangerous.  However, it allows someone that knows what they're doing to implement workflows orders of magnitude more efficiently than without it.  That's what ultimately convinced me this was worth the risk.

That said, this needs a better description of what's actually being done here.  @ChayaSt -- can you change the docstring to mention that what this flag really does is to have the PSF share references to types in the parameter set, so if the references are changed in the parameterset, the corresponding parameter in the structure will be changed *and vice versa*.

swails:
In addition to the new docstring, this also needs a test.

ChayaSt:
I'm seeing some weird behavior with the tests I'm writing. 

Here's the test I tried

```
def test_copy_parameters(self):
        """ Tests copy_parameters option in load_parameters """

        top = psf.CharmmPsfFile(get_fn('ala_ala_ala.psf'))
        top.load_parameters(parmset=param22, copy_parameters=False)
        b = param22.bond_types[(top.atoms[0].type, top.atoms[1].type)]
        b.k = 200
        a = param22.angle_types[(top.atoms[1].type, top.atoms[0].type, top.atoms[2].type)]
        a.k = 20
        d = param22.dihedral_types[('X', top.atoms[4].type, top.atoms[6].type, 'X')]
        d[0].phi_k = 0.300
        assert(top.bonds[0].type == param22.bond_types[(top.atoms[0].type, top.atoms[1].type)])
        assert(top.angles[0].type == param22.angle_types[(top.atoms[1].type, top.atoms[0].type, top.atoms[2].type)])
        assert(top.dihedrals[0].type == param22.dihedral_types[('X', top.atoms[4].type, top.atoms[6].type, 'X')])

        param22.bond_types[(top.atoms[0].type, top.atoms[1].type)] = BondType(300, 1.040)
        param22.angle_types[(top.atoms[1].type, top.atoms[0].type, top.atoms[2].type)] = AngleType(k=40, theteq=109.5)
        param22.dihedral_types[('X', top.atoms[4].type, top.atoms[6].type, 'X')][0] = DihedralType(phi_k=0.200, per=3,
                                                                                                   phase=0.00, scee=1.00,
                                                                                                   scnb=1.00)
        assert(top.bonds[0].type != param22.bond_types[(top.atoms[0].type, top.atoms[1].type)])
        assert(top.angles[0].type != param22.angle_types[(top.atoms[1].type, top.atoms[0].type, top.atoms[2].type)])
        #assert(top.dihedrals[0].type != param22.dihedral_types[('X', top.atoms[4].type, top.atoms[6].type, 'X')])
```

I got an ```AssertionError``` with the last ```assert```. Turns out, when I change the ```dihedral_type``` through assignment, it still changes the dihedral parameters in the structure. But this doesn't happen with the bonds nor angles. When I comment out the last assert, the test passes.

parmed/charmm/psf.py
```diff
@@ -458,6 +458,32 @@ def load_parameters(self, parmset):
         parmset : :class:`CharmmParameterSet`
             List of all parameters
 
+        copy_parameters : bool
+            if False, parmset will not be copied.
+
+            Warning:
+            -------
+            Not copying parmset will allow ParameterSet and Structure to share reference to types.
```
swails: allow -> cause; reference -> references

parmed/charmm/psf.py
```diff
@@ -458,6 +458,32 @@ def load_parameters(self, parmset):
...
+            If you modify the original parameter set, the references in Structure list_types will be silently modified.
+            However, if you change any reference in the parameter set, then that reference will no longer be shared with
+            structure.
+
+            Example where the reference in ParameterSet is changed. This will NOT modify the paraemters in the psf.
+
+            psf.load_parameters(parmset, copy_parameters=False)
+            parmset.dihedral_type[('a1', 'a2', a3', a4)] = DihedralType(1, 2, 3)
```
swails: Use `angle_types` as an example instead of `DihedralType` -- this should actually be a `DihedralTypeList`, since `dihedral_types` is a Fourier sequence.

swails:
You should use `self.assertEqual` to go through the unittest framework rather than using the raw assert.

That said, what you fell into was the one "major" risk I didn't describe -- for `dihedral_types`, `ParameterSet` stores `DihedralTypeList`, which is a `list` subclass that has extra goodies attached to it.  *This* is the reference that `CharmmPSFFile` tracks when you call `load_parameters`.  What you did was assign a new instance of `DihedralType` to an existing `DIhedralTypeList`.  While this changes the reference for the first term in that list, it doesn't change the reference shared between the parameter set and structure.

This is what you're doing for the bonds and angles:

```python
>>> class CustomObject(object):
...     def __init__(self, value):
...         self.value = value
... 
>>> a = CustomObject(10)
>>> b = a
>>> a.value
10
>>> b.value = 15
>>> a.value
15
>>> b = CustomObject(20)
>>> b.value
20
>>> a.value # not changed!
15
```

But for dihedrals, it's like this:

```Python
>>> la = [CustomObject(10), CustomObject(20)]
>>> lb = la
>>> lb[0].value = 25
>>> la[0].value
25
>>> lb[0] = CustomObject(0)
>>> lb[0].value
0
>>> la[0].value
0
```

As I said, *not* copying is very risky.  As you've seen, it tricks and traps even experienced Python programmers.

ChayaSt:
> As I said, not copying is very risky. As you've seen, it tricks and traps even experienced Python programmers.

Yeah, I see that. I'll add a note about ```DihedralTypeList``` to the warning in the docstring.

swails:
What it boils down to is you're not *actually* doing assignment to the dihedral type when you think you are.  You are calling `DihedralTypeList.__setitem__` under the hood.  That's what the `[]` accessor does.

ChayaSt:
I added a test that passes and an extra warning for ```DihedralType```

Thanks for the explanation. This was interesting and I think I understand it now :).

swails:
Looks good.  Thanks!

I'm about to do a version bump in a PR I'm submitting now, so I'll just pull this in.

PR #673, opened 2016-04-14
Box test

hainm:
test will fail.


hainm:
seems the failure is due to openmm with gb neck error.

swails:
The OpenMM error arose because a new version was just released that has better error checking on GBneck radii.  I'll look into why this is happening and issue a PR to fix it at lunchtime.

In the meantime, I'd like to try and make a concerted effort to avoid adding test files unless it's necessary, particularly binary ones.  You can trigger this error simply by giving a box array with what should be a permissible shape.  Something like this:

```Python
pmd.load_file('prmtop', xyz='inpcrd', box=np.random.rand(10, 6))
```
In this case you are free to choose any existing prmtop you want.  Can you open a new PR that doesn't have these binary files in the history?

hainm:
Okie

hainm:
new PR: https://github.com/ParmEd/ParmEd/pull/675

Issue #472, opened 2015-11-06
git tag to parmed dev version?


hainm:
I just added this stuff to pytraj, my idea is to tell the git commit from version info.

What do you think about adding it to parmed?

https://github.com/Amber-MD/pytraj/blob/master/scripts/base_setup.py#L102-L182

demo:
`pytraj version =  0.1.2.dev12+7721e4c`

swails:
+1, with a small variation.  I don't want it part of the `parmed.version` string.  Either a new attribute `parmed.git_revision` or `parmed.version.git_revision`.  Right now, `parmed.version` is ordered (you don't get that with git revisions), and tuple-based (behaves a bit like Python's `sys.version_info`).

The nice thing about attaching the git revision is that it's immune to accidentally forgetting to bump the version number (although I still have to look into versioneer and see if that can fit the bill).

hainm:
> . I don't want it part of the parmed.version string.

agree. as long as we have something to inspect after installing parmed.

hainm:
oh, unrelated FYI in case you don't know yet: if you move parmed to amber-md, `git clone https://github.com/ParmEd/ParmEd` is still legit.

Work for pytraj

```
$ git clone https://github.com/pytraj/pytraj
Cloning into 'pytraj'...
remote: Counting objects: 42534, done.
Receiving objects:  28% (11910/42534), 18.61 MiB | 18.51 MiB/s
```

PR #681, opened 2016-04-17
Start versioning ParmEd with versioneer

### LINKED ISSUE #677: Switch to using versioneer to track versions
I used just used it in another project I started at work, and it's pretty straightforward to use.  I liked it enough, let's put it in here as well.

swails:
Closes #677 and #472 

This changes the version attribute to the following format:

```
major.minor.patchlevel[+commits_ahead.g<commit hash>[.dirty]]
```

If it's exactly at a major tagged version, then the version is simply something like 2.5.0.  If the tree is ahead of the version string, we get something like 2.5.0+6.g2bdef28, indicating that this code is 6 commits ahead, and the commit hash is 2bdef28 (the g means "g"it).

When printing the version attribute, you get:

```
>>> import parmed as pmd
>>> pmd.__version__
'2.5.0+6.g2bdef28.dirty'
>>> pmd.version
version(major=2, minor=5, patchlevel=0, commits_ahead=6)
```
You can get the full commit ID easily as well:
```
>>> from parmed import _version
>>> _version.get_versions()['full-revisionid']
'2bdef284b44d2f6faf3070fe30ba5d74ec40a317'
```

Note that lack of a revision ID when we're at an actual release version is not at all a problem, since the only way versioneer knows that it's at a release version is when there's an actual tag.  Which means the version is exactly equivalent to a specific commit hash.


swails:
I should note, if `.dirty` is there, it means that there are uncommitted changes in the repo at time of install.

This will prevent me from having to obsessively (and excessively) version bump.

PR #695, opened 2016-04-20
Add atomic number determination to mol2 file parser

### LINKED ISSUE #693: wrong element?
Issue 1
======

What does `element` mean in ParmEd? I got '0' for Carbon

```python
import parmed as pmd

print(pmd.__version__)

parm = pmd.load_file("08Z.mol2", structure=True)
print(parm.atoms[0])
print(parm.atoms[0].element)
```

out
```bash
2.5.0+29.g172d95f
<Atom C1 [0]; In 08Z 0>
0
```

https://github.com/hainm/amber_things/tree/master/things/parmed/element

**updated**
Issue 2
======
should atom.element be something like "carbon" rather a number?

swails:
Fixes #693

### DIFF
--- a/parmed/formats/mol2.py
+++ b/parmed/formats/mol2.py
@@ -8,6 +8,7 @@ import copy
 from parmed.exceptions import Mol2Error, ParameterWarning
 from parmed.formats.registry import FileFormatType
 from parmed.modeller.residue import ResidueTemplate, ResidueTemplateContainer
+from parmed.periodic_table import element_by_name, AtomicNum
 from parmed.residue import AminoAcidResidue, RNAResidue, DNAResidue
 from parmed.structure import Structure
 from parmed.topologyobjects import Atom, Bond
@@ -341,13 +342,23 @@ class Mol2File(object):
                             raise Mol2Error('Residue connection atom %s not '
                                             'found in residue %d' % (a, residx))
             if structure:
+                for atom in struct.atoms:
+                    atom.atomic_number = _guess_atomic_number(atom.name,
+                                                              atom.residue)
                 return struct
             elif len(rescont) > 0:
                 if not multires_structure and mol_info[0]:
                     restemp.name = mol_info[0]
                 rescont.append(restemp)
+                for res in rescont:
+                    for atom in res.atoms:
+                        atom.atomic_number = _guess_atomic_number(atom.name,
+                                                                  res)
                 return rescont
             else:
+                for atom in restemp.atoms:
+                    atom.atomic_number = _guess_atomic_number(atom.name,
+                                                              restemp)
                 return restemp
         except ValueError as e:
             raise Mol2Error('String conversion trouble: %s' % e)
@@ -569,3 +580,15 @@ class Mol2File(object):
                     dest.write('\n')
         finally:
             if own_handle: dest.close()
+
+def _guess_atomic_number(name, residue=None):
+    """ Guesses the atomic number """
+    # Special-case single-atom residues, which are almost always ions
+    name = ''.join(c for c in name if c.isalpha())
+    if residue is None or len(residue.atoms) == 1:
+        if len(name) > 1:
+            try:
+                return AtomicNum[name[0].upper() + name[1].lower()]
+            except KeyError:
+                return AtomicNum[element_by_name(name)]
+    return AtomicNum[element_by_name(name)]
--- a/parmed/formats/mol2.py
+++ b/parmed/formats/mol2.py
@@ -343,8 +343,10 @@ class Mol2File(object):
                                             'found in residue %d' % (a, residx))
             if structure:
                 for atom in struct.atoms:
-                    atom.atomic_number = _guess_atomic_number(atom.name,
-                                                              atom.residue)
+                    anum = _guess_atomic_number(atom.name, restemp)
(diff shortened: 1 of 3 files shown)


swails:
I think #696 further supports that this PR should be merged.

Pinging @davidlmobley again for input (see discussion in #693).

hainm:
Definitely should be merged. If not, I can not visualize mol2 in NGLView by using ParmEd as a backend.

hainm:
How ParmEd and nglview can be used for teaching (derived from David's comment)

https://github.com/mdtraj/mdtraj/issues/597

davidlmobley:
I only have a moment, so I can't review code in detail right now. But
looking at the differences, it's not obvious to me how this guesses. I tend
to worry about what happens for "weird" atom names such as GAFF atom types.
Will it behave correctly there? Or what about Tripos atom naming? Those are
the two obvious cases if you're coming from a "AMBER protein forcefield"
kind of expectation.

Thanks.


On Mon, Apr 25, 2016 at 5:43 PM, Hai Nguyen <notifications@github.com>
wrote:

> Definitely should be merged. If not, I can not visualize mol2 in NGLView
> by using ParmEd as a backend.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/ParmEd/ParmEd/pull/695#issuecomment-214573718>
>



-- 
David Mobley
dmobley@gmail.com
949-385-2436

hainm:
Can we just use information from gaff to guess atomic number?

hainm:
I meant a template.

swails:
> I meant a template.

The mol2 *is* the template.

> But
> looking at the differences, it's not obvious to me how this guesses.

It guesses based on atom names, and failing that (perhaps because the atom name has no letters or no match in the periodic table), atom type names.  It takes connectivity into account as well.  The first letter is used to look up the element unless that atom has no bonds.  If that atom has no bonds, the first two letters are used.  That prevents atoms like NA from being matched to nitrogen and CA to carbon when they are in fact sodium and calcium ions (same with CL).

There will be instances where it fails (like SOD naming from CHARMM being matched as sulfur), but in a file format where elements are not specified, it is very difficult to get around this.  For the common use-cases, the element guessing here will be quite accurate (and I'm always open to proposed improvements).

In many places, atomic number is convenient, but not critically important (e.g., in PDB file writing or mol2 to OFF file conversions), so that such element guessing should have a lot more upside than down.  Of course, for mol2->mol2 processing, it has no effect at all.

davidlmobley:
Sounds good. I agree that there is no way to get around this without
elements being specified, so as long as it handles the "normal" stuff we
deal with in a sensible way then I'm OK with it.


On Mon, Apr 25, 2016 at 6:23 PM, Jason Swails <notifications@github.com>
wrote:

> I meant a template.
>
> The mol2 *is* the template.
>
> But
> looking at the differences, it's not obvious to me how this guesses.
>
> It guesses based on atom names, and failing that (perhaps because the atom
> name has no letters or no match in the periodic table), atom type names. It
> takes connectivity into account as well. The first letter is used to look
> up the element unless that atom has no bonds. If that atom has no bonds,
> the first two letters are used. That prevents atoms like NA from being
> matched to nitrogen and CA to carbon when they are in fact sodium and
> calcium ions (same with CL).
>
> There will be instances where it fails (like SOD naming from CHARMM being
> matched as sulfur), but in a file format where elements are not specified,
> it is very difficult to get around this. For the common use-cases, the
> element guessing here will be quite accurate (and I'm always open to
> proposed improvements).
>
> In many places, atomic number is convenient, but not critically important
> (e.g., in PDB file writing or mol2 to OFF file conversions), so that such
> element guessing should have a lot more upside than down. Of course, for
> mol2->mol2 processing, it has no effect at all.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/ParmEd/ParmEd/pull/695#issuecomment-214579596>
>



-- 
David Mobley
dmobley@gmail.com
949-385-2436

hainm:
we can train "ParmEd" to learn by using ten of thousands of ligands with known atomic numbers.

davidlmobley:
> we can train "ParmEd" to learn by using ten of thousands of ligands with
known atomic numbers

I have no idea what you mean. Machine learning is all well and good, but
you can't very well teach a machine to understand several different
languages simultaneously without a mechanism to establish what language you
are learning. That is, you can think of each different atom naming scheme
as a different language. The user (hopefully) knows what atom naming scheme
(language) they are using, but the file doesn't specific this and there is,
as far as I know, no way to infer it uniquely from the typically short atom
names which will be used. It could be one ParmEd understands, or it could
be something else of the user's construction.

Ultimately the field will need to move to file formats which actually
specify elements/atomic numbers.

On Mon, Apr 25, 2016 at 8:34 PM, Hai Nguyen <notifications@github.com>
wrote:

> we can train "ParmEd" to learn by using ten of thousands of ligands with
> known atomic numbers.
>
> —
> You are receiving this because you were mentioned.
> <https://github.com/ParmEd/ParmEd/pull/695#issuecomment-214596984>
>



-- 
David Mobley
dmobley@gmail.com
949-385-2436

hainm:
> Ultimately the field will need to move to file formats which actually
specify elements/atomic numbers.

yeah, agree.

PR #699, opened 2016-04-28
Add a compress_whitespace argument to the mol2 file writer

### LINKED ISSUE #698: Temporary optional adjustment to .mol2 writer?
Is there any chance you'd temporarily be willing to accept an adjustment to the `.mol2` writer which would manipulate the amount of whitespace used to delimit fields in the ATOMS section of `.mol2` files? We could submit a PR.

We've encountered an issue where the OpenEye tools can't process partial charges in `.mol2` files output by ParmEd because ParmEd increases the amount of whitespace separating fields. Apparently the OpenEye tools only process the first 80 characters on each line (even though there is no such limit in the file specifications that I can tell), so partial charges get truncated when roundtripping a `.mol2` file. 

I've submitted a bug report to OpenEye, but the timescale for releases is not super fast and our development will need to proceed anyway. The simplest work around would be to add an optional argument to the `.mol2` writer which would try to compress whitespace to hit an 80 character limit when possible (so the normal behavior would be as now and someone would only get this if they ask for it). Alternatively, we'll write some other tool which will postprocess ParmEd output `.mol2` files to reduce the whitespace. :( 

@camizanette , this is your issue.

Probably also relevant to @jchodera , given the extensive use of `.mol2` files and OpenEye tools in his group.

swails:
Closes #698

### DIFF
--- a/parmed/formats/mol2.py
+++ b/parmed/formats/mol2.py
@@ -374,7 +374,7 @@ class Mol2File(object):
     #===================================================
 
     @staticmethod
-    def write(struct, dest, mol3=False, split=False):
+    def write(struct, dest, mol3=False, split=False, compress_whitespace=False):
         """ Writes a mol2 file from a structure or residue template
 
         Parameters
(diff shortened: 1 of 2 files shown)


davidlmobley:
@swails - would you mind either incrementing the version number or letting me know what the easiest way to make sure I'm running at least this version of ParmEd? I need to (a) update my conda recipe to ask for at least this version, and (b) add a test to my code to make sure we're running at least this version.

swails:
versioneer automatically updates versions, so this should have a unique version associated with it already.

To assert this PR or higher, do this:

```python
import parmed as pmd

assert pmd.version > (2, 5, 1, 10)
```

This will always work (because the next version increment I do will be 2.5.2 or 2.6).  You can use the command:

```
parmed --version
```

to get the current version of ParmEd.  The output will look something like this:

```
$ parmed --version
parmed: Version 2.5.1+10.gf90f7a0
```

Where 2.5.1 is the closest tag, +10 means 10 commits after that, and .<hash> is the commit <hash> in git.  So the version tuple is major, minor, patch, commit.

davidlmobley:
Thanks!

PR #718, opened 2016-06-04
Write the XPLOR flag to CHARMM PSF files if it is warranted

### LINKED ISSUE #715: Issues in running CHARMM on ParmEd output
I'm having problems running CHARMM on my outputs from Parmet. The PSF files were generated by the command structure.save(charmm_output_psf, format='psf',overwrite=True), where charmm_output_psf is the filename saved to. I didn't see other options to use.

are generated with lines like:

         1 SYS      1        CLP      C1       C3LTU   -0.003799       12.0100

However, these seem to be throwing errors of the type:

    At line 246 of file /home/mish4610/software/charmm/source/io/psfres.src (unit = 90, file = 'CBC-G1/CBC-G1_converted.psf')
    Fortran runtime error: Bad value during integer read

Where that specific line 246 is:
              read(u,fmt01a,end=6) &
                   ii,lsegid,lresid,lres,atype(i),iac(i), &
                   cg(i),amass(i),imove(i),ech(i),eha(i)

Which seems to expect three more integers after the atomic mass. I thought about finding another read mode, but I can't seem to find any of them that expect amass(i) as the last argument.  Looking at PSFFile, there doesn't seem to be other ways to write the psffile atom records, that don't just end in the mass.

In contrast, I found a psf file with atom entries of the type 

         1 A        1        OAH      C1       CG2R61    0.00000       12.0110           0   0.00000     -0.301140E-02

Seemed to be read fine, with energies generated correctly. 

Any hints or ideas for getting energies out?  Perhaps there's some logic to the code I'm missing, and there's a flag that things would run correctly with?

Thanks!

swails:
Fixes #715

### DIFF
--- a/parmed/formats/psf.py
+++ b/parmed/formats/psf.py
@@ -75,7 +75,7 @@ class PSFFile(object):
             attribute, it will be used to print the PSF file. Otherwise, it will
             be treated like a string and a file will be opened, printed, then
             closed
-        vmd : bool 
+        vmd : bool
             If True, it will write out a PSF in the format that VMD prints it in
             (i.e., no NUMLP/NUMLPH or MOLNT sections)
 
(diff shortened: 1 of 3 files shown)


mrshirts:
I've downloaded and verified that it produced the XPLOR line when given this setup, and that the file is read correctly afterwards.

mrshirts:
So, does the omnia release not have these fixes as all? I tried pulling down from that, but these aren't here.  Just wondering . . .

PR #726, opened 2016-06-12
Set ifbox to 3 for general triclinic

### LINKED ISSUE #724: GROMACS->Amber for dodecahedron
I'm trying to convert a GROMACS topology with rhombic dodecahedron to Amber. The generated amber topology file has ifbox=2 which is incorrect. And Amber complains that the box doesn't have the 109 angle for truncated octahedral (which it should for ifbox=2).

swails:
I "knew" this is what 3 was for, but didn't actually realize 2 didn't work for the general triclinic case.  This adds that to fix #724


swails:
This is getting replaced by #844

Issue #727, opened 2016-06-12
Converted name into CHARMM format apparently can't be read by CHARMM


mrshirts:
One of the atom types created on the production if CHARMM files from an AMBER structure apparently has an illegal name.  The NA+ name in AMBER apparently leads to a CHARMM atom type NA+LTU (appearing in the .psf, .rtf, and .prm files), which is not recognized by CHARMM.  

I get the error: 

   PARRDR> WARNING: ATOM FOR NBOND NA+L DOESNT EXIST
   ****\*  WARNING  ****\* PARRDR no nonbond parameters for atom type:  15  NA+L
   NO nonbond interactions will be computed for this atom type.

Which I missed before, since the simulations seemed to run fine.

I have verified that if I simply change NA+LTU to NALTU (or to NALTUU, so the length remains constant), the it reads it just fine, so I think it may just be that '+' is an illegal character. 

Jason, this error should appear in the files I sent you earlier.

The only thing different is that I am now using the following input file (yes, the .crd file is actually called .par in the set I sent you -- I've now corrected it in the code for future runs, but left it the same here to be consistent).

   ! CHARMM Energy for CBC-G1.inp
   open read card unit 10 name "CBC-G1.rtf"
   read rtf card unit 10
   open read card unit 20 name "CBC-G1.prm"
   read para card unit 20
   read psf card name "CBC-G1.psf"
   read coor card name "CBC-G1.par"
   read coor card comp name "CBC-G1.par"
   crystal define cubic 40.25360320 40.25360320 40.25360320 90.00000000 90.00000000 90.00000000
   crystal build noper 0
   nbond inbfrq -1 imgfrq -1  -
   elec ewald pmew fftx 48 ffty 48 fftz 48 kappa 0.20822755 order 4 -
   vdw vatom lrc cutnb 15. cutim 15. ctofnb 15. ctonnb 15.
   energy
  stop

swails:
#728 claims that this was fixed by that PR.  If I recall, I think it was.  If I'm mistaken, feel free to reopen.

PR #728, opened 2016-06-12
Make sure that + does not appear in any charmm atom type names

swails:
+ seems to be an illegal character.

### DIFF
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -37,10 +37,11 @@ def _typeconv(name):
     if isinstance(name, integer_types):
         return name
     if name.upper() == name:
-        return name.replace('*', 'STR')
+        return name.replace('*', 'STR').replace('+', 'PL').replace('-', 'MI')[:8]
     # Lowercase letters present -- decorate the type name with LTU --
     # Lower To Upper
-    return ('%sLTU' % name.upper()).replace('*', 'STR')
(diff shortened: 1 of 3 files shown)


swails:
Fixes #727

mrshirts:
When testing this, I got the error:

 PARRDR> WARNING: Atom:   15 "NAPL" does not match RTF name "NAPL"
 PARMIO> NONBOND, HBOND lists and IMAGE atoms cleared.
  
 CHARMM>    read psf card name "CBC-G1/CBC-G1_converted.psf"
 VOPEN> Attempting to open::CBC-G1/CBC-G1_converted.psf::
 MAINIO> Protein structure file being read from unit  90.
 PSFRD2> Reading PSF in the expanded format.
 TITLE>  
 TITLE>  *
Atom first        163         0 NAPLLT problem in psf

I THINK it might be that the name is now too long.  The formatting is off (columns not lined up with the new name) but  if I shorten the name from NAPLLTU NAPLTU, while keeping the bad formatting, it still runs.

If I replace mentions in the new code of 'PL' and 'MI' with 'P' and 'M', it seems to work.  However, I don't know if that will work in general.

swails:
@mrshirts -- I went ahead and implemented the +->P and -->M (instead of PL and MI).  There is always the possibility of name collisions, but honestly I think that is *highly* unlikely.

So since you said that worked, I think this can be merged.

PR #730, opened 2016-06-14
Escape strings in gromacs .top comment

### LINKED ISSUE #729: Issue with the way parmed interprets the command line
ParmEd, when printing .top files (perhaps other files) prints the command line that the ParmEd file was called with in a comment. The converter I am using that calls Parmed include some options with strings that include newlines (since I'm passing in nonbonded parameters lines for files.  However, when printing to the .top file, it prints the newline literally (rather than \n), resulting in an uncommented line in the .top file that can't be interpreted. 

Any thoughts on a good way to resolve this?

ctk3b:
Fixes #729 

Might require some py2/3 compatibility tweaks


swails:
Thanks!

PR #731, opened 2016-06-15
Call finalize on gbsa forces

jlmaccal:
The amber custom gbsa forces now require calling `finalize` before use. This change adds this call, swallowing the resulting exception that occurs with older versions of openmm.

The call to finalize is needed because the optimized versions of `GBn` and `GBn2` cache the atom parameters in order to upload only the necessary values of `d0` and `m0`. The particles are only added to the force when `finalize` is called.


swails:
Good call, thanks!

Issue #733, opened 2016-08-15
add alternate atom label when writing to pdb file from LES parm7 + rst7?


hainm:
currently, with a given LES parm7 and rst7, I got 

```bash
...
ATOM     16  CE  LYS     1      -0.975   8.165  10.101  0.00  0.00           C  
ATOM     17  CE  LYS     1      -0.963   8.369  10.347  0.00  0.00           C  
ATOM     18  HE2 LYS     1      -0.780   9.159  10.504  0.00  0.00           H  
ATOM     19  HE2 LYS     1      -2.034   8.549  10.446  0.00  0.00           H  
ATOM     20  HE3 LYS     1      -0.339   8.047   9.224  0.00  0.00           H  
ATOM     21  HE3 LYS     1      -0.454   9.229  10.782  0.00  0.00           H  
ATOM     22  NZ  LYS     1      -2.393   8.077   9.689  0.00  0.00           N  
ATOM     23  NZ  LYS     1      -0.618   8.282   8.901  0.00  0.00           N  
```

do you want to add code to add 'A, B, ...' labels for alternative atoms?

```bash
ATOM     16  CE ALYS     1      -0.975   8.165  10.101  0.00  0.00           C
ATOM     17  CE BLYS     1      -0.963   8.369  10.347  0.00  0.00           C
ATOM     18  HE2ALYS     1      -0.780   9.159  10.504  0.00  0.00           H
ATOM     19  HE2BLYS     1      -2.034   8.549  10.446  0.00  0.00           H
ATOM     20  HE3ALYS     1      -0.339   8.047   9.224  0.00  0.00           H
ATOM     21  HE3BLYS     1      -0.454   9.229  10.782  0.00  0.00           H
ATOM     22  NZ ALYS     1      -2.393   8.077   9.689  0.00  0.00           N
ATOM     23  NZ BLYS     1      -0.618   8.282   8.901  0.00  0.00           N
```
related to cpptraj issue: https://github.com/Amber-MD/cpptraj/issues/388

hainm:
I already have a working code, just want to ask if you would like to have it in ParmEd.

If yes, what's your favorite syntax?

```python
parm.write_pdb('new.pdb', label_alternates=True)
```
?

swails:
I'd rather it just *write* the alternates.  How did you do this, by the way?  Because each `Atom` has an `other_atoms` dict that maps the label (e.g., 'B') with an `Atom` instance of its alternate location.

[...]

PR #735, opened 2016-08-15
Write alternates label for LES parm

hainm:
#733


parmed/amber/_amberparm.py
```diff
@@ -1996,6 +1998,31 @@ def _add_missing_13_14(self, ignore_inconsistent_vdw=False):
...
 
+    def _get_atom_dict_for_alternate_labels(self, residues):
+        from collections import defaultdict
+
+        atom_dict = defaultdict(list)
+        for resid, residue in enumerate(residues):
+            atom_dict[resid] = defaultdict(list)
+            for atom in residue.atoms:
+                atom_dict[resid][atom.name].append(atom)
+        return atom_dict
```
swails: Move the import to the top of the file.  Also, `atom_dict` looks to me like it should not be a dict at all, but just a simple list:

```Python
atom_collection = [defaultdict(list) for r in residues]
for ad, res in zip(atom_collection, residues):
    for atom in res:
        ad[atom.name].append(atom)
return atom_collection
```

Also, you can use `self.residues` rather than passing in a `residues` object.

parmed/amber/_amberparm.py
```diff
@@ -1996,6 +1998,31 @@ def _add_missing_13_14(self, ignore_inconsistent_vdw=False):
...
+
+    def _label_alternates(self, residues=None):
+        # if we know which residue has alternative atoms, we do not need
+        # to iterate all residues
+        if residues is None:
+            residues = self.residues
```
swails: What is the use-case for `residues` *not* being `self.residues`?  If there is no use-case, get rid of the option to pass in `residues`.

parmed/amber/_amberparm.py
```diff
@@ -1996,6 +1998,31 @@ def _add_missing_13_14(self, ignore_inconsistent_vdw=False):
...
+        atom_with_residue_dict = self._get_atom_dict_for_alternate_labels(residues=residues)
+        possible_labels = list('ABCDEF')
+        for _, residue_dict in atom_with_residue_dict.items():
```
swails: Import `iteritems` from `six` and use that instead (much better in Python 2)

parmed/amber/_amberparm.py
```diff
@@ -1996,6 +1998,31 @@ def _add_missing_13_14(self, ignore_inconsistent_vdw=False):
...
+            for atom_name, atom_list in residue_dict.items():
```
swails: Same as above

swails:
Any idea why the tests are failing?  Looks like some old ParmEd code is sneaking in there somehow.  Maybe from ambermini?

swails:
Doesn't look like ambermini...

hainm:
Per the failure, I think it's from old ParmEd code. For example, one of the failure is due to rdkit.from_smiles does not have `coordinates=` keyword.

Per other comments, I will finish them tmr.

parmed/amber/_amberparm.py
hainm: > Move the import to the top of the file.

just deleted that line since the import is already at the top.

> Also, atom_dict looks to me like it should not be a dict at all, but just a simple list:

ah, I took my code from phenix-amber code. In that case, only a subset of residues having alternative atoms are used. I just updated to your suggestion.

> Also, you can use self.residues rather than passing in a residues object.

updated.

parmed/amber/_amberparm.py
hainm: > What is the use-case for residues not being self.residues? If there is no use-case, get rid of the option to pass in residues.

it's has use-case in another context where you know specific residues that having alternative atoms.

(saying you have 1000 residues but only few of them has alternative atoms. So you don't need to iterate all 1000 residues). But I updated the code in ParmEd context.

parmed/amber/_amberparm.py
hainm: done

parmed/amber/_amberparm.py
hainm: done

parmed/amber/_amberparm.py
```diff
@@ -1996,6 +1998,31 @@ def _add_missing_13_14(self, ignore_inconsistent_vdw=False):
...
+                if len(atom_list) > 1:
+                    for atom, label in zip(atom_list, possible_labels):
```
hainm: I have never met any case there is 6 alternative positions for an atom.

But just updated the code.

hainm:
uhm I think the failure is due to  conda.

Originally `parmed` will be installed from either `omnia` (via `openmm`) or `ambermd` (via `pysander`) channel. And then `python setup.py install` will overwrite the old install.

Not sure what happens to conda recently.

hainm:
good for review (again) now.

swails:
Ah!  `pysander` is bringing in ParmEd as a dependency.  Thanks for hunting this down!

parmed/amber/_amberparm.py
```diff
@@ -267,6 +267,7 @@ def from_rawdata(cls, rawdata):
             inst.box = _copy.copy(rawdata.box)
         inst.hasbox = inst.box is not None
         inst.hasvels = inst.velocities is not None
+        inst._label_alternates()
```
swails: Should we do this *just* for LES topologies?

parmed/amber/_amberparm.py
```diff
@@ -377,6 +378,7 @@ def from_structure(cls, struct, copy=False):
                 dt.per = 1.0
         inst.remake_parm()
         inst._set_nonbonded_tables(nbfixes)
+        inst._label_alternates()
```
swails: Same as above.

swails:
The only minor concern I have is that `inst._label_alternatives()` may be somewhat costly for large systems.  Should we do it only if the LES pointer is non-zero?  (I think that's the ``NCOPY`` pointer).

After we answer that, this is ready to go.  Thanks!

hainm:
> Should we do it only if the LES pointer is non-zero?

indeed we should. For non-LES case, we can explicitly call `_label_alternativecs` if we know we need to do so.

parmed/amber/_amberparm.py
```diff
@@ -379,7 +380,8 @@ def from_structure(cls, struct, copy=False):
                 dt.per = 1.0
         inst.remake_parm()
         inst._set_nonbonded_tables(nbfixes)
-        if inst.pointers['NCOPY'] >= 2:
+        n_copy = inst.pointers.get('NCOPY', 1)
+        if n_copy >= 2:
```
swails: It could have been done
```python
if inst.pointers.get('NCOPY', 1):
    inst._label_alternatives()
```

but this way gives me higher coverage :).

swails:
Looks good.

parmed/amber/_amberparm.py
hainm: if inst.pointers.get('NCOPY', 1):
Won't work since it's alway True. :p
Do you mean?
if inst.pointers.get('NCOPY', 1) >= 2:
Hai

swails: haha, yes.  Thanks :smiley:

I already merged this, though.  As I said, it looks good (and the coverage appears higher the way you have it)

PR #736, opened 2016-08-15
add symmetry for pdb

hainm:
This is my proposed code, just a draft version without cleaning up

```python
In [5]: parm = pmd.load_file('./4lzt.pdb')

In [6]: parm.symmetry.data
Out[6]: 
array([[ 1.,  0.,  0.,  0.],
       [ 0.,  1.,  0.,  0.],
       [ 0.,  0.,  1.,  0.]])

In [7]: print(parm.symmetry.write_string())
REMARK 290   SMTRY1   1  1.000000  0.000000  0.000000        0.00000
REMARK 290   SMTRY2   1  0.000000  1.000000  0.000000        0.00000
REMARK 290   SMTRY3   1  0.000000  0.000000  1.000000        0.00000
```

Please give review about the API keywords. thanks.


parmed/symmetry.py
```diff
@@ -0,0 +1,57 @@
+import numpy as np
+
+class Symmetry(object):
+    def __init__(self, lines=None):
```
swails: Converting PDB lines into a `Symmetry` object should be done as a method in the PDB parsing class.  Currently, making a `Symmetry` object from anything other than a set of PDB lines following a REMARK 290 is awkward.  `Symmetry` should have a format-agnostic API.

It should either take the full tensor as an array (and store it as a numpy array) or something else that feels natural to people that regularly work with symmetry operators.

You should not be able to instantiate an "empty" `Symmetry` object -- if a format doesn't define it, then `structure.symmetry` is left as `None` and this constructor will never be called.

swails: So the way I envision this API looking is something like this:

```python

class Symmetry(object):
    def __init__(self, tensor):
        self.data = np.array(tensor, copy=False, subok=True)
        if self.data.shape != whatever_shape_is_expected:
            raise ValueError('Tensor shape is unexpected!')
```

parmed/symmetry.py
```diff
@@ -0,0 +1,57 @@
...
+        for index, arr in enumerate(self.data):
+            arr_list = [1 + index % 3, 1 + int(index/3)] + arr.tolist()
+            line = "REMARK 290   SMTRY" + fmt % tuple(arr_list)
+            lines.append(line)
+
+        output = '\n'.join(lines)
+        return output
+
+if __name__ == '__main__':
+    lines = """
```
swails: This should become a test case instead.

parmed/symmetry.py
```diff
@@ -0,0 +1,57 @@
...
+
+    def parse(self, lines):
+        data = []
+        for line in lines:
+            if line.startswith('REMARK 290   SMTRY'):
+                data.append(line.split()[4:])
+
+        return np.asarray(data, dtype='f8')
+
+    def write_string(self):
```
swails: This should also be moved as a method on the PDB parser/writer.

swails:
I made some suggestions.  Looks good on the whole -- just some rearranging to give a more consistent API.

Also, would be great to have Symmetry reading (and writing!) support for PDBx/mmCIF files also! :smiley:

hainm:
I just updated code that follows your comments.

> Also, would be great to have Symmetry reading (and writing!) support for PDBx/mmCIF files also! :smiley:

yeah, I will do it after playing more with this format.

parmed/formats/pdb.py
```diff
@@ -312,8 +313,12 @@ def parse(filename, skip_bonds=False):
         ZEROSET = set('0')
         altloc_ids = set()
```
swails: delete this line (too much whitespace)

parmed/formats/pdb.py
```diff
@@ -746,6 +760,12 @@ def write(struct, dest, renumber=True, coordinates=None, altlocs='all',
             dest.write('CRYST1%9.3f%9.3f%9.3f%7.2f%7.2f%7.2f %-11s%4s\n' % (
                     struct.box[0], struct.box[1], struct.box[2], struct.box[3],
                     struct.box[4], struct.box[5], struct.space_group, ''))
+        if struct.symmetry is not None:
+            fmt = '%d%4d%10.6f%10.6f%10.6f%15.5f\n'
+            for index, arr in enumerate(struct.symmetry.data):
+                arr_list = [1 + index % 3, 1 + int(index/3)] + arr.tolist()
```
swails: Use integer division (`//`)

test/test_parmed_formats.py
```diff
@@ -962,6 +964,45 @@ def test_pdb_write_format(self):
...
+            for line in remark_290_lines.split():
+                self.assertTrue(line.strip() in buffer)
+
+        # 4lzt
+        pdbfile = get_fn('4lzt.pdb')
+        parm = pmd.load_file(pdbfile)
+        remark_290_lines = """
+        REMARK 290   SMTRY1   1  1.000000  0.000000  0.000000        0.00000            
+        REMARK 290   SMTRY2   1  0.000000  1.000000  0.000000        0.00000            
+        REMARK 290   SMTRY3   1  0.000000  0.000000  1.000000        0.00000
```
swails: The triple-quotes here mean that the leading whitespace caused by the indentation is part of the lines themselves.  These spaces should be removed (so just remove the indentation from the strings themselves)

test/test_parmed_formats.py
```diff
@@ -962,6 +964,45 @@ def test_pdb_write_format(self):
...
+        REMARK 290   SMTRY1   2 -1.000000  0.000000  0.000000       17.52500            
+        REMARK 290   SMTRY2   2  0.000000 -1.000000  0.000000        0.00000            
+        REMARK 290   SMTRY3   2  0.000000  0.000000  1.000000       21.18500            
+        REMARK 290   SMTRY1   3 -1.000000  0.000000  0.000000        0.00000            
+        REMARK 290   SMTRY2   3  0.000000  1.000000  0.000000       20.25000            
+        REMARK 290   SMTRY3   3  0.000000  0.000000 -1.000000       21.18500            
+        REMARK 290   SMTRY1   4  1.000000  0.000000  0.000000       17.52500            
+        REMARK 290   SMTRY2   4  0.000000 -1.000000  0.000000       20.25000            
+        REMARK 290   SMTRY3   4  0.000000  0.000000 -1.000000        0.00000    
+        """
```
swails: Same comment as above.

swails:
Can you merge master into this branch to get the tests to pass here?

hainm:
good for review now.

swails:
Looks good.  Thanks!

PR #745, opened 2016-08-24
Change overload to override

rafwiewiora:
New PR here, should be without conflicts now.


swails:
Failed test was just one that for some reason didn't run.  Looks fine to me.

Issue #334, opened 2015-08-07
[todolist?] sdf format


hainm:
```python
In [1]: pmd.load_file('https://raw.githubusercontent.com/hainm/open-forcefield-group/master/ideas/bayesian-atomtype-sampling/datasets/zinc.sdf')
---------------------------------------------------------------------------
FormatNotFound                            Traceback (most recent call last)
<ipython-input-1-3d7fc394bb95> in <module>()
----> 1 pmd.load_file('https://raw.githubusercontent.com/hainm/open-forcefield-group/master/ideas/bayesian-atomtype-sampling/datasets/zinc.sdf')

/home/haichit/anaconda3/lib/python3.4/site-packages/parmed/formats/registry.py in load_file(filename, *args, **kwargs)
    128     else:
    129         # We found no file format
--> 130         raise FormatNotFound('Could not identify file format')
    131 
    132     # We found a file format that is compatible. Parse it!

FormatNotFound: Could not identify file format
```

swails:
I'm not against it, but I'd have to learn the format first.  If you want to add a parser, I'd certainly be happy to accept it ;).

hainm:
Both you and I do not need this, so close.

PR #748, opened 2016-08-26
parse sdf format via rdkit

hainm:
still  better than nothing. added very minimal test.

related to https://github.com/ParmEd/ParmEd/issues/334

### DIFF
--- a/parmed/formats/registry.py
+++ b/parmed/formats/registry.py
@@ -102,6 +102,8 @@ def load_file(filename, *args, **kwargs):
         - ``.gz`` : gzip compressed file
         - ``.bz2`` : bzip2 compressed file
 
+    SDF file is loaded via `rdkit` package. 
+
     Examples
     --------
 
@@ -136,6 +138,11 @@ def load_file(filename, *args, **kwargs):
 
     >>> load_file('1aki.ff99sbildn.top')
     <GromacsTopologyFile 40560 atoms [9650 EPs]; 9779 residues; 30934 bonds; parametrized>
+
+    Load a SDF file -- only works with rdkit installed
+
+    >>> load_file('mol.sdf', structure=True)
+    <Structure 34 atoms; 1 residues; 66 bonds; NOT parametrized>
 
     Raises
     ------
--- a/parmed/formats/sdf.py
+++ b/parmed/formats/sdf.py
@@ -27,12 +27,8 @@ class SDFFile(object):
         is_fmt : bool
             True if it is a sdf file, False otherwise
         """
-        try:
-            words = linecache.getline(filename, lineno=4).split()
-            return len(words) >= 3 and words[-1] in ('V2000', 'V3000')
-        except IndexError:
-            # 4-th line is empty
-            return False
+        words = linecache.getline(filename, lineno=4).split()
+        return len(words) >= 3 and words[-1] in ('V2000', 'V3000')
 
     @staticmethod
     def parse(filename, structure=False):
--- a/parmed/formats/sdf.py
+++ b/parmed/formats/sdf.py
@@ -1,6 +1,6 @@
 """
 This module contains parsers for sdf file format 
-extension described at [TODO] 
+extension described at https://www.cas.org/content/chemical-suppliers/example-sdf
 """
 from __future__ import print_function, division, absolute_import
 
--- a/parmed/formats/sdf.py
+++ b/parmed/formats/sdf.py
@@ -31,7 +31,7 @@ class SDFFile(object):
         try:
             for line in f:
                 if not line.strip(): continue
-                if line.startswith('M  END') or  line.startswith('$$$$'):
+                if line.startswith('M  END') or line.startswith('$$$$'):
                     return True
                 else:
                     continue
--- a/parmed/formats/sdf.py
+++ b/parmed/formats/sdf.py
@@ -27,10 +27,11 @@ class SDFFile(object):
         is_fmt : bool
             True if it is a sdf file, False otherwise
         """
-        line = linecache.getline(filename, lineno=4)
         try:
-            return line.split()[-1] in ['V2000', 'V3000']
+            words = linecache.getline(filename, lineno=4).split()
+            return len(words) >= 3 and words[-1] in ('V2000', 'V3000')
         except IndexError:
+            # 4-th line is empty
             return False
 
     @staticmethod
--- a/parmed/formats/sdf.py
+++ b/parmed/formats/sdf.py
@@ -1,0 +1,44 @@
+"""
+This module contains parsers for sdf file format 
+extension described at [TODO] 
+"""
+from __future__ import print_function, division, absolute_import
+
+from parmed.formats.registry import FileFormatType
+from parmed.utils.io import genopen
+from parmed.utils.six import add_metaclass
+from parmed import rdkit
+
+@add_metaclass(FileFormatType)
+class SDFFile(object):
+    """ Class to read SDF file """
+
+    @staticmethod
+    def id_format(filename):
+        """ Identify the file as a SDF file format or not
+
+        Parameters
+        ----------
+        filename : str
+            Name of the file to test whether or not it is a sdf file
+
+        Returns
+        -------
+        is_fmt : bool
+            True if it is a sdf file, False otherwise
+        """
+        f = genopen(filename, 'r')
+        try:
+            for line in f:
+                if not line.strip(): continue
+                if line.startswith('M  END') or  line.startswith('$$$$'):
+                    return True
+                else:
+                    continue
+            return False
+        finally:
+            f.close()
+
+    @staticmethod
+    def parse(filename, structure=False):
+        return rdkit.from_sdf(filename, structure=structure)
--- a/parmed/formats/sdf.py
+++ b/parmed/formats/sdf.py
@@ -3,6 +3,7 @@ This module contains parsers for sdf file format
 extension described at https://www.cas.org/content/chemical-suppliers/example-sdf
 """
 from __future__ import print_function, division, absolute_import
+import linecache
 
 from parmed.formats.registry import FileFormatType
 from parmed.utils.io import genopen
@@ -27,17 +28,11 @@ class SDFFile(object):
         is_fmt : bool
             True if it is a sdf file, False otherwise
         """
-        f = genopen(filename, 'r')
+        line = linecache.getline(filename, lineno=4)
         try:
-            for line in f:
-                if not line.strip(): continue
-                if line.startswith('M  END') or line.startswith('$$$$'):
-                    return True
-                else:
-                    continue
+            return line.split()[-1] in ['V2000', 'V3000']
+        except IndexError:
             return False
-        finally:
-            f.close()
 
     @staticmethod
     def parse(filename, structure=False):
--- a/parmed/formats/sdf.py
+++ b/parmed/formats/sdf.py
@@ -6,7 +6,6 @@ from __future__ import print_function, division, absolute_import
 import linecache
 
 from parmed.formats.registry import FileFormatType
-from parmed.utils.io import genopen
 from parmed.utils.six import add_metaclass
 from parmed import rdkit
 
--- a/parmed/formats/__init__.py
+++ b/parmed/formats/__init__.py
@@ -3,13 +3,14 @@ A package dealing with different file formats and automatic detection of those
 formats
 """
 
-__all__ = ['load_file', 'PDBFile', 'CIFFile', 'Mol2File', 'PSFFile', 'PQRFile']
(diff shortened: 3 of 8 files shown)


swails:
This is awesome!

One comment here -- the fact that RDKit is used to do the parsing is an implementation detail.  This parsing should be in an `SDFFile` class inside the `parmed/formats` package.  It should also be assigned the metaclass `FileType` (and implementing an `format_id` function and putting the parsing in a `parse` staticmethod will automatically implement SDF file autodetection -- but this can always be done later).

SMILES processing may benefit from being moved to the `formats` package, but that can wait.

hainm:
The `pmd.load_file` is independent from other packages and I don't know if it's ok to raise ImportError 
if users loading `.sdf` format. Using `pmd.rdkit.from_sdf` explicitly says that users should aware of `rdkit`.  What do you think? Otherwise, we can have detail doc about using `rdkit` as reader backend for some formats.

Minor point: It seems that a single `sdf` file a many molecules, how we deal with that if make a `SDFFile` parser class?

swails:
> The pmd.load_file is independent from other packages and I don't know if it's ok to raise `ImportError`

Yes, it's fine.  I'd rather have `pmd.load_file` work when you have rdkit installed on an SDF file than never work.  The import error will also tell you what you need to do to get ParmEd to parse that file.

If your script depends on the SDF file being parsed, rdkit will be needed regardless of whether or not `pmd.load_format` automatically recognizes the format.

> if users loading .sdf format. Using pmd.rdkit.from_sdf explicitly says that users should aware of rdkit. What do you think? Otherwise, we can have detail doc about using rdkit as reader backend for some formats.

As I said before, that's just an implementation detail :).  My proposal leaves open the avenue of replacing the SDF parser with a native one in the future if we choose to without changing the API.

Furthermore, the `ImportError` that says "couldn't import RDKit" also tells users explicitly what they need to do :).  You can also catch the import error and print out an even more detailed error message if you wanted to.

Basically, looking forward the correct place to put this is `parmed.formats`.

hainm:
Okie, I can add SDFFile next week.

Do you have comment about my "minor point" above?

swails:
> Minor point: It seems that a single sdf file a many molecules, how we deal with that if make a SDFFile parser class?

Same way we deal with Mol2 files, which are in the same boat.  If it defines one molecule, return a `Structure`.  Otherwise, return a list of `Structure`.

hainm:
Thanks.

hainm:
added `SDFFile` class. Let's see if travis is green.

parmed/formats/registry.py
```diff
@@ -137,6 +139,11 @@ def load_file(filename, *args, **kwargs):
     >>> load_file('1aki.ff99sbildn.top')
     <GromacsTopologyFile 40560 atoms [9650 EPs]; 9779 residues; 30934 bonds; parametrized>
 
+    Load a SDF file -- only works with rdkit installed
+
+    >>> load_file('mol.sdf', structure=True)
+    <Structure 34 atoms; 1 residues; 66 bonds; NOT parametrized>
```
hainm: uhm, the bonds are wrongly detected. the z-coordinates of SDF file seems to 0.

parmed/formats/sdf.py
```diff
@@ -0,0 +1,44 @@
...
+            True if it is a sdf file, False otherwise
+        """
+        f = genopen(filename, 'r')
+        try:
+            for line in f:
+                if not line.strip(): continue
+                if line.startswith('M  END') or line.startswith('$$$$'):
+                    return True
+                else:
+                    continue
```
swails: This is going to parse the entire file if it's not an SDF file, which could substantially slow down file format detections.  Is there any way we can limit this to just looking at a few lines?

hainm: > Is there any way we can limit this to just looking at a few lines?

I checked this link https://www.cas.org/content/chemical-suppliers/example-sdf and seems that 'No'. But since this format is for small molecule, we can just check a part of it, may be first 100-200 lines?

hainm: but what does parmed format detection work?

does it go over all the reader class and the loop stops when 'id_format=True'? 
If yes, we can always check sdf in the end of the loop.

swails: The `FileFormatType` metaclass checks the attributes of a class after it's been constructed and checks to see if there is a `id_format` method implemented on that class.  If there is, it adds the newly constructed class to a parser registry.

The `parmed.load_file` then iterates through the registry and passes the file to the `id_format` method until one of the classes returns `True` (indicating that the file format has been identified).  If that class has a `parse` method, it then passes the filename to that `parse` method.  Otherwise, it passes the filename to the main constructor.

Since the registry is a dict (and you can never be sure what order the classes are constructed in the first place), there is no well-defined order for which class formats are identified first.  So you have to assume that any class's `id_format` method will be called for any kind of file.  That's why it's important that `id_format` be able to tell, as best as possible, whether a file is a particular format or not within <20 lines or so.  Most classes do it within 5 lines.

Also note that the format ID doesn't have to be perfect -- it just has to catch *most* of the use-cases (for some definition of "most").  I would base the format ID on the header of each section and stipulate that it must occur within 5 lines or so (and only ignore blank lines or ones with valid comments, if that exists).

hainm:
How's about detecting 'V2000' or 'V3000'?

https://en.wikipedia.org/wiki/Chemical_table_file#Molfile

parmed/formats/sdf.py
```diff
@@ -3,6 +3,7 @@
 extension described at https://www.cas.org/content/chemical-suppliers/example-sdf
 """
 from __future__ import print_function, division, absolute_import
+import linecache
```
swails: I've never heard of this module.  Is it a built-in?  Is it in Python 2?

parmed/formats/sdf.py
```diff
@@ -27,17 +28,11 @@ def id_format(filename):
         is_fmt : bool
             True if it is a sdf file, False otherwise
         """
-        f = genopen(filename, 'r')
+        line = linecache.getline(filename, lineno=4)
```
swails: Will it *always* be line # 4?

parmed/formats/sdf.py
```diff
@@ -27,17 +28,11 @@ def id_format(filename):
...
         try:
-            for line in f:
-                if not line.strip(): continue
-                if line.startswith('M  END') or line.startswith('$$$$'):
-                    return True
-                else:
-                    continue
+            return line.split()[-1] in ['V2000', 'V3000']
```
swails: Can we also test the number of words in that line?  So something like this:

```python
words = line.split()
return len(words) = <blah> and words[-1] in ('V2000', 'V3000')
```
?

parmed/formats/sdf.py
hainm: it's in standard lib: https://www.google.com/search?q=linecache&ie=utf-8&oe=utf-8

parmed/formats/sdf.py
hainm: Yes I think so. According to what I read: https://en.wikipedia.org/wiki/Chemical_table_file

parmed/formats/sdf.py
hainm: let me check.

hainm: seem we can not check the number of words. 

For example: n_words = 8 in http://infochim.u-strasbg.fr/recherche/Download/Fragmentor/MDL_SDF.pdf, page 78; n_words = 10 for the `test.sdf` in this PR.

parmed/formats/sdf.py
```diff
@@ -0,0 +1,38 @@
...
+            Name of the file to test whether or not it is a sdf file
+
+        Returns
+        -------
+        is_fmt : bool
+            True if it is a sdf file, False otherwise
+        """
+        line = linecache.getline(filename, lineno=4)
+        try:
+            return line.split()[-1] in ['V2000', 'V3000']
```
swails: Going based on the file format spec at https://en.wikipedia.org/wiki/Chemical_table_file#Molfile, I would make one small change.  That line says that the first two numbers are the number of atoms and number of bonds, and gives no meaning to the rest of the numbers.  What this tells me is that that field *must* have at least 3 words.  So I would change this to the following:

```python
words = linecache.getline(filename, lineno=4).split()
return len(words) >= 3 and words[-1] in ('V2000', 'V3000')
```

Easy peasy :)

hainm: cool. updated.

parmed/formats/sdf.py
```diff
@@ -27,10 +27,11 @@ def id_format(filename):
...
             True if it is a sdf file, False otherwise
         """
-        line = linecache.getline(filename, lineno=4)
         try:
-            return line.split()[-1] in ['V2000', 'V3000']
+            words = linecache.getline(filename, lineno=4).split()
+            return len(words) >= 3 and words[-1] in ('V2000', 'V3000')
         except IndexError:
+            # 4-th line is empty
             return False
```
swails: This doesn't have to be in a try-except block. If the 4th line is empty, `len(words) >= 3 and words[-1] in ('V2000', 'V3000')` will short-circuit and return `False`.  You literally *just* need those two lines :).

hainm: I see. So Python does not check 'words[-1]` if `len(words) == 0`.

swails: That's true for any language that short-circuits.  Most languages guarantee short-circuiting in the language standard.  Examples are C (and anything derived from C like C++, Java, Go, C#, etc.), Python, ruby, Swift, perl, and probably awk.

In fact, the only language I know *doesn't* guarantee it in the spec is Fortran.  But the major compilers all implement it, anyway.  What this means is you can do something like this in C:

```c
if (denom != 0 && numerator/denom > 0) {
    // Add some code
}
```

and be guaranteed not to get a divide-by-zero error. But if you tried that in Fortran, the standard does not require compilers to skip the divide-by-zero (although I've used that kind of pattern before and never seen it fail in practice with modern F90 compilers).

Short-circuiting is also responsible for these tricks in Python:

```python
import random

def stupid():
    a = random.random() - 0.5
    mylist = []
    if a > 0:
        mylist.append(0)
    return mylist or 'The list was empty!'

stupid()
```

If you run this a number of times, you'll see output like the following:

```python
>>> stupid()
'The list was empty!'
>>> stupid()
[0]
>>> stupid()
'The list was empty!'
>>> stupid()
'The list was empty!'
>>> stupid()
[0]
```

As a general rule, boolean operations will return the first value that duplicates the total expression's boolean return value.  So if `mylist` is filled with a 0 (which happens ~50% of the time), then `mylist or 'The list was empty!'` evaluates to `True` because `mylist` is boolean `True`.  Short-circuiting allows the boolean value of `mylist` to represent the boolean value of the whole expression.  On occasions where `mylist` does not get appended, the empty list evaluates to `False` and the second value in the expression (the static string `The list was empty!`) represents the boolean value of the expression, so it gets returned instead.

In fact, this even works if the second value is boolean `False` (try replacing `The list was empty` with `False` and you'll see it returns both `[0]` and `False`.  I like this trick and use it on occasion where it simplifies some conditionals.  [See here for example](https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/netcdffiles.py#L777-L778):

```python
    def __del__(self):
        self.closed or (hasattr(self, '_ncfile') and self._ncfile.close())
```

If `self.closed` is True, it satisfies the whole conditional and the second part is skipped.  Otherwise, it moves on to `(hasattr(self, '_ncfile') and self._ncfile.close())`.  If the object died in the constructor, `_ncfile` will not be set as an attribute, so the `hasattr` call will be `False`.  Since it's connected by `and`, the second part of that conditional is short-circuited and the `AttributeError` is avoided.  Therefore, `self._ncfile.close()` is only executed if and only if the object went through its constructor successfully *and* it was not closed explicitly before.  The equivalent code is this:

```python
    def __del__(self):
        if not self.closed and hasattr(self, '_ncfile'):
            self._ncfile.close()
```

But I like my trick, even if I'd be yelled at in the corporate world for using it :smiley:.

One of the main places I use this kind of trick is doing something like this:

```python
return return_value or []
```

to return an empty list instead of `None` in some relevant situations.  Because that replaces something like this:

```python
if return_value is None:
    return []
else:
    return return_value
```

or more succinctly

```python
if return_value is None:
    return []
return return_value
```

hainm: cool. thanks for very detail explanation.

swails:
Remind me here -- is this ready to merge?  Were all of the comments addressed?

hainm:
yes, ready to merged, followed all of your comments.

swails:
Great!

swails:
Thanks!

hainm:
thanks2

PR #749, opened 2016-08-28
add visualization

hainm:
I think it's convenient to add this. What do you think? 

```python
parm.visualize()
parm[':1-20'].visualize()
```

parmed

There is no test yet since I am not sure if you are ok with this PR.

Another advantage is amber user can use `parmed` to visualize netcdf trajectory (and others) on window.

### DIFF
--- a/devtools/travis-ci/install.sh
+++ b/devtools/travis-ci/install.sh
@@ -34,7 +34,7 @@ else # Otherwise, CPython... go through conda
         conda install -y -n myenv boost==1.59.0 -c omnia
         # uninstall parmed from ambermd
         conda uninstall -y parmed -n myenv
-        conda install nglview -y -c bioconda
+        conda install -y -n myenv nglview -c bioconda
     else
         # Do not install the full numpy/scipy stack
         conda create -y -n myenv python=$PYTHON_VERSION numpy nose pyflakes=1.0.0 \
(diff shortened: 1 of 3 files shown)


swails:
Cool!  👍 -- just let me know when to merge.

hainm:
it's good for review now.

swails:
If you add that to the documentation somewhere I can regenerate it.  Do you want to add docs (with examples) to this PR, or another one?

hainm:
do you mean docs for the website? I will add another PR.

swails:
> do you mean docs for the website? I will add another PR.

Sounds good.  I'll go ahead and merge this then.

By the way, does this only work in Jupyter Notebook?  I presume so since nglview is javascript-based?

hainm:
> does this only work in Jupyter Notebook? I presume so since nglview is javascript-based?

you're right. This is for Jupyter notebook (as noted in the doc).

```python
+    def visualize(self, *args, **kwargs):
 +        """Use nglview for visualization. This only works with Jupyter notebook
 +        and require to install `nglview`
```

btw, I will add `parmed` as a backend for `mdsrv` (that only require web-browser).

https://github.com/Amber-MD/mdsrv

hainm:
@swails where can I upload images files? to this repo? or just a link?

swails:
in the doc directory, you mean?

If you can host them somewhere without adding them to the repo that would be best, unless they are small or you promise not to change them :).  Git does poorly managing versions for binary files, so frequent updates of images or PDFs causes Github repos to grow too large too fast.

hainm:
>  you promise not to change them 

ha ha, won't promise. I know the pain for binary format. I think I would create a dummy git repo and upload to that.

swails:
how many images are we talking about here?

hainm:
2-3, but i think it's better not to host in main repo.

PR #754, opened 2016-09-08
Replace `1/6` and similar expressions with `1.0/6`

tom--lee:
There are quite a few instances of the expression `1/6` being used for the value of one sixth, especially when raising to the 1/6th power such as when converting between sigma and r_min. Being integer division, `1/6` evaluates to 0. Worse still, `2**(1/6)` evaluates to 1 rather than 1.12246... so that you could end up with force parameters that kind of look right but aren't. It is possible that the error was not detected by tests because some test files contained the same mistake.

I made the replacements through brute-force use of sed, searching for `1/6` and `1 / 6`. I can't guarantee I didn't miss any similar unintentional integer divisions.

### DIFF
--- a/parmed/amber/_amberparm.py
+++ b/parmed/amber/_amberparm.py
@@ -796,7 +796,7 @@ class AmberParm(AmberFormat, Structure):
         self.LJ_radius = []  # empty LJ_radii so it can be re-filled
         self.LJ_depth = []   # empty LJ_depths so it can be re-filled
         self.LJ_types = {}   # empty LJ_types so it can be re-filled
-        one_sixth = 1 / 6    # we need to raise some numbers to the 1/6th power
+        one_sixth = 1.0/6    # we need to raise some numbers to the 1.0/6th power
 
         pd = self.parm_data
         acoef = pd['LENNARD_JONES_ACOEF']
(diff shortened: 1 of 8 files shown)


tom--lee:
I just realized that this is only a problem in Python 2. Dividing two integers in Python 3 gives a float. I guess that might be how this arose.

swails:
This is a severe problem if it actually occurs, and will result in a great number of the tests failing.  You can verify this by replacing "1/6" everywhere you see it with "1//6" or "1 / 6" by "1 // 6" (which forces integer division in both Python 2 and Python 3).

All of the tests pass for both Python 2 and Python 3, though, since every file in ParmEd (or at the very least every file that actually performs division) import future division, which makes division in Python 2 behave like Python 3.  For example:

```python
Python 2.7.11 (default, Mar  1 2016, 18:40:10) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> 1 / 6
0
>>> 2 ** (1/6)
1
```

Shows the problem you mentioned.  However, prefixing with the right division import:

```python
Python 2.7.11 (default, Mar  1 2016, 18:40:10) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.2 (clang-700.1.81)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from __future__ import division
>>> 1 / 6
0.16666666666666666
>>> 2 ** (1/6)
1.122462048309373
```

and the problem does not occur.  You can check, and every file that you made this change also contains a `from __future__ import division`.  Also, since all of the tests will most certainly catch this problem, I'm not concerned about it creeping in unnoticed.

tom--lee:
Ah, that makes a lot more sense, apologies for not checking more thoroughly before submitting this. I had no idea that the division behavior could be changed so dramatically!

Issue #762, opened 2016-09-28
PDBWarning: ANISOU record does not match previous atom


hainm:
we got lots of warning for `1gdu`.

```python
pmd.download_PDB('1gdu')
```

trace (edited to fit to screen)
```bash
... parmed/parmed/formats/pdb.py:508: PDBWarning: ANISOU record does
 not match previous atom
  'atom', PDBWarning)
```

swails:
Should be fixed in #763

PR #763, opened 2016-09-30
Fix erroneous warnings with ANISOU records

swails:
The mistake was very dumb.  I was parsing the insertion code from ANISOU records in the 28th character field instead of the 27th.

This might also fix another bug when two residues with different insertion codes come right after each other but are the same kind of residue.

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -25,7 +25,7 @@ import warnings
 
 #++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
-def _compare_atoms(old_atom, new_atom, resname, resid, chain, segid):
+def _compare_atoms(old_atom, new_atom, resname, resid, chain, segid, inscode):
     """
     Compares two atom instances, along with the residue name, number, and chain
     identifier, to determine if two atoms are actually the *same* atom, but
@@ -45,6 +45,8 @@ def _compare_atoms(old_atom, new_atom, resname, resid, chain, segid):
         The chain identifier that the new atom would belong to
     segid : ``str``
         The segment identifier for the molecule
+    inscode : ``str``
+        The insertion code for the residue
 
     Returns
     -------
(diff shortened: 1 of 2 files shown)


swails:
Fixes #762

swails:
> Should this have assertion?

Not necessary.  This was a classic TDD flow.  I wrote the test first, ran the code, and verified that it failed the way I expected it to fail (since `setUp` promotes `PDBWarning` to a fatal error).

I implemented the fix and the test passed.  That's really all that needs to happen for that test to serve its purpose.  There are lots of other tests that do detailed comparisons.

hainm:
> since setUp promotes PDBWarning to a fatal error

I see.

dacase:
Same sort of typo-like misreading of the residue name in pdb.py.  Line 518 reads:
```
rname = line[17:21].strip()
```

But the index should be `17:20`.  Compare line 250, where the residue name is extracted from an ATOM record.  Since the only purpose of getting `rname` from the ANISOU card is to check that it matches the residue name from the assoicated ATOM record, this looks like a clear typo.  I guessing ParmEd users don't see a lot of PDB files with ANISOU records, or you would be seeing errors.

Apologies for not just submitting a merge request: I've been trying to figure out how to fix lines 548-549, which don't work at all for me, and don't give any relevant information about `which` ANISOU card is misbehaving.  But am sort of swamped with other things right now.

dacase:
This is still a definite problem, and I don't think things are really fixed.  But the complexity of the python code at this point is prevented me from easily proposing a pull request.  Some notes:

* The (presumed) error in line 518 of pdb.py (see previous comment) is still there.
* When a mismatch is found, the error message printed to stdout is garbled.  Here is an example:
```
/Users/case/phenix/conda_base/lib/python2.7/site-packages/parmed/formats/pdb.py:552: PDBWarning: ANISOU record does not match previous atom
  'atom', PDBWarning)
```

I'm not sure what `'atom'` part right before "PDBWarning" is supposed to be here.  But the absence of any information about *which* ANISOU card raised the error makes it really difficult to localize the problem.  (This part of the comment amplifies the somewhat vague complaint about lines 548-549. It would be neat to (optionally?) print out the offending ANISOU line itself, but I don't know the correct way to do that in this context.

Issue #631, opened 2016-03-04
Transfer Amber topology and coordinate files to CHARMM or GROMACS formats


pengfeili1:
Hi Jason,

I want to use ParmEd to transfer AMBER topology and coordinate files to CHARMM and GROMACS formats. Can you tell me how to do that? Thanks!

Warm regards,
Pengfei

swails:
Hey, sorry!  This is quite easy:

```Python
import parmed as pmd

amber = pmd.load_file('prmtop', 'inpcrd')

# Save a GROMACS topology and GRO file
amber.save('gromacs.top')
amber.save('gromacs.gro')

# Save a CHARMM PSF and crd file
amber.save('charmm.psf')
amber.save('charmm.crd')
```

Doesn't get much easier :smiley:

swails:
Here are the docs for `.save` -- it picks the default file type via filename extension:

```
save(self, fname, format=None, overwrite=False, **kwargs)
    Saves the current Structure in the requested file format. Supported
    formats can be specified explicitly or determined by file-name
    extension. The following formats are supported, with the recognized
    suffix and ``format`` keyword shown in parentheses:
    
        - PDB (.pdb, pdb)
        - PDBx/mmCIF (.cif, cif)
        - PQR (.pqr, pqr)
        - Amber topology file (.prmtop/.parm7, amber)
        - CHARMM PSF file (.psf, psf)
        - CHARMM coordinate file (.crd, charmmcrd)
        - Gromacs topology file (.top, gromacs)
        - Gromacs GRO file (.gro, gro)
        - Mol2 file (.mol2, mol2)
        - Mol3 file (.mol3, mol3)
        - Amber ASCII restart (.rst7/.inpcrd/.restrt, rst7)
        - Amber NetCDF restart (.ncrst, ncrst)
    
    Parameters
    ----------
    fname : str
        Name of the file to save. If ``format`` is ``None`` (see below), the
        file type will be determined based on the filename extension. If the
        type cannot be determined, a ValueError is raised.
    format : str, optional
        The case-insensitive keyword specifying what type of file ``fname``
        should be saved as. If ``None`` (default), the file type will be
        determined from filename extension of ``fname``
    overwrite : bool, optional
        If True, allow the target file to be overwritten. Otherwise, an
        IOError is raised if the file exists. Default is False
    kwargs : keyword-arguments
        Remaining arguments are passed on to the file writing routines that
        are called by this function
    
    Raises
    ------
[...]

PR #767, opened 2016-10-11
amber to others: example

hainm:
update example from your comment: https://github.com/ParmEd/ParmEd/issues/631#issuecomment-193555313


swails:
The failure makes no sense -- looking at the log, the build succeeded.  I'm fine merging this.

PR #769, opened 2016-10-22
fix bytes + add some tests for minimization + increase code coverage

hainm:
Still got problem with 
`Problem minimizing structure with scipy and sander:
b'\tSTOP: TOTAL NO. of ITERATIONS EXCEEDS LIMIT'
`

### DIFF
--- a/parmed/tools/simulations/sanderapi.py
+++ b/parmed/tools/simulations/sanderapi.py
@@ -131,4 +131,4 @@ def minimize(parm, igb, saltcon, cutoff, tol, maxcyc):
     if not results.success:
         print('Problem minimizing structure with scipy and sander:',
               file=sys.stderr)
-        print(b'\t' + results.message)
+        print('\t' + results.message.decode())
--- a/parmed/tools/simulations/sanderapi.py
+++ b/parmed/tools/simulations/sanderapi.py
@@ -23,6 +23,12 @@ def energy(parm, args, output=sys.stdout):
     """
     Compute a single-point energy using sander and print the result to the
     desired output
+
+    Parameters
+    ----------
+    parm : Structure
+    args : ArgumentList
+    output : file handler, default sys.stdout
     """
     global HAS_SANDER
     if not HAS_SANDER:
--- a/parmed/tools/simulations/sanderapi.py
+++ b/parmed/tools/simulations/sanderapi.py
@@ -131,4 +131,4 @@ def minimize(parm, igb, saltcon, cutoff, tol, maxcyc):
     if not results.success:
         print('Problem minimizing structure with scipy and sander:',
               file=sys.stderr)
-        print('\t' + results.message)
+        print(b'\t' + results.message)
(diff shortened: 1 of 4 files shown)


parmed/tools/simulations/sanderapi.py
```diff
@@ -131,4 +131,4 @@ def energy_function(xyz):
     if not results.success:
         print('Problem minimizing structure with scipy and sander:',
               file=sys.stderr)
-        print('\t' + results.message)
+        print(b'\t' + results.message)
```
swails: Why not just decode results.message?

hainm:
> Why not just decode results.message?

Because I only need to add a 'b'.

swails:
The error being reported is this:

```
  File "/home/travis/build/ParmEd/ParmEd/test/test_parmedtools_actions.py", line 1809, in test_minimize_sanderapi
    sanderapi.minimize(parm, igb=igb, saltcon=0., cutoff=999., tol=1E-5, maxcyc=10)
  File "/home/travis/miniconda/envs/myenv/lib/python3.4/site-packages/ParmEd-2.6.1+5.g34463d5-py3.4-linux-x86_64.egg/parmed/tools/simulations/sanderapi.py", line 103, in minimize
    raise SimulationError('Could not import sander')
parmed.tools.exceptions.SimulationError: Could not import sander
```

Any reason why this could be happening?

hainm:
I don't know. I will check tmr.

PS: github review comments are so complicated now :(

hainm:
ok, so I know why that (sander stuff) happened: conda is getting weirder.

Originally we got trouble with `parmed` from `ambermd` channel (`parmed` comes with `pysander`), so we (me) uninstall it. But with current conda version, uninstalling `parmed` will uninstall `pysander` too. 

Will try to solve it.

hainm:
Good for review now

- test failed with osx due to the title travis issue, not due to Parmed
- significantly increased code coverage for sanderapi
- update doc

swails:
conda remove --force

Should do it.  

--
Jason M. Swails

hainm:
> conda remove --force

ah, I just comment 'conda uninstall ...' and let ParmEd master overwrites the version in ambermd.

hainm:
I don't know. I will investigate tmr.

swails:
We should have paid attention here.  If we look at coveralls, the coverage dropped to 17% (!!)

What happened was that a bug in pysander (chamber setup is not cleaned up completely, so running a "normal" prmtop after running a chamber-style prmtop results in an error message about missing prmtop flags) caused a system exit to be called.  But the exit code was 0, so Travis did not flag it as a failure.

The coverage was so horrible (dropped from ~92% to 17%) because most of the tests didn't even get run (and any that may have failed just got masked by the sysexit call from pysander).

I'm fixing all of this in #783, but I should be more careful about inspecting things that look out of place (like a 75% coverage hit on a commit that should have improved test coverage) before merging...

PR #776, opened 2016-11-10
Read velocities from amber files if available

ctk3b:
I encountered this while playing with one of the InterMol amber [input files](https://github.com/shirtsgroup/InterMol/blob/master/intermol/tests/amber/stress_tests/solv/solv.rst7). This fix appears to work for my use case but I am not very familiar with the amber portions of ParmEd. Will this interfere with anything else?

### DIFF
--- a/parmed/amber/_amberparm.py
+++ b/parmed/amber/_amberparm.py
@@ -217,7 +217,7 @@ class AmberParm(AmberFormat, Structure):
             if not hasattr(f, 'coordinates') or f.coordinates is None:
                 raise TypeError('%s does not have coordinates' % xyz)
             self.coordinates = f.coordinates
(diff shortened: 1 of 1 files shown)


hainm:
> This fix appears to work for my use case

what does it fix? and what did you encounter?

ctk3b:
Velocities simply aren't read in even if the file contains them.

ctk3b:
Failures appear to be related to an issue with Travis OSX machines: https://github.com/travis-ci/travis-ci/issues/6307

ctk3b:
Any further insights here? I don't think this should be intrusive - the velocity data is currently simply being ignored.

swails:
Sorry, I thought this was already merged. I clicked merge in my phone, but it didn't work because the CI failed. I'll merge it now. Change looks good.

PR #777, opened 2016-11-10
Improve logging in tests and allow OS X tests to fail.

swails:


swails:
I incorporated the warning reduction into #783

PR #789, opened 2016-12-03
Fix an obscure bug when converting to Amber topology in some cases.

### LINKED ISSUE #784: Problems when writing and then reading a prmtop and crd from a Structure
I'm attempting to convert an OpenMM system we created from our new SMIRFF forcefield format into AMBER format via ParmEd, and successfully am able to write a `prmtop` and `crd` with ParmEd, but when I load them again, I get this error:

```
/Users/dmobley/anaconda/lib/python2.7/site-packages/simtk/openmm/openmm.pyc in addException(self, particle1, particle2, chargeProd, sigma, epsilon, replace)
  14854             the index of the exception that was added
  14855         """
> 14856         return _openmm.NonbondedForce_addException(self, particle1, particle2, chargeProd, sigma, epsilon, replace)
  14857 
  14858 

Exception: NonbondedForce: There is already an exception for particles 2612 and 2615
```

I'm attaching the prmtop and crd in question which I'm trying to load. The issue can be reproduced by using:
`structure = parmed.amber.AmberParm( 'system.prmtop', 'system.crd')`

It seems possible there are actually two issues -- i.e. it's possible something is amiss in MY `System` which results in an incorrect prmtop and crd being written that then can't be read back in properly. However, it's probably a ParmEd bug that the system WRITES properly but fails on read.

If you'd like to see this in context, I have an IPython notebook in the `issue184` branch of `smarty` which is where I discovered the problem. I'm attempting to convert a simple mixture of cyclohexane and ethanol from being an OpenMM system into an AMBER prmtop and crd. Here's the IPython notebook: https://github.com/open-forcefield-group/smarty/blob/issue184/examples/testing/test_system_conversion/convert_to_amber.ipynb
and here's the branch (see `smarty/examples/testing/test_system_conversion` for input files): https://github.com/open-forcefield-group/smarty/tree/issue184

[Archive.zip](https://github.com/ParmEd/ParmEd/files/607403/Archive.zip)

swails:
This gets rid of the dumb ``signs`` attribute that I had completely forgotten I
added (one of the early vestiges left over from the original code so many years
ago). It now uses ignore_end and improper to get the signs to be added to the
Amber topology file.

Also delete a note that is no longer accurate.

Fixes #784


davidlmobley:
@swails - I'm testing this. However, the version string seems not to be correct (I'm getting 0, 0, 0, 0). Maybe that's automatic? In any case, I have to adjust some code elsewhere as it won't test if the version is not at least 2.5.4.

davidlmobley:
@swails - aside from the problem with the version string, this seems to fix my problem with the duplicate exception parameter error as described in #784.

The other issue discussed in https://github.com/open-forcefield-group/smarty/pull/187 remains open. Shall I create a separate issue on it, or are you clear enough as to the problem?

swails:
> @swails - I'm testing this. However, the version string seems not to be correct (I'm getting 0, 0, 0, 0). Maybe that's automatic?

The version string is built automatically by [Versioneer](https://pypi.python.org/pypi/versioneer/), which extracts it from git tags and revision history.  What this means is that if you don't have the same tags in your local repository as I do in mine (or as this repository has), then versioneer won't generate the correct version string.

That version string is then parsed to create the actual version tuple that you can compare against.  You can use the command `git fetch -t <remote>` where `<remote>` is *this* remote repository.  That will fetch the tags loaded in this remote to your local repository.  I tried managing the version string by hand myself and that (predictably) proved to be unmaintainable and unreliable.  Versioneer has its drawbacks, but on the whole is the best solution I've found for simple versioning.

> Shall I create a separate issue on it, or are you clear enough as to the problem?

Another issue here would help me keep track of things.  If I recall correctly, it's a problem with 1-4 scaling factors.

davidlmobley:
I'll try to create another issue in a bit, but yes, it's an issue with 1-4 scaling factors getting lost/becoming inconsistent somehow.

swails:
I was waiting to merge this until the 1-4 scaling issue got figured out.  It seems that it's all worked out, so am I good to merge this now?

davidlmobley:
Yes. Thanks!

davidlmobley:
(Could you also perhaps cut a release with this fix somewhat soon? I'm going to end up incorporating a test into Travis-CI for `smarty` relatively soon which may trigger this issue, so it would be nice to get it into a release fairly soon.)

swails:
Yea, I'll release an update to AmberTools and cut a release sometime within the week or so.

PR #794, opened 2016-12-08
Fix some issues in GROMACS conversion

### LINKED ISSUE #791: Incorrect charge on an atom after reading to and writing from GROMACS format
I've found a bug of some sort where an atom in a residue ends up with an incorrect charge after writing to GROMACS format (or it has its charge swapped with another symmetry-equivalent atom, or atoms are re-named and re-ordered).

Inputs and outputs are attached, and to reproduce it, please use this:
```python
#!/bin/env python

import parmed
from simtk.openmm.app import PDBFile
import simtk.openmm.app as app

# Load PDB
pdbfile = PDBFile('cyclohexane_ethanol_0.4_0.6.pdb')

# Load AMBER system
parm = parmed.load_file('out.prmtop', 'out.crd')
ambersys = parm.createSystem(nonbondedMethod= app.NoCutoff, constraints = None, implicitSolvent = None)
print(parm.atoms[126].charge, parm.atoms[126].name)


# Save to GROMACS
structure = parmed.openmm.topsystem.load_topology( pdbfile.topology, ambersys, pdbfile.positions )
# Same behavior with this:
#structure = parmed.openmm.topsystem.load_topology( parm.topology, ambersys, parm.positions )
print(structure.atoms[126].charge, structure.atoms[126].name)
structure.save( 'bug.top', overwrite = True, format="gromacs")
structure.save( 'bug.gro', overwrite = True, format = "gro")

# Load GROMACS system
top = parmed.load_file('bug.top')
print(top.atoms[126].charge, top.atoms[126].name)
```

Output is:
```
(-0.07549999670733112, 'C1')
(-0.07549999670733112, 'C1')
(-0.0761, 'C1')
```

As you can see, `C1` in residue `ZBE` (which happens to be cyclohexane) has a charge magnitude slightly less than 0.076 before GROMACS format is involved, and slightly over 0.076 after GROMACS format is involved. This isn't rounding -- the carbon atoms in cyclohexane all have charges of -0.0755 except this one, which is (for some reason) -0.0761. In AMBER, this charge appears to end up on atom 128; after a round-trip from GROMACS, it ends up on atom 126.

@swails , any insights? In my test, this is making it so energies are not identical after conversion to GROMACS format.

[Archive.zip](https://github.com/ParmEd/ParmEd/files/637654/Archive.zip)

swails:
Fixes #791 

To do:

* [x] Fix charge scale determination
* [x] Fix molecule 'equivalence' testing

The second one is going to be harder to do since right now the equivalence check is based on atom and residue names... ☹️


davidlmobley:
Thanks, @swails . This seems to have fixed the immediate issue, at least in my first test, but I am running into a secondary issue which may be a different problem or may be something on my end. I'll update when I have more info.

davidlmobley:
Actually, no. It seems my IPython notebook I was running had gone wonky and was running into a different, earlier issue. After restarting it, I'm still having the same basic issue. I'll try and troubleshoot further and get you a minimal test case, @swails .

I also have a separate issue (which is somehow related) earlier where, when I take my initial system, dump to AMBER format, and read it back in, I get different scaling factors for the 1-4 parameters, even before involving GROMACS at all. I'm still struggling with what the best way is to allow you to reproduce this easily, but I'll see if I can do it by serializing and deserializing.

swails:
> After restarting it, I'm still having the same basic issue.

What issue is this?  Throwing an exception converting to GROMACS due to variable 1-4 scaling?

If you're using any Amber XML system files, can you regenerate those in case they were generated with an older ParmEd version that didn't properly carry exceptions over?

davidlmobley:
> What issue is this? Throwing an exception converting to GROMACS due to variable 1-4 scaling?

Yes, sorry for being unclear.

I've also run into a few problems in my python installation and needed to finish migrating to Python3.5, so I just wiped everything and started over from scratch in Python 3.5, so I'll soon test on a clean install.

I'm re-generating everything from scratch -- not using any old files.

davidlmobley:
OK, @swails - I have a case to reproduce (one) of the problems I'm currently seeing.

I really have at least two problems at present (only one of which is directly related to your present fix):
1. My original system has different exclusion parameters before and after saving to AMBER format
2. Taking my AMBER format system to GROMACS results in different exclusion parameters

Currently, we're focusing on fixing the second of these, but I'll try to create a test case for you to be able to look at the first as well. 

To reproduce the second, please use the attached files and this code:
```python
#!/bin/env python

import parmed
from simtk.openmm.app import PDBFile
import simtk.openmm.app as app

# Load PDB
pdbfile = PDBFile('cyclohexane_ethanol_0.4_0.6.pdb')

# Load AMBER system
parm = parmed.load_file('out.prmtop', 'out.crd')
ambersys = parm.createSystem(nonbondedMethod= app.NoCutoff, constraints = None, implicitSolvent = None)


# Save to GROMACS
#structure = parmed.openmm.topsystem.load_topology( pdbfile.topology, ambersys, pdbfile.positions )
structure = parmed.openmm.topsystem.load_topology( parm.topology, ambersys, parm.positions )
structure.save( 'bug.top', overwrite = True, format="gromacs", combine='all')
structure.save( 'bug.gro', overwrite = True, format = "gro")

# Load GROMACS system
gromacstop = parmed.load_file('bug.top')
gromacstop.positions = parm.positions

# Print AMBER energies
print(parmed.openmm.energy_decomposition_system(parm, parm.createSystem()))
print(parmed.openmm.energy_decomposition_system(gromacstop, gromacstop.createSystem()))
```

The output is rather different in the nonbonded term:
```
[('HarmonicBondForce', 42.56436386960996), ('HarmonicAngleForce', 109.54924584764584), ('PeriodicTorsionForce', 381.4186716756584), ('NonbondedForce', 13022.850374108442), ('CMMotionRemover', 0.0)]
[('HarmonicBondForce', 42.56436386960996), ('HarmonicAngleForce', 109.54924584764584), ('PeriodicTorsionForce', 381.41867156519044), ('NonbondedForce', 13465.541535695964), ('CMMotionRemover', 0.0)]
```

Separate code from `openmoltools` tells me this is because of different exclusion interactions, such as this:
```
Error: NonBondedForce Exception atom (1542, 1545) has squared charge values of 0.001426 and 0.001188 (elementary charge)**2, respectively.
```
[cyclohexane_ethanol_0.4_0.6.pdb.zip](https://github.com/ParmEd/ParmEd/files/640365/cyclohexane_ethanol_0.4_0.6.pdb.zip)
[Archive 2.zip](https://github.com/ParmEd/ParmEd/files/640367/Archive.2.zip)

swails:
> To reproduce the second, please use the attached files and this code:

The out.prmtop file here is broken.  None of the dihedrals here are indicated that the end-group interactions are to be ignored, which I know to be false based on previous fixes I've done.  What this means is that the exceptions in out.prmtop are already broken, so converting this to GROMACS *should* fail.

If I use `out.prmtop` and `out.crd` from your original `Archive.zip` file, then I get this:

```
[('HarmonicBondForce', 42.564197250290974), ('HarmonicAngleForce', 109.54929866522606), ('PeriodicTorsionForce', 381.4186968652314), ('NonbondedForce', 12147.284217827897), ('CMMotionRemover', 0.0)]
[('HarmonicBondForce', 42.564197250290974), ('HarmonicAngleForce', 109.54929866522606), ('PeriodicTorsionForce', 381.41869674776575), ('NonbondedForce', 13739.695550579163), ('CMMotionRemover', 0.0)]
```

Something clearly went wrong here, so I need to investigate.  However, the variable 1-4 scaling exception should now be fixed.

davidlmobley:
@swails - so, yes, I don't have a problem with the variable 1-4 scaling exception anymore. I do still have a problem with exceptions if I take my original system, dump to prmtop/crd, re-load, and compare to the original system. This is not exactly the same as the issue I described here: https://github.com/ParmEd/ParmEd/issues/791#issuecomment-265833833 since that issue just results in a difference in nonbonded forces, and this one results in a difference in 1-4 electrostatics, but I'm guessing that the source might be the same. (Possibly something is wrong in my original system which (a) results in a difference in nonbonded forces when dumping to XML and AMBER and reading back in, but (b) results only in a difference in 1-4 electrostatics if I just dump to AMBER and read back in...)

Please let me know if you have suggestions on what I can look at to troubleshoot on my end/make things easier for you. Obviously there's at least one problem still on the ParmEd end of things (otherwise the nonbonded energy would remain the same after dumping to prmtop/crd and reading back in) but there may be something wrong on my end too.

swails:
Got it!  The problem here was actually not the 1-4 electrostatic scaling factor, but rather the Lennard-Jones scaling factor.  I found out that my last fix (which @ctk3b described as taking the scaling factor from the `adjusts` list rather than the dihedral list) didn't actually copy over the Lennard-Jones factor.

This was actually intentional, as I was relying on GROMACS to print out pair-specific Lennard-Jones parameters for each of the exceptions.  The only problem is that I didn't set the flag that triggers the GROMACS writer to actually do this.  Once I did, the energy agreed to within 0.1 kcal/mol.

That last little bit was due to the fact that in the GROMACS file, I write the pair-specific parameters out to 5 decimal points.  If I swap out an exact scaling factor, the agreement improves to machine precision.  If I increase the precision to 8 decimal points, I get agreement to 0.002 kcal/mol, which I'm satisfied with.

Re-running the script above now gives me this output:

```
[('HarmonicBondForce', 42.564197250290974), ('HarmonicAngleForce', 109.54929866522606), ('PeriodicTorsionForce', 381.4186968652314), ('NonbondedForce', 12147.284217827897), ('CMMotionRemover', 0.0)]
[('HarmonicBondForce', 42.564197250290974), ('HarmonicAngleForce', 109.54929866522606), ('PeriodicTorsionForce', 381.41869674776575), ('NonbondedForce', 12147.28675499277), ('CMMotionRemover', 0.0)]
```

@davidlmobley -- I suspect that you have some weird versions lying around if you are seeing 1-4 electrostatic constants differing after file conversions.  I haven't seen that happen since I first issued this PR.

Try the latest version.  The last thing I have left to do for this is to implement proper molecule hashing so I don't inadvertently assign two molecules that appear to be the same to the same moleculetype if they have different parameters.  That way, the `combine=None` default will work as expected and give identical results to `combine='all'`.

swails:
OK, this pull request is complete from my end.  Here's what's been fixed:

1. Exceptions are properly translated into GROMACS if they come from a source that explicitly defines them (such as OpenMM). This is true for both L-J and electrostatic (the former is handled the most general way possible -- i.e., the way CHARMM does it with pair-specific L-J parameters for 1-4s)
1. 1-4 scaling factors are now taken properly from AMBER files by ignoring the scaling factors on torsions whose end-groups are ignored (fixes the original exception)
1. Molecules are assigned distinct types now if their nonbonded parameters differ from other molecules that have the same atoms and residues.

Let me know if there are still outstanding issues, and I'll merge it (and cut a new release).

davidlmobley:
@swails - I just wiped all my old files and re-ran all of my tests and **now, everything passes** -- comparing my original system to the system after re-reading from AMBER format, comparing AMBER to GROMACS, comparing original to GROMACS, etc. **NICE WORK!**

Thanks very much for all your effort on this. I really appreciate all the time you spent on this.

I seem to have a real knack for breaking things. Sorry about that.  But thanks for all the fixes!!!

swails:
Whew!  That's good news.  Big thanks to @ctk3b, who likely saved me a lot of time with help diagnosing the problems.

> I seem to have a real knack for breaking things. Sorry about that. But thanks for all the fixes!!!

In this case, you can't break what's not already broken.  Thanks for helping me test out the corner cases.  ParmEd has been a real learning experience, and some of the design decisions made at the beginning made the corners a bit more treacherous than perhaps they had to be.

But as I was told in a small course I recently took, the beginning of a project is the time at which you have the worst knowledge of the problem that you will ever have.

I'm going to go ahead and merge this.  I'll cut a release soon.

davidlmobley:
@swails - can you do a new release?

kartheekpitta:
@swails  Is there a way to convert topologies from amber to gromacs format by using default parmed module in Ambertools16 instead of using direct ParmED from github. I have found discrepancies in the obtained free energy profiles.  I have used tleap to obtain amber parameters followed by ParmED from github to obtain gromacs topologies from amber, profile is labeled as ParmED. Similarly  inbuilt port in gromacs has been used to generate topologies and the same is labeled as Default PORT. As I am falling short of time I would like to use parmed in ambertools though ACPYPE is available.
 [comp](https://user-images.githubusercontent.com/28443937/33936146-845dd10e-e024-11e7-800e-2545b63cb4a8.png)

davidlmobley:
If you think you're seeing energy errors, @kartheekpitta , you should start a new issue and provide input files to reproduce so we can check energies, etc. No one is going to be able to identify a potential bug if you can't give more details. How do you know that the difference in profile you're seeing is not a sampling problem? (And what are you even doing?)

Is there a reason you think that ParmEd in Ambertools16 will avoid whatever problem you are seeing with ParmEd here?

kartheekpitta:
I am trying to calculate bending free energy of DNA using umbrella sampling, with 1 ns sampling time for each window in both the cases. However, I have noticed discrepancies when I tried with different reaction coordinates and models also. I just want to test with parameters generated in the Ambertools. I will be happy to provide input files.

davidlmobley:
Are the discrepancies you notice more than the discrepancies you would expect based on random variations/different trials of the same thing? This sounds to me like something rather difficult to calculate.

I expect that if you want help with the problem you will want to provide example inputs.

kartheekpitta:
In the example bending energy barrier is high for ParmED topologies which is not  acceptable based on barrier values reported in the literature. While the free energy profiles obtained from amber PORT (parmbsc1) for gromacs are in good agreement with the literature. tleap input and parm.py (conversion) are provided. Thank you !
#leap.in
---------------------------
source /home/kartheek/amber16/dat/leap/cmd/leaprc.DNA.bsc1
source /home/kartheek/amber16/dat/leap/cmd/leaprc.water.tip3p
bdna=loadpdb "8gnew26.pdb"
solvatebox dna TIP3PBOX 15 iso
addions2 dna Na+ 0
saveamberparm dna 8gnew26_ang.prmtop 8gnew26_ang.inpcrd
savepdb dna 8gnew26_amber_ang.pdb
--------------------------------------------------
#parm.py for conversion
import parmed as pmd
#amber = pmd.load_file('prmtop', 'inpcrd')
amber = pmd.load_file('8gnew26_ang.prmtop', '8gnew26_ang.inpcrd')
#save a GROMACS topology and GRO file
amber.save('8gnew26_ang.top')
amber.save('8gnew26_ang.gro')

PR #796, opened 2016-12-09
Win msys2

hainm:
- try very minimal tests (import parmed and numpy)
- got ImportError with `pwd`, so I set it as None :D (wait for your feedback)

log: https://ci.appveyor.com/project/hainm/parmed-9qurq/build/1.0.6

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -10,6 +10,10 @@ import copy
 from datetime import datetime
 import math
 import os
+try:
+    import pwd
+except ImportError:
+    pwd = None
 import re
 try:
     from string import letters
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -18,10 +18,6 @@ install:
       wget
       diffutils
     \""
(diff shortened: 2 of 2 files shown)


parmed/gromacs/gromacstop.py
```diff
@@ -10,6 +10,10 @@
 from datetime import datetime
 import math
 import os
+try:
+    import pwd
+except ImportError:
+    pwd = None
```
swails: What purpose does this serve?

hainm: if not, `python setup.py install` will fail with ImportError for `pwd`. (it's dummy fix. lol).

swails: But why?  That doesn't make any sense...  `pwd` is imported below and `ImportError` is caught.

Should the stuff that comes after `pwd` is imported be relegated to the `else` block of the try-except?

hainm: man, that `pwd` stuff was fixed a while ago and I just fork my repo, not `ParmEd/ParmEd` repo.

PR #802, opened 2016-12-15
Add atomic number guessing based on mass for PSF files.

### LINKED ISSUE #798: wrong atomic_number for PSF
We discussed about atomic_number in several issues but I still do not "feel" a final conclusion about this.
Is that `ParmEd` won't try to guess atomic_number?

Usage case:
```python
In [8]: parm = pmd.load_file('ala3.psf')

In [9]: [atom.mass for atom in parm if atom.atomic_number == 1]
Out[9]: []
```

swails:
Fixes #798


swails:
run jenkins

PR #812, opened 2017-01-11
add: get_id_format

hainm:
does parmed have it yet? 

it's good to detect format for further processing
useful case: I am rewriting [pdb4amber](https://github.com/Amber-MD/pdb4amber) using ParmEd as its backend to extend file formart support (cif). reduce program only works with `pdb`, so we can convert whatever format to `pdb` first if `reduce` is required. 

So is it
- ok to add here?
- method name?

### DIFF
--- a/parmed/formats/registry.py
+++ b/parmed/formats/registry.py
@@ -54,6 +54,20 @@ class FileFormatType(type):
                 PARSER_ARGUMENTS[name] = ()
         super(FileFormatType, cls).__init__(name, bases, dct)
 
+def get_id_format(filename):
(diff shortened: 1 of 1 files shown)


swails:
Why can't you do that with something like this:

```python
import parmed as pmd

if not pmd.formats.PDBFile.id_format(filename):
    pmd.load_file(filename).save('temporary.pdb')
```

or something like that?  Are you sure you actually need more than `load_file`?

hainm:
ah yes. I can do that. Will re-open this PR if there is any strong use for it.

Issue #778, opened 2016-11-10
RuntimeError: Cannot set cell lengths on old restart


hainm:
Hi

I got this error when trying to setting box for rst7 file
```python
In [9]: x = pmd.load_file('./1gdu_asu.rst7')
In [10]: x.box = x.box
```
trace
```bash
    390     def cell_lengths(self, stuff):
    391         if self._status == 'old':
--> 392             raise RuntimeError('Cannot set cell lengths on old restart')
    393         if not self._coords_written:
    394             raise RuntimeError('Coordinates must be written before box')
```

The I was trying to set `x._status='new'` but getting new error
```bash
RuntimeError: Coordinates must be written before box
```

Is this expected? If yes, then why is it?

swails:
Sorry, I never responded here.  If you look at the setters for these attributes, they write the values to file when you set them.  Since the format of the Amber restart file is the following:

```
title
natom
x1 y1 z1 x2 y2 z2
...
[vx1 vy1 vz1 vx2 vy2 vz2
...]
[a b c alpha beta gamma]
```

You need to set the coordinates first, the velocities next (if you want to add velocities) and the unit cell last.  Not the best API choice, but one that I made early on.  I would recommend against using that class directly -- just work with Structure or AmberParm instead.

swails:
So yes, this is expected.

hainm:
The thing is  I do not use that class directly. I used load_file to read rst7 file and got that class.

So how can I make my example in my 1st post work?

hainm:
So I assuming that there is structure=True option for that class (just like loading mol2 file)?

hainm:
@swails: what's your thought here?

Adding `structure=True` is easy enough (I do have working code), just ask about API.

swails:
The API on this file sucks, but I'm wary of changing it.

What ParmEd needs is a little bit of pytraj (without the added dependency).  It should have a "trajectory-like" class that is simply a container for phase space configurations (box, velocities, and coordinates) and have a `save` attribute that spits out any supported file types.

Then all of the trajectory parsers should create an instance of that class (or subclass) when parsing.

But that's a much bigger undertaking, and fairly low priority given that the aim is not to replace pytraj, and pytraj can already do this.

If you already have working code, though, feel free to raise a PR and I'll have a look.

hainm:
[...]

PR #815, opened 2017-01-11
save to StringIO

hainm:
Only support mol2, pdb, cif for now. It's a bit complicated to implement others (e.g: format='amber').


swails:
Why just mol2, pdb, and cif?

hainm:
Because it's easy. :D 

(tried with format='amber' but getting some errors and I do not have time to debug yet).

hainm:
i will push more and let's debug.

swails:
I was just wondering which write methods don't support writing to file objects and require file names

hainm:
just have quick check, only `['RST7', 'AMBER',]` can not write to StringIO.
Seems doable for me.

hainm:
all (except netcdf) formats are supported now.

hainm:
Also update `id_format` for PDBFile so it can take `StringIO`.
This is for `pdb4amber` to check if the format is pdb. (if not `pdb` and if `reduce` is used to add hydrogen, `parmed` will convert whatever (.cif) to pdb for `reduce`).

hainm:
good for review now.

parmed/amber/asciicrd.py
```diff
@@ -87,7 +95,8 @@ def positions(self):
 
     def close(self):
         """ Close the open file handler """
-        self.closed or self._file.close()
+        if self._own_handle:
+            self.closed or self._file.close()
         self.closed = True
```
hainm: so what's your solution here? 

something like this?

```python
     def close(self):
         """ Close the open file handler """
        if self._own_handle:
            self.closed or self._file.close()
            self.closed = True
```

This class is a bit weird to me (vs. others): Related issue: https://github.com/ParmEd/ParmEd/issues/778

parmed/amber/asciicrd.py
```diff
@@ -95,8 +95,7 @@ def positions(self):
 
     def close(self):
         """ Close the open file handler """
-        if self._own_handle:
-            self.closed or self._file.close()
+        self.closed or self._file.close()
         self.closed = True
```
swails: What you had before was fine, just move the `self.closed` inside the if block.

parmed/amber/amberformat.py
```diff
@@ -498,13 +498,15 @@ def rdparm_slow(self, fname):
...
+            own_handle = True
         else:
-            lines = fname.readlines()
+            prm = fname
+            own_handle = True
+            assert hasattr(prm, 'read'), 'must be a file object'
 
         # Open up the file and read the data into memory
-        for line in lines:
+        for line in prm:
```
hainm: iterate file object here to save memory.

parmed/amber/asciicrd.py
```diff
@@ -76,7 +76,8 @@ def __init__(self, fname, natom, hasbox, mode='r', title=None):
         self._full_lines_per_frame = self.natom * 3 // self.CRDS_PER_LINE
         self._nextras = self.natom * 3 - (self._full_lines_per_frame *
                                           self.CRDS_PER_LINE)
-        self.closed = False
+        if self._own_handle:
+            self.closed = False
```
hainm: really need to set this flag here so `StringIO` won't be closed somewhere.

hainm:
uhm, need to debug old parm format parsing.

hainm:
all Linux tests are passed (OSX still not run yet). Good for review.

parmed/amber/amberformat.py
```diff
@@ -490,36 +497,43 @@ def rdparm_slow(self, fname):
         fmtre = re.compile(r'%FORMAT *\((.+)\)')
         version = None
 
+        if isinstance(fname, string_types):
+            prm = genopen(fname, 'r')
+            own_handle = True
+        else:
+            prm = fname
+            own_handle = True
+            assert hasattr(prm, 'read'), 'must be a file object'
```
swails: Only use an `assert` for code that by design should only be hit in the case of a bug in *our* code.  Python has 2 modes of "optimization" when compiling:

1. `-O` == get rid of all asserts
2. `-OO` == get rid of all docstrings

(At least this is what I recall.)  The whole purpose here is to make `assert` less unpleasant to sprinkle around the code, since a simple flag will eliminate any possible performance hit in a production distribution.

This isn't particularly well-publicized though, which is why people tend to abuse it :)

parmed/amber/asciicrd.py
```diff
@@ -61,7 +62,14 @@ def __init__(self, fname, natom, hasbox, mode='r', title=None):
...
             raise ValueError("%s mode must be 'r' or 'w'" % type(self).__name__)
-        self._file = genopen(fname, mode)
+        if isinstance(fname, string_types):
+            self._file = genopen(fname, mode)
+            self._own_handle = True
+        elif hasattr(fname, 'read'):
+            self._file = fname
+            self._own_handle = False
+        else:
+            raise RuntimeError("Don't know how to handle %s" % fname)
```
swails: `RuntimeError` -> `TypeError`, and then say it's an unsupported type in the error message

hainm: updated

parmed/amber/amberformat.py
swails: Can you change the assert to a `raise TypeError('%s must be a file name or file-like object' % prm)`?

swails:
After the last little change I'll be happy to merge this!

parmed/amber/amberformat.py
hainm: updated with a test.

PR #819, opened 2017-01-17
add space_group to cif

hainm:
The code does not work yet. Can you help? :D 

```python
import parmed as pmd

parm = pmd.load_file('SCM_A.pdb')
print(parm.space_group)

parm.save('test.cif', overwrite=True)
print(pmd.load_file('test.cif').space_group)
```

output
```
P 1 21 1
P 1
```

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -1419,6 +1419,11 @@ class CIFFile(object):
             cell.appendAttribute('angle_gamma')
             cell.append(struct.box[:])
             cont.append(cell)
+        # symmetry
+        sym = containers.DataCategory('symmetry')
+        sym.appendAttribute('space_group_name_H-M')
+        sym.append(struct.space_group)
+        cont.append(sym)
         if coordinates is not None:
             coords = np.array(coordinates, copy=False, subok=True)
             try:
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -1422,7 +1422,7 @@ class CIFFile(object):
         # symmetry
         sym = containers.DataCategory('symmetry')
         sym.appendAttribute('space_group_name_H-M')
-        sym.append(struct.space_group)
+        sym.append([struct.space_group])
         cont.append(sym)
         if coordinates is not None:
             coords = np.array(coordinates, copy=False, subok=True)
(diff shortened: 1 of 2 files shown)


swails:
I added some commits (by accident to your PR... not sure how I had permission for that).

The problem is that the CIF parser was looking for the space group in the wrong section :).  Works for me now.

swails:
What the heck *was* this branch?? I am so confused.

hainm:
I think that's the new github feature that allows you to push to my branch.

test/test_parmed_formats.py
```diff
@@ -1581,6 +1581,18 @@ def test_download(self):
...
+    def test_cif_symmetry(self):
+        """ Tests that symmetry is parsed from mmCIF files correctly """
+        self.assertEqual(download_CIF('1aki').space_group, 'P 21 21 21')
+
+    def test_cif_space_group_written_from_structure(self):
+        parm = pmd.load_file(get_fn('SCM_A.pdb'))
+        self.assertEqual(parm.space_group, 'P 1 21 1')
+        written = get_fn('test.cif', written=True)
+        parm.write_cif(written)
+        parm2 = pmd.load_file(written)
```
hainm: I added a test here for space_group (converted from pdb format).

However, got error
```bash
            symmetry_obj = cont.getObj('symmetry')
            if symmetry_obj is not None:
>               struct.space_group = symmetry_obj.getRow(0)[1]
E               IndexError: list index out of range
```

hainm: The output `cif` is not correct (from my code)

```bash
##
_symmetry.space_group_name_H-M  P
```

`space_group` should be `'P 1 21 1'`

swails:
Wherever you created this branch from was *way* behind the master...

hainm:
:D, probably I checked out from my git repo. 
Let me rebase.

parmed/formats/pdb.py
```diff
@@ -1422,7 +1422,7 @@ def write(struct, dest, renumber=True, coordinates=None,
         # symmetry
         sym = containers.DataCategory('symmetry')
         sym.appendAttribute('space_group_name_H-M')
-        sym.append(struct.space_group)
+        sym.append([struct.space_group])
```
hainm: cool. passed.

swails:
Should be fixed.  The problem is that the writer takes an iterable as the container and then writes out elements zipped with their attribute name.  So it was treating the space group string as that iterable and printing only the 1 element in the symmetry group: `P`.

Passing a list whose one element is the space group fixed the problem.

See if this works for you.

hainm:
yeah, that works for me.

PR #825, opened 2017-01-24
add callback function to minimize (so we can visualize the minimization process with nglview)

hainm:
callback


parmed/tools/simulations/sanderapi.py
```diff
@@ -103,7 +103,7 @@ def energy(parm, args, output=sys.stdout):
             output.write('     EHbond   = %20.7f' % e.hbond)
         output.write('\nTOTAL    = %20.7f\n' % e.tot)
 
-def minimize(parm, igb, saltcon, cutoff, tol, maxcyc, disp=True):
+def minimize(parm, igb, saltcon, cutoff, tol, maxcyc, disp=True, callback=None):
```
hainm: may be only need to pass `kwargs` so we can add more in the future?

hainm:
test

hainm:
add another gif here for future reference :D 

pdb4amber_bdna_minimization

hainm:
ParmEd + tleap + nglview

PR #828, opened 2017-01-26
oops, fix syntax error for minimize

hainm:
Can you turn on the minimization test?
or make different test file for this.

It's clearly that this bug is not captured in previous PR #825


hainm:
@swails can you have a quick look and merge? it's a tiny fix and https://github.com/Amber-MD/pdb4amber/pull/35 is waiting for this fix. thanks.

PR #831, opened 2017-02-07
Add option to name types based on OpenMM Atom.id

ctk3b:
When loading an OpenMM `Topology` into `Structure`, `AtomType` names are
autogenerated and types with equivalent LJ params are combined. This
PR adds an option to instead take the `AtomType`'s name from the optional
Openmm `Atom.id` attribute instead.

This change should not affect any previous behavior.

### DIFF
--- a/test/test_parmed_openmm.py
+++ b/test/test_parmed_openmm.py
@@ -88,7 +88,7 @@ class TestOpenMM(FileIOTestCase):
         self.assertEqual(len(parm.bonds), len(structure.bonds))
 
     def test_load_topology_use_atom_id_as_typename(self):
-        """ Tests loading an OpenMM Topology and System instance """
+        """ Tests loading an OpenMM Topology and using Atom.id to name types """
         import warnings
         ommparm = app.AmberPrmtopFile(get_fn('complex.prmtop'))
         parm = load_file(get_fn('complex.prmtop'))
--- a/test/test_parmed_openmm.py
+++ b/test/test_parmed_openmm.py
@@ -98,8 +98,7 @@ class TestOpenMM(FileIOTestCase, TestCaseRelative, EnergyTestCase):
         for pmd_atom, omm_atom in zip(parm.atoms, ommparm.topology.atoms()):
             omm_atom.id = pmd_atom.type
         structure = openmm.load_topology(ommparm.topology, system,
-                                         xyz=parm.positions,
-                                         use_atom_id_as_typename=True)
+                                         xyz=parm.positions)
 
         self.assertEqual(len(parm.atoms), len(structure.atoms))
         self.assertEqual([a.type for a in parm.atoms],
--- a/test/test_parmed_openmm.py
+++ b/test/test_parmed_openmm.py
@@ -1,21 +1,23 @@
 """ Tests some OpenMM-specific functionality """
 from __future__ import print_function, division, absolute_import
 
-import math
-import numpy as np
-from parmed.utils.six.moves import StringIO
-import os
-import parmed as pmd
-from parmed import openmm, load_file, exceptions, ExtraPoint, unit as u
-import unittest
-from utils import (get_fn, mm, app, has_openmm, FileIOTestCase, CPU,
-                   TestCaseRelative)
 from collections import OrderedDict
-
+import math
+import os
+import unittest
 import warnings
 
+import numpy as np
+
+import parmed as pmd
+from parmed.utils.six.moves import StringIO
+from parmed import openmm, load_file, exceptions, ExtraPoint, unit as u
+from utils import (get_fn, mm, app, has_openmm, FileIOTestCase, CPU,
+                   TestCaseRelative, EnergyTestCase)
+
+
 @unittest.skipUnless(has_openmm, "Cannot test without OpenMM")
-class TestOpenMM(FileIOTestCase, TestCaseRelative):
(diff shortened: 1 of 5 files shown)


hainm:
should it have energy test too?

ctk3b:
Good call, @hainm .  I added a test. As a side note, the `_check_energies` function should probably be factored out at some point into one of the testing base classes.

swails:
From an API perspective, what is the use-case for having both options (i.e., `use_atom_id_as_typename` as `True` and `False`)?  What about having the default behavior just be to use the atom ID as the type name if the ID is set (and otherwise fall back on the automatic naming that was originally implemented)?

My reasoning here is that this feature doesn't seem to me to be important enough to have an option, but the behavior you implemented is clearly better.  So why not just use the better behavior when we can?

test/test_parmed_openmm.py
```diff
@@ -87,6 +88,50 @@ def test_load_topology(self):
...
+
+            for term in all_terms:
+                if term not in ene1:
+                    self.assertAlmostEqual(ene2[term], 0)
+                elif term not in ene2:
+                    self.assertAlmostEqual(ene1[term], 0)
+                else:
+                    self.assertRelativeEqual(ene2[term], ene1[term], places=5)
+
+        import warnings
```
swails: `warnings` is already imported at the top of this file

ctk3b:
I'm totally OK with that. I was primarily trying to preserve 100% backwards compatibility since this was kind of a niche use case of the atom.id field. I do not believe that this is standard behavior.

ctk3b:
@swails Let me know what you prefer. Your suggestion is definitely cleaner but could theoretically break something. I do not know if people frequently use `atom.id` for anything other than a simple index.

swails:
I'd say go ahead and make the change. If id is not set, does not exist as an attribute, or is Falsey, fall back on the current scheme. How does that sound?

ctk3b:
I've made the changes with one caveat which is that I don't test for False-y but for non-integer since `atom.id` is just an atom index by default.

swails:
Why is Travis failing??

swails:
It's failing because ambermini conda package seems to be broken.  @jchodera and @hainm -- any ideas what's happening here and why ambermini would be broken?

swails:
See details here: https://travis-ci.org/ParmEd/ParmEd/jobs/199634017#L521

hainm:
:D bug related to `post-link.sh` in conda. I added [amber_registration](https://github.com/Amber-MD/ambertools-conda-build/blob/master/recipe/amber_registration.py) so user can register quickly (and we can keep track of who's using ambertools binary).

This has been working fine but the new conda did not work properly (not sure if this is the intention).
Related issue I opened here: https://github.com/conda/conda/issues/4573

Meanwhile: @ctk3b can change "ambermd" channel to "omnia" channel.

hainm:
uhm I actually don't know why conda install `ambermini` from `ambermd` channel.

It used to install that from `omnia`: https://api.travis-ci.org/jobs/196746959/log.txt?deansi=true

(I changed nothing).

swails:
> uhm I actually don't know why conda install ambermini from ambermd channel.

Because the ambermd channel is in my default channel list because I need pysander.  And ambermini from ambermd is assigned a higher version number.

@ctk3b -- the best thing to do here I think is pin the ambermini version we are using (at least for now) to 16.16.  I'll submit a PR to your branch for this.

hainm:
> And ambermini from ambermd is assigned a higher version number.

I see. Just realize that I recently upload `ambermini` to ambermd channel (after keeping and testing in my `hainm` channel).

swails:
Did you coordinate with John on this?  This is a highly unexpected source of confusion to utilize the exact same package name.  Ideally I'm sure he'd like to defer maintenance to Amber for this package, but it's probably best not to unleash this as a surprise.  (Of course there's a good chance it was just surprising to me).

swails:
I submitted a PR ctk3b/ParmEd#1 for this, so when @ctk3b merges it, the build failures should be fixed and I can merge this.

hainm:
> Did you coordinate with John on this? 

Yes, @jchodera visited us and discussed about `ambermini`. We keep the same programs in `ambermini` from `omia` and from `ambermd` to make sure there's no confusion. We also added `ambermini` target in AmberTools/src/Makefile.

I only added the registration (I could remove it if needed).

I think @dacase prefers to have AMBER to maintain both `ambernimi` and `ambertools`. 

Just FYI, how to build in ambermini and ambertools:

```bash
cd AmberTools/src
export AMBER_BUILD_TASK=ambermini
conda build conda-recipe
export AMBER_BUILD_TASK=ambertools
conda build conda-recipe
# and so on
```

swails:
Have you tested `amber_registration` yet?  You never specified a URL to POST to, so I doubt it would work even if you uncommented it... :)

Also, do you know how the form data is posted?  I presume it's in the body (rather than as query parameters).

swails:
> Yes, @jchodera visited us and discussed about ambermini.

Excellent!  Surprising only to me is the way it should be ;).  Nice job with the conda stuff, @hainm.

hainm:
Per registration stuff, I will reply to you via email. :D

ctk3b:
All good! Thanks for fixing this.

swails:
Thanks!

PR #834, opened 2017-02-17
Fix unicode encode/decoding in .top comment

ctk3b:
Users on windows machines are unable to write `.top` files due to the unicode handling in the descriptive header:

```python
SyntaxError: (unicode error) 'unicodeescape' codec can't decode bytes in position 2-3: truncated \UXXXXXXXX escape
```

The following appears to fix it and should preserve #729 which wasn't completely correctly fixed (#730 would end up writing `; \b'.....'` to the `.top`)


ctk3b:
Looks like one of the travis builds stalled.

swails:
Restarted.

swails:
Thanks!

mrshirts:
Unfortunately, this seems to have reintroduced https://github.com/ParmEd/ParmEd/issues/729 again; if there is a newline in the command, a newline is inserted in the topology file w/o a #.

mrshirts:
Or, none of this code appears to be in the omnia-installed Parmed 2.7.3, at least (neither the old version, or the new version).  Works when I manually edit the file in the egg . . .

ctk3b:
That seems peculiar since the test explicitly checks for new line arguments getting a proper comment `;` inserted...

ctk3b:
When you "manually edit the file", are you saying you added the changes from this patch which aren't in the omnia release?

mrshirts:
Yes.  There could be something odd screwing things up in my install process, of course.

swails:
Just for clarification, my current understanding is that the issues you're reporting here are *not* in the latest version of the code here, correct?

If the code here is not in the ParmEd egg file (that is, the egg file has code that is older than you expect), then there are 2 likely explanations:

1. These fixes were committed since the last time I cut a release
1. An older version of ParmEd was installed somehow

Number 2 can happen, for instance, if you installed AmberTools (which comes bundled with an older version of ParmEd).

Of course there are other possible explanations, but those are the two that have bitten me before.

mrshirts:
It may be 
> An older version of ParmEd was installed somehow
> Number 2 can happen, for instance, if you installed AmberTools (which comes bundled with an 
> older version of ParmEd).

I'm in the middle of testing some other things (work OK if I download directly from Git - but I will try to circle back and revisit this when I can play around with how things are configured.

hainm:
You should always check the version
```python
python -c "import parmed; print(parmed.__version__)"
```

PR #837, opened 2017-02-20
Convert RB torsions when writing to AMBER

### LINKED ISSUE #714: topology file conversion from Gromacs to AMBER
Hello, 

I have a issue converting the topology from GROMACS to AMBER. I keep running into "GromacsTopologyFile" is not defined error. I used the script given in the documentation for conversion.

from parmed import gromacs, amber, unit as u
gmx_top = GromacsTopologyFile('pmaawaterFE20mer2.top')
gmx_gro = GromacsGroFile.parse('pmaawaterFE20mer2.gro')
gmx_top.box = gmx_gro.box # Needed because .prmtop contains box info
gmx_top.positions = gmx_gro.positions
amb_prm = AmberParm.from_structure(gmx_top)
amb_prm.write("pmaa.top")
amb_inpcrd = amber.AmberAsciiRestart("pmaa.crd", mode="w")
amb_inpcrd.coordinates = gmx_top.coordinates
amb_inpcrd.box = gmx_top.box
amb_inpcrd.close()

Attached is the file I'm trying to convert. 
Thanks.

[files.tar.gz](https://github.com/ParmEd/ParmEd/files/296244/files.tar.gz)

ctk3b:
When writing to AMBER and RB torsions are defined in the structure, convert them to a series of proper dihedrals instead. 

The tests here seem to be ok but I do still fined discrepancies when running energy comparisons on InterMol using [this branch](https://github.com/ctk3b/InterMol/tree/energies) and the same input files. This would suggest that the way the output energies are evaluated differs between ParmEd and InterMol (using OpenMM energy decomposition vs. running sander and parsing that output). 

I have more reason to assume that this is an InterMol problem but I won't have time to dive deeper until tomorrow so I'm just going to leave this up in case someone has any insights.

Resolves #714


parmed/topologyobjects.py
```diff
@@ -2539,34 +2537,41 @@ def from_rbtorsion(cls, rbtorsion):
...
-        f3 = -c3 / 2.0
-        f4 = -c4 / 4.0
+
+        phi = 0 * u.degrees
+        fc0 = 1.0*c0 + 0.5*c2 + 0.3750*c4
+        fc1 = 1.0*c1 + 0.75*c3 + 0.6250*c5
+        fc2 = 0.5*c2 + 0.5*c4
+        fc3 = 0.25*c3 + 0.3125*c5
+        fc4 = 0.125*c4
+        fc5 = 0.0625*c5
```
swails: Are you sure these maths are the same?  I think @maaren recently fixed the f1,f2,f3,f4 assignments and got them agreeing with GROMACS (and I *think* there was a test to verify that DihedralTypeList and RBTorsionType gave the same energies).

If the two forms are equivalent, it's definitely not immediately apparent to me.

ctk3b: I am using the same tests that @maaren implemented. This is the conversion to a dihedral of the form 
```
fc0 + sum_i=1^6 fci (cos(nx-phi)
```
which can be decomposed into proper dihedrals. I will add more tests tomorrow to make sure all the bases are covered and my validation issue is resolved but we have been using this as our canonical dihedral form in InterMol.

More info: https://github.com/shirtsgroup/InterMol/blob/master/intermol/gromacs/gromacs_parser.py#L205-L218

swails: Fair enough :)

parmed/amber/_amberparm.py
```diff
@@ -328,6 +332,22 @@ def from_structure(cls, struct, copy=False):
...
+            proper_types = DihedralTypeList.from_rbtorsion(dihedral.type)
+            for proper_type in proper_types:
+                proper = _copy.copy(dihedral)
+                proper.type = proper_type
+                struct.dihedrals.append(proper)
+                struct.dihedral_types.append(proper.type)
+                proper.type.list = struct.dihedral_types
+
+        del struct.rb_torsions[:]
+        struct.rb_torsion_types.clear()
```
swails: `TrackedList` inherits from `list`, and `list.clear` was added in Python 3.  Remove this line to fix the Python 2 errors. (the `del struct.rb_torsions[:]` should handle the clearing)

parmed/amber/_amberparm.py
```diff
@@ -328,6 +332,22 @@ def from_structure(cls, struct, copy=False):
...
+        if hasattr(struct, 'parameterset'):
+            struct.parameterset.rb_torsion_types.clear()
```
swails: Why is this needed?  I'm a little concerned about what this is doing

parmed/amber/_amberparm.py
```diff
@@ -328,6 +332,22 @@ def from_structure(cls, struct, copy=False):
...
+
+        # Convert all RB torsions to propers and delete the originals.
+        for dihedral in struct.rb_torsions:
+            proper_types = DihedralTypeList.from_rbtorsion(dihedral.type)
+            for proper_type in proper_types:
+                proper = _copy.copy(dihedral)
+                proper.type = proper_type
+                struct.dihedrals.append(proper)
+                struct.dihedral_types.append(proper.type)
+                proper.type.list = struct.dihedral_types
```
swails: You can remove this `list` assignment if you add the calls to `claim()` described below

parmed/amber/_amberparm.py
```diff
@@ -328,6 +332,22 @@ def from_structure(cls, struct, copy=False):
...
+
+        del struct.rb_torsions[:]
+        struct.rb_torsion_types.clear()
+        if hasattr(struct, 'parameterset'):
+            struct.parameterset.rb_torsion_types.clear()
+
```
swails: At the end here, you should add a call to `struct.dihedrals.claim()` and `struct.dihedral_types.claim()` to make sure the list elements know about their containers.

parmed/amber/_amberparm.py
ctk3b: Good point. I had intended to also add the proper types back to the parameter set. Do you think that would be worth doing or just leave the parameter set completely untouched?

ctk3b:
I am still showing discrepancies between some gromacs and amber conversion energies. I'm going to try to add a few more examples as integration tests that I was working on yesterday. I'll ping you when I have more confidence that this is working correctly

ctk3b:
Ok @swails  if all these tests make it, I think we can be reasonably confident about this and I'll just have to figure out what's going on in InterMol.

parmed/amber/_amberparm.py
swails: Just leave parameter set completely untouched.  Only GROMACS has that attribute attached to it or does anything with it.  If the `GromacsTopologyFile` -> `AmberParm` conversion puts one on the resulting `AmberParm`, it can just be deleted entirely (after all -- atom type truncation in the Amber prmtop to 4 characters will render that set useless, anyway).

swails:
Looks good!  I'll merge it if tests pass.

Thanks!!

swails:
Great work!  Thanks!

PR #838, opened 2017-02-21
Fix negative index writing in PDB files

swails:
This commit adds a test to make sure that negative indices are respected
in PDB file writing. In particular, negative residue numbers are used
somewhat frequently (especially in RNA systems), and the current code
just takes the % 10000 (for residues) and % 100000 (for atoms). Which
yields very weird behavior.

Also add a test for this.


swails:
run jenkins

Issue #460, opened 2015-10-21
Support converting between Ryckaerts-Bellemans and proper torsions


swails:
See shirtsgroup/InterMol#262

swails:
Implemented by #845

PR #845, opened 2017-03-05
Fixed remaining issue in converting rb

maaren:
With this patch I am able to reproduce with Intermol rb dihedral energies between gmx and amber. The remaining differences are due to amber only having 4 numbers after the decimal point and conversion to kJ/mol:

Name                         type     input(gromacs)    output (amber)      diff (amber)

methanol:            dihedral         0.01634456        0.01631760       -0.00002696
1-butanol             dihedral        -1.27702858       -1.27695680        0.00007178


parmed/amber/_amberparm.py
```diff
@@ -387,17 +387,17 @@ def from_structure(cls, struct, copy=False):
...
+#            if dt.phi_k == 0 and dt.per == 0:
+#                dt.per = 1.0
+#            elif dt.per == 0:
+#                warn('Periodicity of 0 detected with non-zero force constant. '
+#                     'Changing periodicity to 1 and force constant to 0 to '
+#                     'ensure 1-4 nonbonded pairs are properly identified. This '
+#                     'might cause a shift in the energy, but will leave forces '
+#                     'unaffected', AmberWarning)
+#                dt.phi_k = 0.0
+#                dt.per = 1.0
```
swails: Where is the need to delete this?  The comments here indicate this addressed a real bug, and deleting it will likely reintroduce it.

parmed/topologyobjects.py
```diff
@@ -2555,14 +2555,20 @@ def from_rbtorsion(cls, rbtorsion):
         fc4 = 0.125*c4
         fc5 = 0.0625*c5
 
+        fc0 = (4 * c0 + 4 * c1 + 4 * c3 - c4 + 4 * c5 ) / 8.
+        fc1 = (-8 * c1 - 6 * c3 - 5 * c5) / 8
+        fc2 = (c2 + c4) / 2
+        fc3 = (-4 * c3 - 5 * c5) / 16
+        fc4 =  c4 / 8
+        fc5 = -c5 / 16
```
swails: doesn't this just overwrite lines 2551 to 2556 above?

parmed/amber/_amberparm.py
maaren: Reading the warning, it looked like a hack to me... Also I had to comment this out to get the energies right. The converted RB dihedrals have sometimes periodicity of 0 and a non-zero force constant, eg:

  %FLAG DIHEDRAL_FORCE_CONSTANT
%FORMAT(5E16.8)
  2.25000000E-01 -1.12500000E-01
%FLAG DIHEDRAL_PERIODICITY
%FORMAT(5E16.8)
  3.00000000E+00  0.00000000E+00
%FLAG DIHEDRAL_PHASE
%FORMAT(5E16.8)
  0.00000000E+00  0.00000000E+00

parmed/topologyobjects.py
maaren: I will clean this up

parmed/amber/_amberparm.py
swails: It *is* a hack, but it's a hack that is needed to work around an implicit assumption about the Amber prmtop file encoded within pmemd.

The consequence of eliminating this code is that 1-4 interactions may be computed incorrectly (which is why this code was added in the first place).  We need to find some way to avoid introducing this known bug.  Another option would be to print the constant energy shift as part of the warning so it can be validated by hand.

Note that a constant energy shift makes no difference in any thermodynamical (or dynamical) property, so I'm more comfortable having energies differ and forces remain constant sometimes than getting them wrong (subtly and silently) in pmemd (and pmemd.cuda) later...

maaren: OK, I understand. I will see if I can get some tests working.

swails:
Sorry, missed with my click :frowning_face:.  Fixes #460

parmed/amber/_amberparm.py
mrshirts: (sorry, missed this thread earlier). Ugh, it would be much better not to have any constant energy shift if there is any way of preventing it.  When going between programs, the same structure should have the same energy.  It makes the logic for any check much more complicated energies aren't equal where they should be. And dihedrals are by far the hardest part to debug.

mrshirts:
See comment above - it would be far, far preferable not to have any constant energy shift if there is any way of preventing it. When going between programs, the same structure should have the same energy. It makes the logic for any check much more complicated energies aren't equal where they should be. And dihedrals are by far the hardest part to check manually.   I've just been running 650 molecules through InterMol/ParmEd, and you can't do a post check manually with any effectiveness.

parmed/amber/_amberparm.py
maaren: I agree and I think this pull request fixes that. I have added three examples. The examples compare the forces from OpenMM with gmx. I don't think it is possible to compare directly with real amber in parmed. Please correct me if I am wrong. The three tests I added are two simple ones (HOOH based, one dihedral, with different parameters) and one (methanol based three dihedrals and 1-4 interaction). Current intermol gives now for methanol:

Energy group summary
=======================================================================
                type     input(gromacs)    output (amber)      diff (amber)
-----------------------------------------------------------------------
Not comparable energies: are likely not to be the same
-----------------------------------------------------------------------
       coulomb total        14.13787136       14.13731760       -0.00055376
          coulomb-14        14.13787136       14.13731760       -0.00055376
              proper         0.01634456        0.01631760       -0.00002696
-----------------------------------------------------------------------
Comparable energy terms: these should be very close
-----------------------------------------------------------------------
               angle         0.19632179        0.19622960       -0.00009219
                bond         0.09382390        0.09372160       -0.00010230
              bonded         0.30649024        0.30626880       -0.00022144
            dihedral         0.01634456        0.01631760       -0.00002696
           nonbonded        14.13787136       14.13731760       -0.00055376
           potential        14.44436160       14.44358640       -0.00077520

I think the differences are because amber output only contains 4 numbers after the decimal point. I don't think it is possible to tell intermol to give out put in kcal/mol. I think it would look better if that was the case.

If someone tells me what mol I need to add proof that this works, please let me know.

maaren:
I think I have now fixed all the failures in existing checks

swails:
> I don't think it is possible to compare directly with real amber in parmed. Please correct me if I am wrong.

sander has a Python API that is used in several places in the tests (ParmEd is directly compatible with the pysander API).  So you *can* do energy comparisons directly with Amber.

swails:
> See comment above - it would be far, far preferable not to have any constant energy shift if there is any way of preventing it. When going between programs, the same structure should have the same energy. It makes the logic for any check much more complicated energies aren't equal where they should be. And dihedrals are by far the hardest part to check manually. I've just been running 650 molecules through InterMol/ParmEd, and you can't do a post check manually with any effectiveness.

in pmemd there is no way to avoid a constant shift -- it simply ignores any pn=0 torsion terms.  But it *can* be avoided in sander (and I'll try to change pmemd so that in Amber 18 released in a year or so it will stop ignoring these constant terms).

I agree, of course, that having energies match "exactly" is highly preferable, bordering on being required.  However, from a practical perspective we will need to accept that sander will agree, but pmemd may have constant shifts... :(

swails:
@maaren -- I submitted a PR into your PR with changes that will preserve the pn=0 terms while still avoiding the bug that was fixed by fudging the pn=0 terms to pn=1. (maaren/ParmEd#1)

Once you merge that PR, and the tests pass here, I'll merge this.

maaren:
Lots of errors. Should I revert?

swails:
No, it's just a linting error.  I'll fix it and submit another PR to you.

mrshirts:
Jason, what needs to be done here at this point on our side?  I want to see if there's a way to get this pushed through so Paul can do his checks cleanly.

swails:
One of the tests is failing here:

```
FAIL: Check equal energies for Gromacs -> Amber conversion of Amber FF
  File "/home/travis/build/ParmEd/ParmEd/test/test_format_conversions.py", line 339, in test_unconvertable_rb_torsion
    self.check_energies(top, cong, parm, cona)
  File "/home/travis/build/ParmEd/ParmEd/test/utils.py", line 89, in check_energies
    self.assertRelativeEqual(ene2[term], ene1[term], places=5)
  File "/home/travis/build/ParmEd/ParmEd/test/utils.py", line 68, in assertRelativeEqual
    (val1, val2, 10**-places, ratio))
AssertionError: 1.4346490265 != 1.39173291498 with relative tolerance 1e-05 (1.030836)
```

Sometimes the error triggers in the nonbonded energy (like it did here) or in the total energy.  But that has to do with the fact that dict key ordering can vary from run to run -- the real issue here is that this particular conversion gets the nonbonded terms wrong.

The OpenMM nonbonded energy for the prmtop file matches what Amber (sander API) gives almost exactly.  So using the decomposition there, I get these results:

```python
(Pdb) term
'nonbonded'                                                                                                                         
(Pdb) ene2[term], ene1[term] # ene2 comes from OpenMM on Amber files, ene1 comes from OpenMM on GROMACS files
(1.4346483627190438, 1.3917322512012957)
(Pdb) with sander.setup(get_fn('rb_torsions.prmtop', written=True), parm.coordinates, None, inp): e, f = sander.energy_forces()
(Pdb) e.elec_14 + e.elec + e.vdw_14 + e.vdw
1.434597210967099
(Pdb) e.elec_14, e.elec, e.vdw_14, e.vdw
(-7.086426056847149, 8.564385510466309, -0.043362242652061275, 0.0)
```
The rest of the terms (torsions, bonds, and angles) match out to machine precision.  @maaren -- do you have these detailed contributions for this system from GROMACS?  The first and third energies (in kcal/mol) are the 1-4 electrostatic and 1-4 Lennard-Jones, respectively, for this system in Amber.

Does the GROMACS nonbonded energy match what OpenMM gave (1.3917 kcal/mol)?  If so, how do the nonbonded components match up with what Amber produces?

swails:
Also, why is the Lennard-Jones energy 0?  That seems *very* strange (especially since the 1-4 Lennard-Jones is *not* zero).  Is it because the only 1-5+ interactions involve atoms with epsilons of 0?  That's the only thing that could explain electrostatics being non-zero and vdW being zero.

swails:
Some more investigation: the GROMACS system created here has a CustomNonbondedForce to handle the Lennard-Jones force, so that's directly comparable.  It *also* gets 0, so that's good.

When I zero-out the charges in the GROMACS system, I get a nonbonded energy of -0.08629 kcal/mol, which would be the 1-4 Lennard-Jones terms.  This is ~double what sander gets (-0.04336).

If I zero-out the 1-4 terms (by setting chgscale to almost 0 and setting all 1-4 epsilons to 0), I get 8.5647 kcal/mol from the GROMACS files.  This matches what sander gets quite well.  With this number, the 1-4 energy (-0.08629), and the total (1.3917), I can isolate the 1-4 electrostatics.  These are -7.08671 kcal/mol, which *also* matches sander quite well.

The problem in this particular case is then clearly the 1-4 Lennard-Jones terms, where it seems to be missing a scaling factor of 2 (or 0.5) somewhere.

swails:
OK, I found the issue.  The problem is in the exception determination.  When I print all of the dihedrals in the prmtop generated from Amber, this is what I get:

```python
        <Dihedral [ign]; <Atom HB1 [1]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG21 [6]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=0, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom HB1 [1]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG22 [7]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=0, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom HB1 [1]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG23 [8]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=0, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom HB2 [2]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG21 [6]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=0, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom HB2 [2]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG22 [7]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=0, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom HB2 [2]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG23 [8]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=0, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral; <Atom OG1 [3]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG21 [6]; In ETH 0>; type=<DihedralType; phi_k=-0.000, per=1, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom OG1 [3]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG21 [6]; In ETH 0>; type=<DihedralType; phi_k=0.468, per=3, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral; <Atom OG1 [3]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG22 [7]; In ETH 0>; type=<DihedralType; phi_k=-0.000, per=1, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom OG1 [3]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG22 [7]; In ETH 0>; type=<DihedralType; phi_k=0.468, per=3, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom OG1 [3]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG23 [8]; In ETH 0>; type=<DihedralType; phi_k=0.239, per=0, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral; <Atom OG1 [3]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG23 [8]; In ETH 0>; type=<DihedralType; phi_k=-0.598, per=1, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral [ign]; <Atom OG1 [3]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG23 [8]; In ETH 0>; type=<DihedralType; phi_k=1.066, per=3, phase=0.000,  scee=2.000, scnb=2.000>>
        <Dihedral; <Atom HB1 [1]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom OG1 [3]; In ETH 0>--<Atom HG1 [4]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
        <Dihedral; <Atom HB2 [2]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom OG1 [3]; In ETH 0>--<Atom HG1 [4]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
        <Dihedral; <Atom HB1 [1]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG21 [6]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
        <Dihedral; <Atom HB1 [1]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG22 [7]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
        <Dihedral; <Atom HB1 [1]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG23 [8]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
        <Dihedral; <Atom HB2 [2]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG21 [6]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
        <Dihedral; <Atom HB2 [2]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG22 [7]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
        <Dihedral; <Atom HB2 [2]; In ETH 0>--<Atom CB [0]; In ETH 0>--<Atom CG2 [5]; In ETH 0>--<Atom HG23 [8]; In ETH 0>; type=<DihedralType; phi_k=0.000, per=1, phase=0.000,  scee=2.000, scnb=10000000000.000>>
])
```

Most of the dihedrals that are not ignored in the nonbonded loop (i.e., the ends of the dihedrals are part of the 1-4 pairlist) have an SCNB set very large, which is done because at some point ParmEd thought those should have been excluded from the 1-4 pairlist.

I think what that means is that when I reshuffled the torsion list (to make sure that pmemd wouldn't throw out needed exception parameters when it threw out N=0, torsions) that happened after the `ignore` flag on that dihedral was set.

I think I at least know where to look now, so hopefully I can get this fixed soon.

maaren:
@swails do you still want me to answer the questions you asked me?

swails:
No, it's pretty clear what the problem is.  There's a change I know I can make to fix it, but it's an accidental fix.  I need to figure out exactly why these scaling terms are being set incorrectly because it smells like a subtle lurking bug.

Thanks.

swails:
Fixed it!  I'll submit a PR to your fork to update this.  After that fix, all of the tests should pass and this can finally be merged.

swails:
OK, the fix is committed and PR raised (maaren/ParmEd#3). Once that gets merged and the tests run here, this is good to go!

maaren:
The checks passed! Thanks @swails!

swails:
Whew! Thanks @maaren and @ctk3b for getting this off the ground! This functionality was sorely needed to support GROMACS->Amber conversions in a meaningful way. Let's merge it.

PR #857, opened 2017-03-23
Bugfix/charmm improper

### LINKED ISSUE #854: Not all impropers written in CHARMM file generation?
When converting some amber files (loaded into amb_structure) to charmm using the command:

```python
 pmd.charmm.CharmmParameterSet.write(
            pmd.charmm.CharmmParameterSet.from_structure(amb_structure),
            top=charmm_output_rtf,
            par=charmm_output_prm)
```
I generated files w/o an error, but when trying to run charmm on the output files, I got the error:

```
<CODES>: No improper parameters for     4 ( CLTU     CCLTU    CDLTU    HALTU   )
<CODES>: No improper parameters for     6 ( CLTU     CCLTU    CDLTU    HALTU   )
<CODES>: A TOTAL OF    2 MISSING PARAMETERS
```

This seems to be a reordering of an existing improper string, so could be an atom ordering issue, but I'm not quite sure.

Molecule is https://en.wikipedia.org/wiki/1,4-Benzoquinone, which I imagine could cause some confusions with impropers.

All input amber and output charmmfiles attached (with .txt appending), along with the stdout for running charmm.

[mobley_3727287.crd.txt](https://github.com/ParmEd/ParmEd/files/853681/mobley_3727287.crd.txt)
[mobley_3727287.frcmod.txt](https://github.com/ParmEd/ParmEd/files/853682/mobley_3727287.frcmod.txt)
[mobley_3727287.inp.txt](https://github.com/ParmEd/ParmEd/files/853680/mobley_3727287.inp.txt)
[mobley_3727287.inpcrd.txt](https://github.com/ParmEd/ParmEd/files/853686/mobley_3727287.inpcrd.txt)
[mobley_3727287.prm.txt](https://github.com/ParmEd/ParmEd/files/853687/mobley_3727287.prm.txt)
[mobley_3727287.prmtop.txt](https://github.com/ParmEd/ParmEd/files/853683/mobley_3727287.prmtop.txt)
[mobley_3727287.psf.txt](https://github.com/ParmEd/ParmEd/files/853684/mobley_3727287.psf.txt)
[mobley_3727287.rtf.txt](https://github.com/ParmEd/ParmEd/files/853685/mobley_3727287.rtf.txt)

[charmm_stdout_mobley_3727287.txt](https://github.com/ParmEd/ParmEd/files/853688/charmm_stdout_mobley_3727287.txt)

swails:
Fix #854 

@mrshirts -- give this a try and see if it fixes the issues you were seeing.

It's almost certainly going to fail some tests, but I want to make sure that it solves the missing parameters you were seeing.


mrshirts:
It does indeed run to completion with a correct dihedral energy with this fix.

mrshirts:
Though I haven't checked that it doesn't screw up other things.

mrshirts:
For what it's worth, this code does appear to be screwing up the amber->amber conversion, in the dihedral terms, for some of the other files I was testing.

swails:
> For what it's worth, this code does appear to be screwing up the amber->amber conversion, in the dihedral terms, for some of the other files I was testing.

Can you explain this a bit more?  All this code does is add some additional keys to the improper type list when creating a parameter set from a topology file.  What exactly does it break?

mrshirts:
At the point where you said "It's almost certainly going to fail some tests, but I want to make sure that it solves the missing parameters you were seeing." It was giving incorrect torsion energies for the amber->charmm conversion. You checked in more code since then, which may have fixed the problem.  I can try again if you'd like.

swails:
> It was giving incorrect torsion energies for the amber->charmm conversion.

Ah, I had misunderstood.  The changes I made were to the output of the Amber frcmod file -- it would not have affected the CHARMM files it generated at all.

I'm personally at a loss here.  I looked at the converted Amber Cornell force field files in the CHARMM toppar directory as inspiration for what to do here.

The only thing I can see that's different between those files and the ones ParmEd generates is that in my stream/prm files the improper terms are started with the token `IMPROPER` while in the Amber files of the toppar directory, it uses `IMPHI`.  Can you try changing the `IMPROPER` to `IMPHI` and see if that makes any difference?

swails:
Nevermind, I found this in `charmm/source/io/parmio.src`:

```Fortran
1321           ELSEIF (WRD == 'IMPR' .OR. WRD=='IMPH') THEN
1322              CURRNT='IMPH'
```

There's clearly no distinction between the two.  It'd be interesting to get the OpenMM energies for both the prmtop and PSF/PRM file that's converted as well as the Amber energy with the prmtop and CHARMM energy with the PSF/PRM.  I've looked through what the conversion code is doing, and I can't see anything that's being done *wrong*.

mrshirts:
Can you clarify a bit?  This would have been the problem before this PR, or
after this PR?

swails:
This PR hasn't changed anything about how impropers are translated from Amber -> CHARMM.  The only thing this PR has done is add a few extra terms to the improper terms representing different permutations of the non-central atoms in the impropers.

The problem is that there are several things that *could* be going wrong here:

1. CHARMM could be using the wrong functional form
2. CHARMM could be using a different definition of the "central" atom in the impropers
3. It could be doing both of the above

I had thought that this had been figured out in past PRs (I had been working with one of Alex MacKerrell's post-docs), so it's a little disheartening that these problems are cropping up again.

mrshirts:
Large apologies for something incorrect I said that probably wasted a fair amount of time. This commit does NOT screw up the energies of the amber ==> CHARMM conversion.  It screws up the energy of the amber ==> amber conversion, in particular the dihedral energy.  In this case, amber ==> amber conversion is done by converting to GROMACS and then back to AMBER.   Before this fix, the amber->amber energy was correct.

Particularly, it seems to be a problem with converting GROMACS to AMBER, as the GROMACS file generated by ParmEd is the same before and after.  

In particular, it appears to be having problems in cases where the in the same gromacs file, the dihedral has multiple entries, each with different periodicities.  

Having some problems identifying the correct before and after files because of the ParmEd changes and what got kept, but I can try to do that if the problem is not clearer after this clarification . . .

swails:
Ah, what's happening here is the Amber -> GROMACS -> Amber round-trip is not resulting in two Amber systems that give the same energy.

I don't see how this is happening.  The test suite [contains several examples of GROMACS -> Amber conversions](https://github.com/ParmEd/ParmEd/blob/master/test/test_format_conversions.py#L20-L352), and the test system there has both improper and proper torsions defined in them.  Likewise there are test conversions going the other way as well, and none of those tests changed in this PR.

Do you see the same issues with the `master` branch?  If the `master` branch works fine, are you printing the parameters inline, or in separate `[ parametertype ]` sections in the GROMACS topology file?  If you're not using the inline option, does switching to the inline option fix things?

Also, I presume you're computing the energy using GROMACS itself and not OpenMM, correct?  It would help to have input files and scripts to do the calculation in GROMACS (and it sounds like this needs to turn into another test case).

mrshirts:
> Ah, what's happening here is the Amber -> GROMACS -> Amber round-trip is not resulting in two Amber systems that give the same energy.

Yes.

> I don't see how this is happening. The test suite contains several examples of GROMACS -> Amber conversions, and the test system there has both improper and proper torsions defined in them. Likewise there are test conversions going the other way as well, and none of those tests changed in this PR.

I'll attach GROMACS files here that cause the issues:

[mobley_6733657_from_amber.gro.txt](https://github.com/ParmEd/ParmEd/files/877673/mobley_6733657_from_amber.gro.txt)
[mobley_6733657_from_amber.top.txt](https://github.com/ParmEd/ParmEd/files/877674/mobley_6733657_from_amber.top.txt)

As well as the starting and ending AMBER files:

Starting:
[mobley_6733657.frcmod.txt](https://github.com/ParmEd/ParmEd/files/877688/mobley_6733657.frcmod.txt)
[mobley_6733657.inpcrd.txt](https://github.com/ParmEd/ParmEd/files/877689/mobley_6733657.inpcrd.txt)
[mobley_6733657.prmtop.txt]

Ending:
(https://github.com/ParmEd/ParmEd/files/877690/mobley_6733657.prmtop.txt)
[mobley_6733657_converted.rst7.txt](https://github.com/ParmEd/ParmEd/files/877691/mobley_6733657_converted.rst7.txt)
[mobley_6733657_converted.prmtop.txt](https://github.com/ParmEd/ParmEd/files/877692/mobley_6733657_converted.prmtop.txt)

> Do you see the same issues with the master branch? If the master branch works fine, are you printing the parameters inline, or in separate [ parametertype ] sections in the GROMACS topology file? If you're not using the inline option, does switching to the inline option fix things?

The master does indeed cause this right now.   You can take a look at the GROMACS file that is causing the problems.

> Also, I presume you're computing the energy using GROMACS itself and not OpenMM, correct? It would help to have input files and scripts to do the calculation in GROMACS (and it sounds like this needs to turn into another test case).

The problem is AMBER energies computer by sander, when the AMBER file is converted from GROMACS.  Gromacs energies are fine.  Energies computed for Amber => GROMACS is correct (i.e. the AMBER energies match the GROMACS energies). GROMACS => AMBER gives an amber file that does not match in energies either the GROMACS file, or the original AMBER file.

swails:
If the master branch has the same issue, how about I merge this PR and move the issue you're reporting to a separate bug fix?  The way I see it, this PR fixes the bug it set out to fix.

mrshirts:
That's fine.  The version I was using before (which I believe was the latest conda release) was not giving the bug, but the master I was, so it's fine to merge this.

PR #864, opened 2017-04-14
Add unit(ed) properties to topology objects

### LINKED ISSUE #492: Add units to topologyobjects
ParmEd has, for some time, understood units and been able to deal with dimensional analysis and automatic unit conversion.  However, all of the parameter types in `topologyobjects.py` do not expose their parameters with units (they accept parameters with units and throw errors for incompatible units).  This has led to some confusion.

I would like to investigate the implications of, and consider implementing, the infrastructure to always supply units to the parameter types.  Possible downsides are severe API breakage and possibly severe performance degradation (as the dimensional analysis can be slow).

swails:
Fixes #492 
united-training-video

### DIFF
--- a/devtools/ci/jenkins/install.sh
+++ b/devtools/ci/jenkins/install.sh
@@ -61,4 +61,3 @@ fi
 
 # Get rid of our environment
 source deactivate
-conda uninstall -yn ${CONDAENV} --all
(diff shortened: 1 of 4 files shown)


swails:
This implements idea 2 from #492.  @jchodera -- I'm going to merge this this evening unless I get comments about it.

PR #867, opened 2017-04-28
Add improper_dihedrals_ordering parameter for Amber FF conversion

rafwiewiora:
Last thing necessary for the Amber FF conversion - this adds the `ordering='amber'` attribute (and @ChayaSt @jchodera will need to use it to add `ordering='charmm'` to their CHARMM ffxml) to the `PeriodicTorsionForce` element.

cc https://github.com/pandegroup/openmm/pull/1560

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -345,8 +345,8 @@ class OpenMMParameterSet(ParameterSet):
         # Structure has separate dihedral entries for each torsion
         if improper_dihedrals_ordering == 'default':
             dest.write(' <PeriodicTorsionForce>\n')
-        elif improper_dihedrals_ordering == 'amber':
-            dest.write(' <PeriodicTorsionForce ordering="amber">\n')
+        else:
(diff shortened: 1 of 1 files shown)


rafwiewiora:
@ChayaSt @jchodera @peastman can you confirm that this is how you will be writing out your PeriodicTorsionForce impropers, not with some different kind of code given the weirdness in CHARMM? Then I will add an `ordering='charmm'` option here too. See https://github.com/pandegroup/openmm/pull/1560#issuecomment-298128151 for rationale.

swails:
run jenkins

peastman:
Looks fine.  You could even simplify it to just

```Python
dest.write(' <PeriodicTorsionForce ordering="%s">\n' % improper_dihedrals_ordering)
```

That way we won't need to update those lines if we add other orderings in the future.

rafwiewiora:
> That way we won't need to update those lines if we add other orderings in the future.

Good point - made this change, but left the `default` case still not writing anything.

@swails this should be good to merge!

swails:
run jenkins

swails:
Is this good to merge then?

rafwiewiora:
Go for it!

PR #873, opened 2017-05-10
alternate conformation chains: 	Extend to ABCDEFGHIJKLMNOPQRSTUVWXYZ

hainm:
cc @dacase


parmed/amber/_amberparm.py
```diff
@@ -2006,7 +2006,7 @@ def _get_atom_collection_for_alternate_labels(self):
 
     def _label_alternates(self):
         atom_collection = self._get_atom_collection_for_alternate_labels()
-        possible_labels = list('ABCDEF')
+        possible_labels = list('ABCDEFGHIJKLMNOPQRSTUVWXYZ')
```
swails: Are you sure you spelled that right?

:)

hainm: not sure. Just copy and paste from StackOverflow. :D

hainm: oh, I think yes. Those string matches to the one in python website

https://docs.python.org/3.1/library/string.html

hainm:
```bash
string.ascii_uppercase
The uppercase letters 'ABCDEFGHIJKLMNOPQRSTUVWXYZ'. This value is not locale-
dependent and will not change.
```

swails: Sorry, stupid joke :(

hainm: ha, this is why I prefer talking face-to-face. :P

PR #878, opened 2017-05-24
Bug fixes

### LINKED ISSUE #869: Selection masks containing the underscore symbol "_"
Drops an error whenever trying to select amber atom type containing "_", eg. scpe water oxygen atomtype "OW_s".

### LINKED ISSUE #870: Parmed leaves behind extra CMAP grids
Version 2.7.3+135.g57c6b8b

Related to Amber-MD/cpptraj#494

Using attached topology file and the following input:
[step5_charmm2amber.parm7.gz](https://github.com/ParmEd/ParmEd/files/979977/step5_charmm2amber.parm7.gz)

```
parm step5_charmm2amber.parm7
strip !:277
outparm parmed.ligand.parm7
go
```
All CMAP information should be removed, but parmed leaves behind two errant CHARMM_CMAP_PARAMETER_XX flags (2 and 4). Also leaves behind the CHARMM_CMAP_INDEX section. Flags CHARMM_CMAP_COUNT, CHARMM_CMAP_RESOLUTION, and other grids are (correctly) removed.

swails:
Fixes #869
Fixes #870

Adds tests for them as well

bugs

### DIFF
--- a/parmed/amber/_chamberparm.py
+++ b/parmed/amber/_chamberparm.py
@@ -502,8 +502,10 @@ class ChamberParm(AmberParm):
             flags_to_delete = [flag for flag in self.flag_list if flag.startswith('CHARMM_CMAP')]
             for flag in flags_to_delete:
                 self.delete_flag(flag)
-            del self.pointers['CMAP']
-            del self.pointers['CMAP_TYPES']
+            if 'CMAP' in self.pointers:
+                del self.pointers['CMAP']
+            if 'CMAP_TYPES' in self.pointers:
+                del self.pointers['CMAP_TYPES']
             return
         # Time to transfer our CMAP types
         data = self.parm_data
--- a/parmed/amber/_chamberparm.py
+++ b/parmed/amber/_chamberparm.py
@@ -20,18 +20,19 @@ along with this program; if not, write to the Free Software
 Foundation, Inc., 59 Temple Place - Suite 330,
 Boston, MA 02111-1307, USA.
 """
-from __future__ import division, print_function
+from __future__ import absolute_import, division, print_function
 
-from parmed.amber._amberparm import AmberParm
-from parmed.constants import (NTYPES, NATYP, IFBOX, TINY, NATOM, SMALL,
-                              DEG_TO_RAD, RAD_TO_DEG)
-from parmed.exceptions import AmberError, AmberWarning
-from parmed.topologyobjects import (UreyBradley, Improper, Cmap, BondType,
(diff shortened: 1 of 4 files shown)


parmed/amber/_chamberparm.py
```diff
@@ -502,8 +502,10 @@ def _xfer_cmap_properties(self):
             flags_to_delete = [flag for flag in self.flag_list if flag.startswith('CHARMM_CMAP')]
             for flag in flags_to_delete:
                 self.delete_flag(flag)
-            del self.pointers['CMAP']
-            del self.pointers['CMAP_TYPES']
+            if 'CMAP' in self.pointers:
```
hainm: can we replace 

```python
+            if 'CMAP' in self.pointers:
                  del self.pointers['CMAP']
```

```python
self.pointers.pop('CMAP_TYPES', None)
```
so on for others.

swails: I actually think `del self.pointers['CMAP']` is marginally clearer about the intent -- `pop` is something you use when you want to use an item after removing it from the collection.  `del` is what you use when you just want it removed.

Did you also know that `pop` is marginally slower than the other approach, even though it's only 1 line? :smiley:

```python
In [9]: S = 'Tribe'

In [10]: def popitem():
   ....:     d = {S: 1}
   ....:     d.pop(S, None)
   ....:     

In [11]: def delitem():
   ....:     d = {S: 1}
   ....:     if S in d:
   ....:         del d[S]
   ....:         

In [12]: %timeit popitem()
The slowest run took 14.37 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 282 ns per loop

In [13]: %timeit delitem()
The slowest run took 16.98 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 239 ns per loop

In [14]: def popitem():
   ....:     d = {}
   ....:     d.pop(S, None)
   ....:     

In [15]: def delitem():
   ....:     d = {}
   ....:     if S in d:
   ....:         del d[S]
   ....:         

In [16]: %timeit popitem()
The slowest run took 13.82 times longer than the fastest. This could mean that an intermediate result is being cached 
1000000 loops, best of 3: 224 ns per loop

In [17]: %timeit delitem()
The slowest run took 30.78 times longer than the fastest. This could mean that an intermediate result is being cached 
10000000 loops, best of 3: 163 ns per loop
```

hainm: cool. thanks for your info. ;)

PR #885, opened 2017-07-22
Update test file that was changed unexpectedly.

swails:


hainm:
may be you can turn on nightly build on travis?

swails:
meh, not much value I think.  This issue was not hard to track down and is one of the first examples that would have been caught by nightly testing.

If it becomes more common (or I lose trust in some upstream dependencies), I may change my mind.

PR #891, opened 2017-09-03
Improved CHARMM patch reading and support for OpenMM ffxml patch writing

jchodera:
This PR improves support for reading CHARMM patches and adds support for writing them in [OpenMM `ffxml` format](http://docs.openmm.org/development/userguide/application.html#patches).

Patches are stored in the `ParameterSet.patches` field as an `OrderedDict` using an improved version of the `PatchTemplate` structure.

The `ResidueTemplate.apply_patch()` convenience function can be used to apply a `PatchTemplate` to generate a modified residue template. An exception is thrown unless
* all atoms to be changed or deleted are present
* all bonds to be added involve extant atoms
* the resulting net charge of the `ResidueTemplate` is integral (to a specified precision)
* the bond graph has exactly one connected component

The last criteria introduces an optional dependency on `NetworkX`, which is handled the same way `Pandas` is handled. A `ResidueTemplate.to_networkx()` convenience function is provided for generating bond graphs.

When writing OpenMM `ffxml` parameter sets, single-residue patches are automatically tested against residues for compatibility using `ResidueTemplate.apply_patch()`. Patches that involve more than one residue are not currently handled, but support will be added in a future PR.

Future improvements (in a separate PR):
* Add support for multi-residue patches 
* Refine logic to `_write_omm_patches` that distinguishes patch atoms that are added vs modified and bonds that are deleted; currently, only one compatible `patch x residue` combination is used to determine this, but no guarantee is made this will work for every `patch x residue` combination.
* `PatchTemplate` may need to implement its own `__copy__` to copy additional information it stores.

Tests have been added for any new functionality added.

Feedback/modifications/suggestions welcomed!

cc: https://github.com/choderalab/openmm-forcefields/issues/22

### DIFF
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -388,9 +388,15 @@ class ResidueTemplate(object):
         # Create a copy
         # TODO: Once ResidueTemplate.from_residue() actually copies all info, use that instead?
         residue = _copy.copy(self)
-        # Process patch
+        # Delete atoms
         for atom_name in patch.delete_atoms:
             residue.delete_atom(atom_name)
+        # Replace atoms
+        for atom in patch.atoms:
+            if atom.name in residue:
+                # Overwrite type and charge
+                residue[atom.name].type = atom.type
+                residue[atom.name].charge = atom.charge
         # Delete impropers
         for impr in patch.delete_impropers:
             residue.remove(impr)
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -11,6 +11,7 @@ from parmed.residue import AminoAcidResidue, RNAResidue, DNAResidue
 from parmed.structure import Structure
 from parmed.topologyobjects import Atom, Bond, AtomList, TrackedList
 from parmed.utils.six import iteritems
+from parmed.exceptions import IncompatiblePatchError
 import warnings
 
 __all__ = ['PROTEIN', 'NUCLEIC', 'SOLVENT', 'UNKNOWN', 'ResidueTemplate',
@@ -143,7 +144,7 @@ class ResidueTemplate(object):
             atom_name = atom.name
 
         if atom_name not in self._map:
-            raise RuntimeError("Could not find atom '%s' in ResidueTemplate, which contains atoms: %s" % (atom_name, list(self._map.keys())))
+            raise KeyError("Could not find atom '%s' in ResidueTemplate, which contains atoms: %s" % (atom_name, list(self._map.keys())))
 
         atom = self._map[atom_name]
 
@@ -412,7 +413,10 @@ class ResidueTemplate(object):
         residue = _copy.copy(self)
         # Delete atoms
         for atom_name in patch.delete_atoms:
-            residue.delete_atom(atom_name)
+            try:
+                residue.delete_atom(atom_name)
+            except KeyError as e:
+                raise IncompatiblePatchError(str(e))
         # Add or replace atoms
         for atom in patch.atoms:
             if atom.name in residue:
@@ -432,25 +436,25 @@ class ResidueTemplate(object):
                         residue.tail = None
                 # Add bond
                 residue.add_bond(atom1_name, atom2_name, order)
-            except:
-                raise Exception('Bond %s-%s could not be added to patched residue: atoms are %s' % (atom1_name, atom2_name, list(residue._map.keys())))
+            except IndexError as e:
+                raise IncompatiblePatchError('Bond %s-%s could not be added to patched residue: atoms are %s' % (atom1_name, atom2_name, list(residue._map.keys())))
         # Delete impropers
         for impr in patch.delete_impropers:
             try:
                 residue.remove(impr)
-            except:
-                raise Exception('Improper %s was not found in residue to be patched.' % impr)
+            except ValueError as e:
+                raise IncompatiblePatchError('Improper %s was not found in residue to be patched.' % impr)
         # Check that the net charge is integral.
         net_charge = residue.net_charge
         is_integral = (round(net_charge, precision) - round(net_charge)) == 0.0
         if not is_integral:
-            raise Exception('Patch is not compatible with residue due to non-integral charge (charge was %f).' % net_charge)
+            raise IncompatiblePatchError('Patch is not compatible with residue due to non-integral charge (charge was %f).' % net_charge)
         # Ensure residue is connected
         import networkx as nx
         G = residue.to_networkx()
         if not nx.is_connected(G):
-            print([ c for c in nx.connected_components(G) ])
-            raise Exception('Patched residue bond graph is not a connected graph.')
+            components = [ c for c in nx.connected_components(G) ]
+            raise IncompatiblePatchError('Patched residue bond graph is not a connected graph: %s' % str(components))
 
         return residue
 
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -143,7 +143,7 @@ class ResidueTemplate(object):
             atom_name = atom.name
 
         if atom_name not in self._map:
-            raise RuntimeError("Could not find atom '%s' in ResidueTemplate, which contains atoms: %s" % (atom_name, self._map.keys()))
+            raise RuntimeError("Could not find atom '%s' in ResidueTemplate, which contains atoms: %s" % (atom_name, list(self._map.keys())))
 
         atom = self._map[atom_name]
 
@@ -257,7 +257,6 @@ class ResidueTemplate(object):
                         inst.connections.append(inst.atoms[idx])
                 else:
                     inst.add_bond(i1, i2)
-        # TODO: Does this method neglect to copy other information, like head, tail, patches, and impropers?
         return inst
 
     @property
@@ -434,7 +433,7 @@ class ResidueTemplate(object):
                 # Add bond
                 residue.add_bond(atom1_name, atom2_name, order)
             except:
-                raise Exception('Bond %s-%s could not be added to patched residue: %s' % (atom1_name, atom2_name))
+                raise Exception('Bond %s-%s could not be added to patched residue: atoms are %s' % (atom1_name, atom2_name, list(residue._map.keys())))
         # Delete impropers
         for impr in patch.delete_impropers:
             try:
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -333,6 +333,16 @@ class ResidueTemplate(object):
 
         return self
 
+    def apply_patch(self, patch):
+        """
+        Apply the specified PatchTemplate to the ResidueTemplate.
+
+        TODO:
+        * How do we handle patches that involve more than one residue?
+        """
+        # TODO
+        raise Exception('Not implemented.')
+
     def to_dataframe(self):
         """ Create a pandas dataframe from the atom information
 
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -224,6 +224,10 @@ class ResidueTemplate(object):
             self._crd = np.array([[a.xx, a.xy, a.xz] for a in self])
         return self._crd
 
+    @property
+    def net_charge(self):
+        return sum([a.charge for a in self])
+
     # Make ResidueTemplate look like a container of atoms, also indexable by the
     # atom name
     def __len__(self):
@@ -311,7 +315,7 @@ class ResidueTemplate(object):
         """
         if not self.atoms:
             raise ValueError('Cannot fix charges on an empty residue')
-        net_charge = sum(a.charge for a in self.atoms)
+        net_charge = self.net_charge
         if to is None:
             to = round(net_charge)
         else:
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -441,7 +441,7 @@ class ResidueTemplate(object):
         # Delete impropers
         for impr in patch.delete_impropers:
             try:
-                residue.remove(impr)
+                residue._impr.remove(impr)
             except ValueError as e:
                 raise IncompatiblePatchError('Improper %s was not found in residue to be patched.' % impr)
         # Check that the net charge is integral.
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -378,11 +378,34 @@ class ResidueTemplate(object):
 
         return self
 
-    def apply_patch(self, patch):
+    def apply_patch(self, patch, precision=4):
         """
         Apply the specified PatchTemplate to the ResidueTemplate.
 
         This only handles patches that affect a single residue.
+
+        An exception is thrown if patch is incompatible because
+        * The patch specifies that an atom is to be deleted that doesn't exist in the residue
+        * A bond specified as being added in the patch does not have both atom names present after adding/deleting atoms from the patch
+        * The new net charge is not integral to the specified precision
+
+        Parameters
+        ----------
+
+        patch : PatchTemplate
+            The patch to apply to this residue
+
+        precision : int, optional
+            Each valid patch should be produce a net charge that is integral to
+            this many decimal places.
+            Default is 4
+
+        Returns
+        -------
+
+        residue : ResidueTemplate
+            A new ResidueTemplate corresponding to the patched residue is returned.
+            The original remains unmodified.
 
         """
         # Create a copy
@@ -391,7 +414,7 @@ class ResidueTemplate(object):
         # Delete atoms
         for atom_name in patch.delete_atoms:
             residue.delete_atom(atom_name)
-        # Replace atoms
+        # Add or replace atoms
         for atom in patch.atoms:
             if atom.name in residue:
                 # Overwrite type and charge
@@ -401,10 +424,21 @@ class ResidueTemplate(object):
                 residue.add_atom(Atom(name=atom.name, type=atom.type, charge=atom.charge))
         # Add bonds
         for bond in patch.bonds:
-            residue.add_bond(bond.atom1.name, bond.atom2.name)
+            try:
+                residue.add_bond(bond.atom1.name, bond.atom2.name)
+            except:
+                raise Exception('Bond %s was not found in residue to be patched.' % bond)
         # Delete impropers
         for impr in patch.delete_impropers:
-            residue.remove(impr)
+            try:
+                residue.remove(impr)
+            except:
+                raise Exception('Improper %s was not found in residue to be patched.' % impr)
+        # Check that the net charge is integral.
+        net_charge = residue.net_charge
+        is_integral = (round(net_charge, precision) - round(net_charge)) == 0.0
+        if not is_integral:
+            raise Exception('Patch is not compatible with residue due to non-integral charge (charge was %f).' % net_charge)
 
         return residue
 
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -439,6 +439,11 @@ class ResidueTemplate(object):
         is_integral = (round(net_charge, precision) - round(net_charge)) == 0.0
         if not is_integral:
             raise Exception('Patch is not compatible with residue due to non-integral charge (charge was %f).' % net_charge)
+        # Ensure residue is connected
+        import networkx as nx
+        G = residue.to_networkx()
+        if not nx.is_connected(G):
+            raise Exception('Patched residue bond graph is not a connected graph.')
 
         return residue
 
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -442,6 +442,23 @@ class ResidueTemplate(object):
 
         return residue
 
+    def to_networkx(self):
+        """ Create a NetworkX graph of atoms and bonds
+
+        Returns
+        -------
+        G : :class:`networkx.Graph`
+            A NetworkX Graph representing the molecule
+
+        """
+        import networkx
+        G = networkx.Graph()
+        for atom in self.atoms:
+            G.add_node(atom.name, charge=atom.charge, type=atom.type)
+        for bond in self.bonds:
+            G.add_edge(bond.atom1.name, bond.atom2.name)
+        return G
+
     def to_dataframe(self):
         """ Create a pandas dataframe from the atom information
 
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -423,11 +423,18 @@ class ResidueTemplate(object):
             else:
                 residue.add_atom(Atom(name=atom.name, type=atom.type, charge=atom.charge))
         # Add bonds
-        for bond in patch.bonds:
+        for (atom1_name, atom2_name, order) in patch.add_bonds:
             try:
-                residue.add_bond(bond.atom1.name, bond.atom2.name)
+                # Remove dangling bonds
+                for name in [atom1_name, atom2_name]:
+                    if name == residue.head:
+                        residue.head = None
+                    if name == residue.tail:
+                        residue.tail = None
+                # Add bond
+                residue.add_bond(atom1_name, atom2_name, order)
             except:
-                raise Exception('Bond %s was not found in residue to be patched.' % bond)
+                raise Exception('Bond %s-%s could not be added to patched residue: %s' % (atom1_name, atom2_name))
         # Delete impropers
         for impr in patch.delete_impropers:
             try:
@@ -443,6 +450,7 @@ class ResidueTemplate(object):
         import networkx as nx
         G = residue.to_networkx()
         if not nx.is_connected(G):
+            print([ c for c in nx.connected_components(G) ])
             raise Exception('Patched residue bond graph is not a connected graph.')
 
         return residue
@@ -660,8 +668,12 @@ class PatchTemplate(ResidueTemplate):
 
     Attributes
     ----------
-    delete : list of str
-        List of atoms that need to be deleted in applying the patch
+    add_bonds : list of (str, str, order)
+        List of bonds that need to be added in applying the patch
+    delete_atoms : list of str
+        List of atom names that need to be deleted in applying the patch
+    delete_impropers : list of tuple of str
+        List of impropers (tuple of atom names) that need to be deleted in applying the patch
 
     See Also
     --------
@@ -675,6 +687,7 @@ class PatchTemplate(ResidueTemplate):
     """
     def __init__(self, name=''):
         super(PatchTemplate, self).__init__(name)
+        self.add_bonds = []
         self.delete_atoms = []
         self.delete_impropers = []
 
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -397,6 +397,9 @@ class ResidueTemplate(object):
                 # Overwrite type and charge
                 residue[atom.name].type = atom.type
                 residue[atom.name].charge = atom.charge
+        # Add bonds
+        for bond in patch.bonds:
+            residue.add_bond(bond.atom1.name, bond.atom2.name)
         # Delete impropers
         for impr in patch.delete_impropers:
             residue.remove(impr)
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -540,7 +540,8 @@ class PatchTemplate(ResidueTemplate):
     """
     def __init__(self, name=''):
         super(PatchTemplate, self).__init__(name)
-        self.delete = []
+        self.delete_atoms = []
+        self.delete_impropers = []
 
 # ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
 
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -397,6 +397,8 @@ class ResidueTemplate(object):
                 # Overwrite type and charge
                 residue[atom.name].type = atom.type
                 residue[atom.name].charge = atom.charge
+            else:
+                residue.add_atom(Atom(name=atom.name, type=atom.type, charge=atom.charge))
         # Add bonds
         for bond in patch.bonds:
             residue.add_bond(bond.atom1.name, bond.atom2.name)
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -128,6 +128,51 @@ class ResidueTemplate(object):
         self.atoms.append(atom)
         self._map[atom.name] = atom
 
+    def delete_atom(self, atom):
+        """ Delete an atom from this residue template, along with corresponding bonds.
+
+        Parameters
+        ----------
+        atom : :class:`Atom` or str
+            The atom or atom name to be deleted
+
+        """
+        if type(atom) is str:
+            atom_name = atom
+        else:
+            atom_name = atom.name
+
+        if atom_name not in self._map:
+            raise RuntimeError("Could not find atom '%s' in ResidueTemplate, which contains atoms: %s" % (atom_name, self._map.keys()))
+
+        atom = self._map[atom_name]
+
+        # Adjust head and tail if needed
+        if self.head == atom:
+            self.head = None
+        if self.tail == atom:
+            self.tail = None
+
+        # Remove all bonds involving this atom
+        for bond in list(self.bonds):
+            if (bond.atom1 == atom) or (bond.atom2 == atom):
+                # Remove this bond from all atoms it spans
+                bond.delete()
+                # Delete the bond from the residue
+                self.bonds.remove(bond)
+
+        # Remove all impropers involving this atom
+        for impr in list(self._impr):
+            if atom in impr:
+                self._impr.remove(impr)
+
+        # Disconnect the atom from the residue so that it does not trigger atom.residue.delete_atom(atom)
+        atom.residue = None
+
+        # Remove the atom from the ResidueTemplate
+        del self._map[atom_name]
+        self.atoms.remove(atom)
+
     def add_bond(self, atom1, atom2, order=1.0):
         """ Adds a bond between the two provided atoms in the residue
 
@@ -179,7 +224,6 @@ class ResidueTemplate(object):
         """
         This constructor creates a ResidueTemplate from a particular Residue
         object
-
         Parameters
         ----------
         residue : :class:`Residue`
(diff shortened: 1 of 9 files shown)


jchodera:
Does ParmEd currently include tests that compare CHARMM system energies with OpenMM energies for systems generated via the `simtk.openmm.app.ForceField` route? If not, I can work on adding that in a separate PR if there is another CHARMM-vs-OpenMM `System` test I can model this on.

jchodera:
Tests pass, so this should be ready for feedback/review, @swails.

swails:
run jenkins

swails:
> Does ParmEd currently include tests that compare CHARMM system energies with OpenMM energies for systems generated via the simtk.openmm.app.ForceField route?

No -- it only includes tests for converting between file formats.  Probably between OpenMM <-> CHARMM, but I'm not positive.

parmed/modeller/residue.py
```diff
@@ -329,6 +377,100 @@ def fix_charges(self, to=None, precision=4):
...
+            try:
+                # Remove dangling bonds
+                for name in [atom1_name, atom2_name]:
+                    if name == residue.head:
+                        residue.head = None
+                    if name == residue.tail:
+                        residue.tail = None
+                # Add bond
+                residue.add_bond(atom1_name, atom2_name, order)
+            except:
```
swails: What exception do you expect to catch here?  Catch only that one.

parmed/modeller/residue.py
```diff
@@ -329,6 +377,100 @@ def fix_charges(self, to=None, precision=4):
...
+                raise Exception('Bond %s-%s could not be added to patched residue: atoms are %s' % (atom1_name, atom2_name, list(residue._map.keys())))
```
swails: Create a new exception type for this in `parmed.exceptions` and raise that.  Raising `Exception` requires you to catch it with an `except` that necessarily catches everything, making it too likely that you'll wind up masking a bug.

parmed/modeller/residue.py
```diff
@@ -128,6 +128,51 @@ def add_atom(self, atom):
...
+            The atom or atom name to be deleted
+
+        """
+        if type(atom) is str:
+            atom_name = atom
+        else:
+            atom_name = atom.name
+
+        if atom_name not in self._map:
+            raise RuntimeError("Could not find atom '%s' in ResidueTemplate, which contains atoms: %s" % (atom_name, list(self._map.keys())))
```
swails: `RuntimeError` -> `KeyError`

parmed/modeller/residue.py
```diff
@@ -329,6 +377,100 @@ def fix_charges(self, to=None, precision=4):
...
+                # Add bond
+                residue.add_bond(atom1_name, atom2_name, order)
+            except:
+                raise Exception('Bond %s-%s could not be added to patched residue: atoms are %s' % (atom1_name, atom2_name, list(residue._map.keys())))
+        # Delete impropers
+        for impr in patch.delete_impropers:
+            try:
+                residue.remove(impr)
+            except:
+                raise Exception('Improper %s was not found in residue to be patched.' % impr)
```
swails: `Exception` -> `KeyError`

parmed/modeller/residue.py
```diff
@@ -329,6 +377,100 @@ def fix_charges(self, to=None, precision=4):
...
+        # Check that the net charge is integral.
+        net_charge = residue.net_charge
+        is_integral = (round(net_charge, precision) - round(net_charge)) == 0.0
+        if not is_integral:
+            raise Exception('Patch is not compatible with residue due to non-integral charge (charge was %f).' % net_charge)
```
swails: `Exception` -> `ValueError`

parmed/modeller/residue.py
```diff
@@ -329,6 +377,100 @@ def fix_charges(self, to=None, precision=4):
...
+        # Ensure residue is connected
+        import networkx as nx
+        G = residue.to_networkx()
+        if not nx.is_connected(G):
+            print([ c for c in nx.connected_components(G) ])
```
swails: Either remove the print statement or use the `logging` module and print this at the debug level with some more information.

parmed/modeller/residue.py
```diff
@@ -329,6 +377,100 @@ def fix_charges(self, to=None, precision=4):
...
+            raise Exception('Patched residue bond graph is not a connected graph.')
```
swails: `Exception` -> `parmed.exceptions.MoleculeError`

parmed/openmm/parameters.py
```diff
@@ -200,13 +204,22 @@ def write(self, dest, provenance=None, write_unused=True, separate_ljforce=False
             except KeyError:
                 warnings.warn('Some residue templates are using unavailable '
                               'AtomTypes', ParameterWarning)
+
+        # DEBUG
+        print('Determining valid patch combinations...')
```
swails: Change this to use `logging` instead of `print`

parmed/openmm/parameters.py
```diff
@@ -200,13 +204,22 @@ def write(self, dest, provenance=None, write_unused=True, separate_ljforce=False
...
+        valid_patch_combinations = self._determine_valid_patch_combinations(skip_residues)
+        for patch_name in self.patches:
+            print('%8s : %s' % (patch_name, valid_patch_combinations[patch_name]))
+        print('')
```
swails: `logging.debug` instead of `print`

parmed/openmm/parameters.py
```diff
@@ -397,9 +412,105 @@ def _write_omm_residues(self, dest, skip_residues):
...
+        """
+        # Attempt to patch every residue, recording only valid combinations.
+        valid_patch_combinations = defaultdict(list)
+        for patch in self.patches.values():
+            for residue in self.residues.values():
+                if residue in skip_residues: continue
+                # Attempt to patch the residue.
+                try:
+                    residue.apply_patch(patch)
+                except Exception as e:
```
swails: Catch a specific exception(s) rather than everything.

parmed/openmm/parameters.py
```diff
@@ -397,9 +412,105 @@ def _write_omm_residues(self, dest, skip_residues):
...
+                    # Patching failed; continue to next patch
+                    print('%8s x %8s : %s' % (patch.name, residue.name, str(e)))
```
swails: `print` -> `logging`

parmed/modeller/residue.py
```diff
@@ -128,6 +128,51 @@ def add_atom(self, atom):
...
+        atom = self._map[atom_name]
+
+        # Adjust head and tail if needed
+        if self.head == atom:
+            self.head = None
+        if self.tail == atom:
+            self.tail = None
+
+        # Remove all bonds involving this atom
+        for bond in list(self.bonds):
```
hainm: why using "list(...)" here?

parmed/modeller/residue.py
```diff
@@ -329,6 +377,100 @@ def fix_charges(self, to=None, precision=4):
...
+        # Add or replace atoms
+        for atom in patch.atoms:
+            if atom.name in residue:
+                # Overwrite type and charge
+                residue[atom.name].type = atom.type
+                residue[atom.name].charge = atom.charge
+            else:
+                residue.add_atom(Atom(name=atom.name, type=atom.type, charge=atom.charge))
+        # Add bonds
+        for (atom1_name, atom2_name, order) in patch.add_bonds:
```
hainm: I though "add_bonds" refers to verb rather a noun?

swails: Good point -- maybe change this to `added_bonds`?

Although in this case this patch indicates a verb should take place -- bonds in this list need to be added when the patch is activated.

jchodera: > Although in this case this patch indicates a verb should take place -- bonds in this list need to be added when the patch is activated.

Right. That's why I went with `delete_atoms` and `delete_impropers`. I could change these too, but the change should probably be consistent.

parmed/modeller/residue.py
jchodera: Creating a `list` to iterate over ensures that when I delete one of the elements of `self.bonds`, I don't mess up the iteration if `self.bonds` is a generator.

parmed/modeller/residue.py
```diff
@@ -128,6 +128,51 @@ def add_atom(self, atom):
...
+        # Remove all bonds involving this atom
+        for bond in list(self.bonds):
+            if (bond.atom1 == atom) or (bond.atom2 == atom):
+                # Remove this bond from all atoms it spans
+                bond.delete()
+                # Delete the bond from the residue
+                self.bonds.remove(bond)
+
+        # Remove all impropers involving this atom
+        for impr in list(self._impr):
```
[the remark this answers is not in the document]
jchodera: Same as above---I just didn't want to mess up the iteration over items if I am removing items from the sequence I am iterating over while I'm doing it.

parmed/modeller/residue.py
hainm: In the line 423, we are using `residue.add_atom(...)`
but in the line 425, we do `... in patch.add_bonds`. It's very confusing.

parmed/modeller/residue.py
jchodera: Adding `IncompatiblePatchError`

parmed/modeller/residue.py
jchodera: Incorporating this debug information into the exception.

parmed/openmm/parameters.py
jchodera: Are there examples of logging output elsewhere I could model this on?

parmed/modeller/residue.py
jchodera: > In the line 423, we are using residue.add_atom(...)
but in the line 425, we do ... in patch.add_bonds. It's very confusing.

One is a method, another is a field. `PatchTemplate` has three fields that instruct the patch on what needs to be done:
* `add_atoms`
* `delete_bonds`
* `delete_impropers`

These are three *actions* that the patch must perform. It performs one of these actions using the `ResidueTemplate.add_atom()` method to add an atom. But to add bonds by name, we must use a different scheme because parmed's `ResidueTemplate.bonds` list cannot contain a list of atom names---they must be `Atom` objects. Patches are allowed to specify names of atoms that are not defiend in the patch as part of bonds being added. This is the clearest workaround that has the smallest impact on the rest of the code.

parmed/modeller/residue.py
```diff
@@ -432,25 +436,25 @@ def apply_patch(self, patch, precision=4):
...
+            except IndexError as e:
+                raise IncompatiblePatchError('Bond %s-%s could not be added to patched residue: atoms are %s' % (atom1_name, atom2_name, list(residue._map.keys())))
         # Delete impropers
         for impr in patch.delete_impropers:
             try:
                 residue.remove(impr)
-            except:
-                raise Exception('Improper %s was not found in residue to be patched.' % impr)
+            except ValueError as e:
+                raise IncompatiblePatchError('Improper %s was not found in residue to be patched.' % impr)
```
hainm: I am not why we need to create this. Do we need to catch it elsewhere?

jchodera:
I think I've got everything!

@swails : Do you have a good example of where `logging` is used? I think printing out the valid patch combinations at the end is very useful for debugging, but I pulled this out because I don't know how parmed logs things elsewhere.

> No -- it only includes tests for converting between file formats. Probably between OpenMM <-> CHARMM, but I'm not positive.

Is there some place where energies are compared between CHARMM and OpenMM? I'm not sure how we can verify the file format conversion is accurate otherwise.

hainm:
@jchodera per logging: https://stackoverflow.com/questions/6918493/in-python-why-use-logging-instead-of-print

parmed/modeller/residue.py
jchodera: I was instructed by @swails to create a new exception type that is as narrow as possible for users of this code to be able to catch when the patch they are attempting to apply is not compatible. If successful, the method returns the patched `ResidueTemplate`. If unsuccessful, the method raises `IncompatiblePatchError`. This is easy to catch if one wants to check whether the patch is incompatible, as we do.

jchodera: Specifically, it's caught in `parmed/openmm/parameters.py` in `_determine_valid_patch_combinations()`. Again, @swails preferred we catch *only* the intended exceptions.

jchodera:
> @jchodera per logging: https://stackoverflow.com/questions/6918493/in-python-why-use-logging-instead-of-print

Oh, I'm onboard with using logging---I just couldn't find a good example of how parmed does this so I could be sure to be 100% consistent with how logging is done elsewhere in parmed. There are lots of program-specific issues such as logging levels, filters, etc.

BTW, is `warnings.warn()` supposed to emit a bunch of useless garbage? My terminal is showing a ton of extra rubbish like
```
/Users/choderaj/github/parmed/ParmEd/parmed/openmm/parameters.py:442: IncompatiblePatchWarning:     ASPP x      ZN2 : Patched residue bond graph is not a connected graph: [{'CB'}, {'CG'}, {'HD2', 'OD2'}, {'HB2'}, {'HB1'}, {'OD1'}, {'ZN'}]
  warnings.warn('%8s x %8s : %s' % (patch.name, residue.name, str(e)), IncompatiblePatchWarning)
/Users/choderaj/github/parmed/ParmEd/parmed/openmm/parameters.py:442: IncompatiblePatchWarning:     ASPP x      LIT : Patched residue bond graph is not a connected graph: [{'CB'}, {'CG'}, {'HD2', 'OD2'}, {'HB2'}, {'HB1'}, {'OD1'}, {'LIT'}]
  warnings.warn('%8s x %8s : %s' % (patch.name, residue.name, str(e)), IncompatiblePatchWarning)
/Users/choderaj/github/parmed/ParmEd/parmed/openmm/parameters.py:442: IncompatiblePatchWarning:     ASPP x     TP3M : Patched residue bond graph is not a connected graph: [{'CG'}, {'OH2', 'H2', 'H1'}, {'CB'}, {'HD2', 'OD2'}, {'HB2'}, {'OD1'}, {'HB1'}]
  warnings.warn('%8s x %8s : %s' % (patch.name, residue.name, str(e)), IncompatiblePatchWarning)
```
where the actual warning is much shorter.

swails:
> Oh, I'm onboard with using logging---I just couldn't find a good example of how parmed does this so I could be sure to be 100% consistent with how logging is done elsewhere in parmed.

I don't use it anywhere... yet.  I didn't start commonly using the `logging` module until I started my current job.  But this is what I typically do:

```python
""" My awesome module with lots of cool functions and classes """

import logging

LOGGER = logging.getLogger(__name__)

def one_of_my_functions(input):
    LOGGER.debug('I received %s as input', input)
```

I will create my own PR where all of the current print statements are replaced by this style of logging

swails:
> BTW, is warnings.warn() supposed to emit a bunch of useless garbage? My terminal is showing a ton of extra rubbish like

The warnings setting by default is set to warn every time a ParmedWarning is emitted.  That can be changed so it only emits one, but if these are expected warnings, you should change it from a warning to a debug statement.

You should only use `warnings.warn` when you encounter something that arguably *should* be fixed.

jchodera:
> I will create my own PR where all of the current print statements are replaced by this style of logging

For now, I've just done what you suggest above, and converted the warnings into a `debug` statement. I'd appreciate any insight you have in actually telling `nosetests` to print that debug info, however, since adding `--logging-level=DEBUG` to
```
nosetests -vs --with-timer --timer-ok=5s --timer-warning=12s --timer-filter=warning,error test/test_parmed_openmm.py:TestWriteParameters
```
doesn't seem to do anything.

Otherwise, this should be ready to go, and I can tackle the energy comparison testing in the next PR. Still trying to figure out where you compare OpenMM to CHARMM energies in the code to test you've done the conversions correctly...

jchodera:
OK, all changes have been made and tests pass.

swails:
> I'd appreciate any insight you have in actually telling nosetests to print that debug info,

Try ``--nologcapture`` instead of ``--logging-level``.  The `-s` means *do not* capture stdout.  When `nose` captures output of any kind, it only dumps captured output if the test fails, I believe.  If you want it printed to the screen regardless, then you don't want it captured by ``nose`` at all.

swails:
run jenkins

parmed/modeller/residue.py
swails: My current understanding is that the only way to see if a patch applies (without duplicating logic here) is to try it and catch any exceptions.  It would be useful, then, to catch an exception that indicated *specifically* this purpose, rather than rely on some kind of built-in exception that has other uses throughout Python.

ParmEd is complex (moreso than it needed to be, but every MD program has enough quirks to make simplicity difficult to achieve), and there are a lot of ways of "breaking" things (Python doesn't impose type safety, and ParmEd simply assumes compatible types wherever they're used in data structures that ParmEd populates with various parsers).  So catching general exceptions like `KeyError` and `IndexError` to identify incompatible patches has a non-negligible likelihood of masking unrelated errors.  It's best to fail as quickly as possible to make debugging as easy as possible, and this new exception furthers that goal.

parmed/openmm/parameters.py
```diff
@@ -205,12 +208,10 @@ def write(self, dest, provenance=None, write_unused=True, separate_ljforce=False
...
                               'AtomTypes', ParameterWarning)
 
-        # DEBUG
-        print('Determining valid patch combinations...')
         valid_patch_combinations = self._determine_valid_patch_combinations(skip_residues)
+        LOGGER.debug('Valid patch combinations:')
         for patch_name in self.patches:
-            print('%8s : %s' % (patch_name, valid_patch_combinations[patch_name]))
-        print('')
+            LOGGER.debug('%8s : %s' % (patch_name, str(valid_patch_combinations[patch_name])))
```
swails: logging functions automatically do %-substitution (and this is caught by pylint, which I'd eventually like to use as a linter instead of pyflakes).  So change this line to

`LOGGER.debug('%8s : %s', patch_name, valid_patch_combinations[patch_name])`

`%s` also automatically calls `__str__`, so `str()` is unnecessary in that case.

jchodera:
> Try --nologcapture instead of --logging-level. The -s means do not capture stdout. When nose captures output of any kind, it only dumps captured output if the test fails, I believe. If you want it printed to the screen regardless, then you don't want it captured by nose at all.

I've tried a ton of variations with these and still can't see any debug logging output. If you find a set of command-line arguments that works for `test/test_parmed_openmm.py:TestWriteParameters`, please let me know. Until then, I presume your logging code just sends this information into the void.

parmed/modeller/residue.py
hainm: @ all: I see.

parmed/openmm/parameters.py
jchodera: Done

jchodera:
OK, I've made the few bytes of changes from the last review. 

I think the algorithmic changes are pretty much complete. If you have further minor stylistic changes to request, I imagine it is literally orders of magnitude faster for you to make those changes yourself in a cleanup PR.

swails:
run jenkins

jchodera:
> This branch is out-of-date with the base branch

Oh, shoot, did some commits go into `master` while you were asking me to make changes? Do I have to update from master or rebase now?

swails:
Yea I merged a pr this morning. I updated your branch so when the testing is done again I'll merge :)

swails:
run jenkins

jchodera:
Hold up a moment. I'm adding one more test to make sure the `ffxml` output can be read into OpenMM. I just found an issue with this test, for example:
https://github.com/jchodera/ParmEd/blob/92eb5e53443c1b19237c7b3c7ea751647b04d76f/test/test_parmed_openmm.py#L577
```
        params.write(get_fn('charmm_conv.xml', written=True),
                     provenance=dict(
                         OriginalFile='par_all36_prot.prm & top_all36_prot.rtf',
                         Reference='MacKerrell'
                     )
        )
```
It turns out that unescaped `&` symbols are not valid in XML. It looks like this may occur several places.

jchodera:
It looks like it's just easiest to modify tests that use `params.write()` to write XML files to also try to read these into OpenMM. It means these tests will only run with `openmm` too, but that will be helpful for catching issues.

jchodera:
I've modified the tests in `test_parmed_openmm.py` to actually try reading the `ffxml` files they generate into `simtk.openmm.app.ForceField` if `has_openmm` is `True`. Surprisingly, this revealed that most of the `ffxml` files were not actually valid XML because of the unprotected `&` in the provenance data.

After correcting this, it still looks like unrelated parts of ParmEd or the tests have some unaddressed bugs:
```
Test writing small XML modifications ... ok (0.0028s)

ERROR: Test that no identical residues are written to XML, using the
  File "/Users/choderaj/github/parmed/ParmEd/test/test_parmed_openmm.py", line 650, in test_not_write_residues_with_same_templhash
    forcefield = app.ForceField(ffxml)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 163, in __init__
    self.loadFile(files)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 231, in loadFile
    template.atoms.append(ForceField._TemplateAtomData(atomName, typeName, self._atomTypes[typeName].element, params))
KeyError: 'K+'

ERROR: Test correct support for the override_level attribute of
  File "/Users/choderaj/github/parmed/ParmEd/test/test_parmed_openmm.py", line 672, in test_override_level
    forcefield = app.ForceField(ffxml)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 163, in __init__
    self.loadFile(files)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 231, in loadFile
    template.atoms.append(ForceField._TemplateAtomData(atomName, typeName, self._atomTypes[typeName].element, params))
KeyError: 'K+'

```
@swails : Any idea what's going on here?

jchodera:
@swails : When writing out provenance information from the `provenance` `dict`, should we automatically protect [special XML characters](https://docs.oracle.com/cd/A97336_01/buslog.102/bc4j/developing_bc_projects/obcCustomXml.htm)?

swails:
Yes they should all be escaped. 

Does the xml writer in Python have a method to escape strings written to xml? It does for encoding to HTTP request query parameters, so it might have something.

jchodera:
> Yes they should all be escaped.

OK, I can add this.

> Does the xml writer in Python have a method to escape strings written to xml? It does for encoding to HTTP request query parameters, so it might have something.

Currently, we're not using any kind of XML writer---the `_write_omm_*` methods all [just write text](https://github.com/ParmEd/ParmEd/blob/master/parmed/openmm/parameters.py#L121-L670).

Any thoughts on how to fix the test failures from functions I haven't touched writing incorrect `ffxml` files?
```
Test writing small XML modifications ... ok (0.0028s)

ERROR: Test that no identical residues are written to XML, using the
  File "/Users/choderaj/github/parmed/ParmEd/test/test_parmed_openmm.py", line 650, in test_not_write_residues_with_same_templhash
    forcefield = app.ForceField(ffxml)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 163, in __init__
    self.loadFile(files)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 231, in loadFile
    template.atoms.append(ForceField._TemplateAtomData(atomName, typeName, self._atomTypes[typeName].element, params))
KeyError: 'K+'

ERROR: Test correct support for the override_level attribute of
  File "/Users/choderaj/github/parmed/ParmEd/test/test_parmed_openmm.py", line 672, in test_override_level
    forcefield = app.ForceField(ffxml)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 163, in __init__
    self.loadFile(files)
  File "/Users/choderaj/miniconda3/lib/python3.5/site-packages/simtk/openmm/app/forcefield.py", line 231, in loadFile
    template.atoms.append(ForceField._TemplateAtomData(atomName, typeName, self._atomTypes[typeName].element, params))
KeyError: 'K+'

```

swails:
XML escaping can be done via the stdlib:

```python
>>> from xml.sax.saxutils import escape
>>> escape('&')
'&amp;'
```

swails:
> Any thoughts on how to fix the test failures from functions I haven't touched writing incorrect ffxml files?

Don't try to read those files back in as XML parameter files in OpenMM (I presume you added that part).  They are only residue templates, so there are no atom types loaded (and as a result, the atom type dict is empty and you'll get a `KeyError` when trying to instantiate *any* residue template).

The alternative is to include a force field file with all of the pertinent atom type information.

swails:
> @swails : Any idea what's going on here?

I'm answering stuff out-of-order, but this particular test only deals with residue templates, which is only half of a force field definition.  Since it's missing the other half, there is no expectation that it would "work" as a force field when loaded into OpenMM.

This doesn't necessarily mean there's a bug since it's entirely reasonable to separate residue definitions from force field parameters/atom types in different files.  As a result, ParmEd has the ability to write files with *just* residue templates or *just* parameters.  If you try to read one of the files in OpenMM with *just* residue templates, it can't find the parameters and crashes.

So for the tests, it's important to check that "complete" force fields be written by ParmEd and read by OpenMM, but there's no need to do this for unit tests involving just residue template conversions.

jchodera:
> Don't try to read those files back in as XML parameter files in OpenMM (I presume you added that part). They are only residue templates, so there are no atom types loaded (and as a result, the atom type dict is empty and you'll get a KeyError when trying to instantiate any residue template).

OK, I'll just skip these for now.

> XML escaping can be done via the stdlib:

Is that sufficient? According to the [docs](https://docs.python.org/3/library/xml.sax.utils.html), only `&`, `<`, and `>` are escaped, but XML requires additional characters to be escaped.

The docs suggest the code is capable of this, but we have to do more work:

> Escape '&', '<', and '>' in a string of data.
> You can escape other strings of data by passing a dictionary as the optional entities parameter.
> The keys and values must all be strings; each key will be replaced with its corresponding value. The characters '&', '<' and '>' are always escaped, even if entities is provided.

> So for the tests, it's important to check that "complete" force fields be written by ParmEd and read by OpenMM, but there's no need to do this for unit tests involving just residue template conversions.

Sure.

jchodera:
Done.

jchodera:
Looks like one test timed out. I'll push a trivial commit.

@swails: Is it OK that I'm not printing anything at this point? Exporting the CHARMM forcefield now takes 20 minutes producing no output while it is doing so, so that's going to probably make some people using it think things have hung. What would you think of adding a progress bar there?

jchodera:
I'm also having zero success getting the `logger.debug()` output to print from the [charmm conversion script](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/convert_charmm.py). I've added
```python
import logging
LOGGER = logging.getLogger(__name__)
LOGGER.setLevel(logging.DEBUG)
```
but still no output. Do I have to grab the logger for the specific file I want logged output from? How would I even figure that out?

jchodera:
Actually, I'm [debugging an issue](https://github.com/choderalab/openmm-forcefields/issues/22#issuecomment-327297379), so hold off on merging this.

swails:
> I'm also having zero success getting the logger.debug() output to print from the charmm conversion script. I've added
> 
>```python
> import logging
> LOGGER = logging.getLogger(__name__)
> LOGGER.setLevel(logging.DEBUG)
>```
> but still no output. Do I have to grab the logger for the specific file I want logged output from? How would I even figure that out?

This is a consequence of how the `logging` module works.  Loggers have a hierarchical relationship in which each logger (except for the "root" logger) has a parent.  They each implement their own behavior -- filters, levels, handlers, etc. -- but by default will propagate messages to their parent `Logger` (unless you set the `propagate` attribute to `False`).  But this is a good thing -- it lets client applications control the debug levels of their 3rd-party libraries, and lets you very selectively turn off certain logging.

For instance, if you want to turn off debug output from the `requests` library, but turn it on for the rest of the application, you can add a single command like such:

```python
import logging

logging.getLogger('requests').setLevel(logging.ERROR)
```

and that will set the warning level to print out only errors *for the entire requests library*.  Even though each submodule therein has its own logger, each one is an ancestor to `getLogger('requests')`, so since each log message generated within requests passes through that logger, it's enough to filter messages there.

Now on to the explanation of what's happening here.  When you call `LOGGER.setLevel(logging.DEBUG)`, you are ensuring that those debug messages do not get filtered out by that logger.  However, the default log level of the root logger remains unchanged.  So your messages get filtered out by *that* logger (and thus never printed to the screen).

The solution is to set the root level logger to `DEBUG` in the script (or interactive session) you are using to do the debugging.  So at the top of that file/REPL session/Jupyter notebook put the following code:

```python
import logging

logging.basicConfig(level=logging.DEBUG)

# Now write the debug script below
```

I don't fully understand the precise behavior of `logging` -- just the broad strokes.

jchodera:
Thanks for the clarification, @swails! 

I've changed this to "[WIP]" because I am debugging a failure that occurs when trying to convert the whole CHARMM36 forcefield. For some reason, we seem to have the now have a "disappearing residue", where a residue previously found compatible with a patch disappears from `self.residues`:
```
Compatible residue not found in self.residues
   patch name: GLYP
   valid patch combinations: ['DELB', 'GLY', 'DGLY', 'GLYM', 'MET2', 'MET1', 'FHEM']
   residue name: DELB
  File "convert_charmm.py", line 74, in <module>
    params_omm.write(ffxml_filename, provenance=provenance)
  File "/Users/choderaj/github/parmed/ParmEd/parmed/openmm/parameters.py", line 235, in write
    self._write_omm_patches(dest, valid_patch_combinations)
  File "/Users/choderaj/github/parmed/ParmEd/parmed/openmm/parameters.py", line 499, in _write_omm_patches
    raise(e)
  File "/Users/choderaj/github/parmed/ParmEd/parmed/openmm/parameters.py", line 493, in _write_omm_patches
    residue = self.residues[residue_name]
KeyError: 'DELB'
```
Honestly, the only thing I can think of that might be happening is that CHARMM uses the same name for both a residue *and* a patch in different places. **shudder**

jchodera:
Yep, that's exactly what's going on:
```
  File "convert_charmm.py", line 68, in <module>
    params = CharmmParameterSet(*charmm_files)
  File "/Users/choderaj/github/parmed/ParmEd/parmed/charmm/parameters.py", line 120, in __init__
    self.read_stream_file(strf)
  File "/Users/choderaj/github/parmed/ParmEd/parmed/charmm/parameters.py", line 943, in read_stream_file
    self.read_topology_file(iter(section))
  File "/Users/choderaj/github/parmed/ParmEd/parmed/charmm/parameters.py", line 900, in read_topology_file
    raise ValueError('CHARMM tried to define these residues that are already defined as patches: %s' % new_residues_that_are_existing_patches)
ValueError: CHARMM tried to define these residues that are already defined as patches: {'GLYP'}
```
I hate you CHARMM...

swails:
Would using a separate dict for residues and patches work?

jchodera:
> Would using a separate dict for residues and patches work?

I've done that, but the issue is that if a residue is defined *both* as a residue and a patch, what is the correct behavior? Is it (1) these input files are incompatible, or (2) the later definition should override the earlier definition when reading the definitions?

swails:
I don't know the mechanics of how CHARMM does things that well.  If it's not obvious from context whether a patch should be used or a residue, then I have no idea what the proper behavior should be.

ChayaSt:
> I've done that, but the issue is that if a residue is defined both as a residue and a patch, what is the correct behavior?

Are they duplicates of each other (the residue with the same name as the patch is the patch applied to the residue) or different final templates with different names? I'll check that. If they're just duplicates of each other we can have the residue override the patch.

jchodera:
Not sure! Can you look to find what `GLYP` is defined as in the CHARMM files you included in the YAML file?

ChayaSt:
> Not sure! Can you look to find what GLYP is defined as in the CHARMM files you included in the YAML file?

`GLP` is the Glycine N-terminus and the a residue in the [lipid](https://github.com/choderalab/openmm-forcefields/blob/98e57c3725d8e4f86860def77e35451e0c9c2a76/charmm/toppar/stream/lipid/toppar_all36_lipid_model.str#L1327) forcefield. I think in general, they're not the same.

jchodera:
@swails : OK, bugs have been eliminated, tests have been updated, and debug code has been removed. Travis passes. This should be ready to go!

swails:
run jenkins

jchodera:
Looks like there's a Jenkins Mac OS X Python 2.7 failure, but it's totally unrelated:
```
Tit   = total number of iterations
Tnf   = total number of function evaluations
Tnint = total number of segments explored during Cauchy searches
Skip  = number of BFGS updates skipped
Nact  = number of active bounds at final generalized Cauchy point
Projg = norm of the final projected gradient
F     = final function value

           * * *

   N    Tit     Tnf  Tnint  Skip  Nact     Projg        F
   99     12     14      1     0     0   1.459D+01  -8.572D+01
  F =  -85.7243138313122     

STOP: TOTAL NO. of ITERATIONS EXCEEDS LIMIT                 

 Cauchy                time 0.000E+00 seconds.
 Subspace minimization time 0.000E+00 seconds.
 Line search           time 0.000E+00 seconds.

 Total User time 0.000E+00 seconds.

Build step 'Execute shell' marked build as failure
Skipping Cobertura coverage report as build was not SUCCESS or better ...
Finished: FAILURE
```

swails:
Actually the error was a failed download due to a server timeout on GitHub. I'll rebuild that job.

jchodera:
Looks like everything passed!

jchodera:
We'll undoubtedly need some cleanup PRs, but this is probably an improvement.

swails:
Thanks!  Looks good!

If you get a chance, some documentation on this in the `docs` folder would also be appreciated.

jchodera:
> If you get a chance, some documentation on this in the docs folder would also be appreciated.

Whoops, that was an oversight!  It would be super helpful if you had any suggestions on what specifically would you like documented and where it should go!

parmed/modeller/residue.py
swails: > Creating a list to iterate over ensures that when I delete one of the elements of self.bonds, I don't mess up the iteration if self.bonds is a generator.

Even if it's not a generator you need `list`.  Python gets grumpy if you change the size of the iterator you're iterating over while you're iterating over it.  By casting to a `list`, you avoid this issue.

(sorry for the zombie comment -- it was something I had meant to say a couple weeks ago and think it may still be marginally helpful)

PR #911, opened 2017-10-27
Support arguments with spaces

swails:
Try to support arguments that have spaces in them by encapsulating them
in quotes


swails:
I need to fold this into #910

PR #913, opened 2017-11-06
write GROMACS top file with molecule specification in separate file

theavey:
Previously, you could write the parameters to a separate file that would then be included in the main top file, but with this PR, the molecules can also be written to a separate file.

This is particularly useful (to me, at least) because ParmEd can be used for format conversion, then combining multiple separate molecules into the same simulation and topology file is much easier when the parameters and molecule specifications are in separate files. If they're all in the same file, the ordering of the final topology file can be annoying because all parameters need to be specified before any moleculetypes are declared. Also, the molecules need to be declared in the order they appear in the coordinate file. The ordering is much easier to change when it only involves moving around single `#include molecule.itp`-type lines.

I've also included a test that is essentially the same as the test of using the `parameters` option.


swails:
Sounds good. You can also revert all the Travis changes — this is getting fixed in my other PR by testing py 3.5 and 3.6 instead. 

I’m almost done with the fixes there and then that can be merged in here to resolve the test failures.

theavey:
Sorry, I really made of mess of my commits and attempted fixes there. This last one is the right one now, with no changes to `travis/install.sh`.

swails:
run jenkins

PR #914, opened 2017-11-07
Bugfix/charmm continuation character

### LINKED ISSUE #908: Line continuation characters in CharmmFile and CharmmStreamFile
CHARMM files seem to support `-` as an end-of-line continuation character.

Lines like [this](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/toppar/stream/misc/toppar_ions_won.str#L498-L499) in are currently causing my conversions to OpenMM ffxml files to fail:
```
NONBONDED nbxmod  5 atom cdiel fshift vatom vdistance vfswitch -
cutnb 14.0 ctofnb 12.0 ctonnb 10.0 eps 1.0 e14fac 1.0 wmin 1.5
```
The cleanest way to handle this would be to update the various ways lines are read in [`parmed/charmm/_charmmfile.py`](https://github.com/ParmEd/ParmEd/blob/master/parmed/charmm/_charmmfile.py) to handle reading these extra lines and appending them correctly.

swails:
Fixes #908

### DIFF
--- a/parmed/charmm/_charmmfile.py
+++ b/parmed/charmm/_charmmfile.py
@@ -51,7 +51,7 @@ class CharmmFile(object):
         for line in self._handle:
             try:
                 idx = line.index('!')
-            except IndexError:
+            except ValueError:
                 # There is no comment...
                 idx = None
                 end = ''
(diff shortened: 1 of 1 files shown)


jchodera:
Looks like this does the trick! Want to merge it and I can test it out more extensively with the CHARMM-to-OpenMM conversion scripts?

jchodera:
Thanks!

PR #915, opened 2017-11-09
Sorting keys is a bad idea!!!

jvermaas:
The order as written in a prm file shouldn't change. I ran into this myself where I was merging together multiple stream files for my own convenience, and noticing that my simulations now complained of missing improper parameter types. Due to the sort here, a simple read parameter files into ParmEd and write them back out again changes your parameters, which is not the point. You can reverse the order of atoms with no effect, but not sort an improper by the dictionary order of the atomtypes, which is what was happening here.


swails:
This is going to break improper assignment when performing CHARMM -> Amber conversions or supporting CHARMM->OpenMM simulations.

The reason I did this is because I could not for the life of me figure out what CHARMM's convention was for listing impropers.  They don't map directly to the ordering in the PSF files (as a general rule), and I found a number of examples where the first atom was *not* the central atom, despite documentation that that was the *usual* behavior.

So I realize that sorting is not the right thing to do, but nothing else I could think of was either, and this way at least worked for parameter assignment.

I'm certainly fine with this change, as long as we figure out a scheme to then properly assign the impropers to the terms in the PSF file.

jvermaas:
In the psf, the improper should either be in the order specified in the parameter file, or its reverse. Eg if the parameter file said A B C D were the types involved, the psf can list them in either ABCD or DCBA order, since in both cases they have the same two planes defined (ABC and BCD planes). The danger here is that your sort is just by dictionary. So if you take an existing parameter type (like NR1  CPH1 CPH2 H from the protein parameters), this might sort to something out of order, like CPH1 CPH2 H NR1. What I did in my own scripts is to try matching against the types in the regular order, the reverse order, the regular order with two wildcards, and the reverse order with two wildcards. If it still doesn't match, the psf has an improper term that wasn't in the parameter file.

jvermaas:
Ahh, here we go. This is the documentation in io.doc. There are five choices for wildcard usage for improper dihedrals;
1) A - B - C - D (all four atoms, double specification allowed) 
2) A - X - X - B
3) X - A - B - C
4) X - A - B - X
5) X - X - A - B
When classifying an improper dihedral, the first acceptable match (from the above order) is chosen. The match may be made in either direction ( A - B - C - D = D - C - B - A).

swails:
Thanks!

jvermaas:
Ohhhh boy did I make a mess of the git. In any event, basically what this amounts to is changing from a strict dictionary sort to determining which order you want to use. So instead of sorted([at1,at2,at3,at4]), you'd do something like:

if at2 < at3 or ((at2 == at3) and (at1 <= at4)):
    key = [at1, at2, at3, at4]
else:
    key = [at4, at3, at2, at1]

This way the sort doesn't accidentally write files that change the physics in CHARMM.

swails:
Check out #981 -- this should fix the problem, I think.

PR #923, opened 2017-11-24
Write nonbond_params in GromacsTopologyFile

mattwthompson:
Hopefully these changes are clear in the code, but I the important change is that I added added a section in `write` that wrote out the `[ nonbond_params ]` directive. I structured it to look like the corresponding section in `read` that reads it in.

Also set `self.nrexcl` in `from_structure` which, along with probably many other attributes, wasn't being set. Did this because I need it in my systems, but shouldn't be unique to them.

@OleinikovasV


parmed/gromacs/gromacstop.py
```diff
@@ -1422,6 +1423,19 @@ def write(self, dest, combine=None, parameters='inline'):
...
+            # Nonbonded parameters
+            if self.has_NBFIX():
+                typemap = dict(self.parameterset.nbfix_types)
+                dest.write('[ nonbond_params ]\n')
+                for key, val in typemap.items():
+                    a1 = key[0]
+                    a2 = key[1]
+                    eps = val[0] # kcal
+                    sig = val[1] # Angstrom
+                    eps *= u.kilocalorie.conversion_factor_to(u.kilojoule)
```
swails: Calculate this value outside the loop so it doesn’t keep getting recalculated inside it.

test/test_parmed_gromacs.py
```diff
@@ -686,6 +686,14 @@ def total_diheds(dlist):
         self.assertGreater(len(top2.urey_bradleys), 0)
         self.assertEqual(len(top2.urey_bradleys), len(ctop.urey_bradleys))
 
+    def test_nonbond_params(self):
+        """ Test the reading  and writing of the `nonbond_params` directive """
+        top = load_file(get_fn('nonbond_params.top'))
+        assert top.has_NBFIX() == True
+        top.write(get_fn('test_nonbond_params.top'))
+        top_test = load_file(get_fn('test_nonbond_params.top'))
+        assert top_test.has_NBFIX() == True
```
swails: Either remove == True or replace == with is

parmed/gromacs/gromacstop.py
```diff
@@ -1422,6 +1423,19 @@ def write(self, dest, combine=None, parameters='inline'):
...
+                    sig *= 0.1
+                    dest.write('{0} {1} 1 {2} {3}\n'.format(
```
swails: What is the 1 here for? Is it always 1?

parmed/gromacs/gromacstop.py
mattwthompson: ~~I'm not sure it is being calculated multiple times. Each unique pair of atom types has a `sig` and `eps` associated with them, and each of these cross interactions should only be interacted with once. Unless `self.parameterset.nbfix_types` has a key for each atom and not atom type.~~ Nevermind, I see you're referring to the conversion factor. I think.

parmed/gromacs/gromacstop.py
mattwthompson: That is `nbfunc`, non-bonded function type. It is 1 for Lennard-Jones or 2 for Buckingham. From what I can tell, ParmEd will read in 2 and set it with `self.nbfunc` if given such a top file, but in this case I'm using `from_structure` and I don't think there is a corresponding attribute to in `Structure`. I can change it to this, which should handle reading and writing a `2` as long as it's always a `GromacsTopologyFile`, but this would need to be set to 1 if converted from a `Structure`.

```
                    dest.write('{0} {1} {2} {3} {4}\n'.format(
                        key[0], key[1], self.nbfunc, sig/2**(1/6), eps))
```

swails:
The tests are failing. Looks like there is a linting error: https://travis-ci.org/ParmEd/ParmEd/jobs/307192106#L708

There’s an unused a2 variable. Can you take a look at it?

Thanks!

parmed/gromacs/gromacstop.py
```diff
@@ -1422,6 +1423,20 @@ def write(self, dest, combine=None, parameters='inline'):
...
+                eps_conversion = u.kilocalorie.conversion_factor_to(u.kilojoule)
+                for key, val in typemap.items():
+                    a1 = key[0]
+                    a2 = key[1]
+                    eps = val[0] # kcal
+                    sig = val[1] # Angstrom
+                    eps *= eps_conversion
+                    sig *= 0.1
+                    dest.write('{0} {1} 1 {2} {3}\n'.format(
+                        key[0], key[1], sig/2**(1/6), eps))
```
swails: Looks like key[0] and key[1] can be replaced with a1 and a2 respectively. Or just get rid of a1 and a2. 

These unused variables are breaking the build because there’s a linting gate that fails the tests when errors are detected.

mattwthompson:
My bad, thanks for noticing that. The tests should be passing now.

mattwthompson:
... and of course they aren't. I'm not very good at reading these log files, and the size of the log file for this project doesn't make it any easier. The error is `Tests genopen reading a ftp remote file ... FAIL`. This doesn't seem related to my changes would I would assume an external server is down. The test fails locally, so I would think so.

mattwthompson:
It is now passing locally, so I assume it was a server issue somewhere else.

swails:
run jenkins

swails:
Thanks!

PR #925, opened 2017-11-27
More helpful error messages when parameter assignment fails

Eigenstate:
Currently ParmEd raises an error message when atoms are bonded or angled to themselves, but doesn't say which atoms. This can make it difficult to find which parameters or atom types are problematic. Now it prints which atoms caused the exception to be thrown as well.


swails:
run jenkins

swails:
run jenkins

PR #926, opened 2017-11-28
allows variable precision output for box vectors for gromacs output

mrshirts:


swails:
The failing test just needs to have its comparison criteria loosened a little.

parmed/gromacs/gromacsgro.py
```diff
@@ -253,17 +254,17 @@ def write(struct, dest, precision=3, nobox=False):
             a, b, c = reduce_box_vectors(*box_lengths_and_angles_to_vectors(
                             *struct.box))
             if all([abs(x-90) < TINY for x in struct.box[3:]]):
-                dest.write('%10.5f'*3 % (a[0]/10, b[1]/10, c[2]/10))
+                dest.write(boxfmt*3 % (a[0]/10.0, b[1]/10.0, c[2]/10.0))
```
ctk3b: Looks sensible to me.

Nitpick: the `10.0` changes are not necessary when doing `from __future__ import division`. It will perform floating point division even when operating on integers.

swails: And what's more is that leaving it as 10 is marginally faster :smiley:

```python
In [1]:

def divide_10(number):
    return number / 10
​
def divide_10_0(number):
    return number / 10.0

In [2]:

%timeit divide_10(123456)
126 ns ± 2.88 ns per loop (mean ± std. dev. of 7 runs, 10000000 loops each)
In [3]:

%timeit divide_10_0(123456)
213 ns ± 2.76 ns per loop (mean ± std. dev. of 7 runs, 1000000 loops each)
```

The reason for this is that Python caches small integers at startup for performance reasons.  As a result, leaving them as 10 requires Python simply to fetch the cached integer, whereas for the latter Python has to create a new float for 10.0 (although it would use the same instance in all 3 places in that statement).

Here is a demo of that:

```python

In [8]:

x = 10
y = 10
a = 10.0
b = 10.0
In [9]:

x is y
Out[9]:
True
In [10]:

a is b
Out[10]:
False
```

Of course this is an unnoticeable cost, the discussion is purely academic :).  And for other Python implementations that may not cache small integers there may be no benefit. (e.g., pypy which ParmEd explicitly supports and provides speedups around 5x over CPython).

swails:
run jenkins

swails:
Still needs one of the test criteria loosened a little.

parmed/gromacs/gromacsgro.py
mrshirts: > 10.0 changes are not necessary when doing from __future__ import division. It will perform floating point division even when operating on integers.

I would argue that it's more transparent to code maintainers that it is a unit conversion that is fundamentally a floating point conversion, not an integer conversion, even if one is guaranteed to get the same answer either way.

mrshirts:
What are the other issues here that need to be resolved?  I'm not sure how to loosen the criteria in a smart way.

parmed/gromacs/gromacsgro.py
mrshirts: And we can always multiply by 0.1 instead.  I wonder what that speed would be.

swails:
> What are the other issues here that need to be resolved? I'm not sure how to loosen the criteria in a smart way.

The test method has a keyword that lets you specify either the absolute or relative tolerance.  So in `test_parmed_gromacs.py`, line 1167, change the code from

```python
    self.assertAlmostEqual(gro.box[0], 74.1008)
```

to

```python
    self.assertAlmostEqual(gro.box[0], 74.1008, places=2)
```

mrshirts:
Huh.  Is this something that was introduced just with the changes I made?  It's not clear to me why this would result in a tolerance needing to be loosened.

parmed/gromacs/gromacsgro.py
swails: > And we can always multiply by 0.1 instead. I wonder what that speed would be.

That would almost certainly be faster :)

swails:
> Huh. Is this something that was introduced just with the changes I made? It's not clear to me why this would result in a tolerance needing to be loosened.

You changed the default precision with which unit cell dimensions are stored.  Why wouldn't you expect this to change the unit cell dimensions read back out from that file beyond the strict tolerance with which the comparison is currently being done?

parmed/gromacs/gromacsgro.py
swails: I had a little more time to look deeper into this.  The [format spec](http://manual.gromacs.org/current/online/gro.html) says that the unit cell dimensions are free-format, space-delimited reals.  This code actually *reduces* the default precision of unit cell vectors, which I'd prefer to avoid.  What about using this instead:

```python
boxfmt = '%%%d.%df ' % (max(varwidth, 10), max(precision, 5))
```

This way we can specify a higher precision, but will never get lower.  Also note the extra space at the end.  This is needed to strictly adhere to the whitespace-delimited policy.  There was never really a chance of going "over" as GRO files store these values in nm, but if users can specify this the chance of running into this problem goes up.

swails: Note that with this change, you should not have to update any of the test case tolerances.

mrshirts:
OK, I will go back and submit a new patch from the new master.  Specifically, it will multiply by 0.1 in the box vector conversion, and use Jason's suggestion for the box conversion with increased precision.  I will refer to this patch and the issue there.

mrshirts:
Suspended in favor of #935.

PR #927, opened 2017-12-01
Redox Potential functionalities into Parmed; Corrections into cpinutil.py

vwcruzeiro:
These were the changes performed:

1) I modified titratable_residues.py for having two different types of residues: pH-active and redox-active. It can now either generate CPIN or CEIN files.

2) Based on modifications on titratable_residues.py, cpinutil.py is now able to generate CPIN files in the
AMBER 18 format and also has a flag to generate the file in the old format compatible with AMBER 16 and older. A warning message is printed to the users telling them not to use the old file format for
temperatures other than 300 K.

3) New AmberTool: ceinutil.py (the equivalent of cpinutil.py for the Redox Potential).

4) New parmed command: changeReduState, equivalent to changeProtState for reduction states

5) I fixed a couple of typos in the table obtained by the command "--describe" on cpinutil.py

6) I removed the automatic assignment of IS reference energies to ES if these were not provided.


swails:
The failing test is a result of the PDB changing 205l silently, so it is no longer identical to the one saved.  #928 fixes that.  Once that's merged and this branch is updated, I can merge this PR.

hainm:
why not using ‘changeReduceState’ or something similar? It’s easier to remember.

vwcruzeiro:
Hello @hainm, even though I agree with you that a command like the one you suggested is easier to memorize, your argument would also apply for why changeProtState was used instead of changeProtonationState. I tried to be consistent with @swails 's choice. Also, ```help changeReduState``` clarifies to the user what the command does.

swails:
`changeProtState` saves `onation` typing, and in my experience "protonation" is often mistyped (at least by me), and the typos can be hard to find.

What about `changeRedoxState`?  It's a bit clearer and only 1 extra letter.  Redox and Reduce are easier to avoid (and find) typos, so there's less of an argument in favor of abbreviation here, I think.

hainm:
> What about changeRedoxState?

sounds good to me too.

vwcruzeiro:
> What about changeRedoxState?

Good idea, I like it (even though I generally say Reduction state :smile: )

parmed/tools/actions.py
```diff
@@ -1349,6 +1352,59 @@ def execute(self):
                                        'change the protonation state')
             # Actually make the change
             self.parm.parm_data['CHARGE'][atom.idx] = atom.charge = charges[i]
+            
+#+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+-+
+
+class changeReduState(Action):
```
swails: Can you change this to `changeRedoxState`?

PR #935, opened 2018-01-01
Added improved handling of increased precision for box vectors.

mrshirts:
This is a follow up of #926 where I make it possible to increase the precision of the box vectors of gromacs output. See discussion there.  I moved this to a new PR to branch it from the current master, instead of a month ago.

### DIFF
--- a/parmed/gromacs/gromacsgro.py
+++ b/parmed/gromacs/gromacsgro.py
@@ -235,35 +235,36 @@ class GromacsGroFile(object):
         varwidth = 5 + precision
         crdfmt = '%%%d.%df' % (varwidth, precision)
         velfmt = '%%%d.%df' % (varwidth, precision+1)
+        boxfmt = '%%%d.%df ' % (max(varwidth, 10), max(precision, 5))
         for atom in struct.atoms:
             resid = (atom.residue.idx + 1) % 100000
             atid = (atom.idx + 1) % 100000
             dest.write('%5d%-5s%5s%5d' % (resid, atom.residue.name[:5],
                                           atom.name[:5], atid))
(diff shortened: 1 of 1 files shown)


mrshirts:
Huh.  Seems to be erroring because it is exceeding time, but individual tests are completing.  I don't see a real reason this should be an issue. Any thoughts?

mrshirts:
Looking at the errors again, two are timeouts, and one is:

 cc -O2 -fPIC -Wimplicit -DCYTHON_CLINE_IN_TRACEBACK=1 -I/usr/include/libxml2 -Isrc -Isrc/lxml/includes -I/home/travis/.pyenv/versions/pypy-4.0.1/include -c src/lxml/etree.c -o build/temp.linux-x86_64-2.7/src/lxml/etree.o -w
    src/lxml/etree.c:1294:3: error: unknown type name ‘PyBaseExceptionObject’
       PyBaseExceptionObject __pyx_base;
       ^
    Compile failed: command 'cc' failed with exit status 1
    creating tmp
    cc -O2 -fPIC -Wimplicit -I/usr/include/libxml2 -I/usr/include/libxml2 -c /tmp/xmlXPathInitEIwT1s.c -o tmp/xmlXPathInitEIwT1s.o
    /tmp/xmlXPathInitEIwT1s.c:2:1: warning: return type defaults to ‘int’ [-Wimplicit-int]
     main (int argc, char **argv) {
     ^
    cc tmp/xmlXPathInitEIwT1s.o -lxml2 -o a.out
    error: command 'cc' failed with exit status 1

None of these seem to be associated with the actual changes in file (PR #936 breaks the topology test, but this one does not break the .gro check).

swails:
Fixed by #936

PR #936, opened 2018-01-01
Increasing precision in the GROMACS output to better match energies

mrshirts:
Increased the precision in order to match energies in AMBER <-> GROMACS <-> AMBER conversions.

I played around and got ROUGHLY the minimum extra amount of precision needed to get matching energies, though it might be a bit of overkill.

An example of how this makes a difference.  This is for one of the molecules in the SAMPL6 prediction (CB8-G3-0) with the posted coordinates, and an abrupt LJ/COUL cutoff of 1.4 nm. Note: for various reasons, easier to use the output of the InterMol conversions script, in kJ/mol, rather than direct amber output.

Previous:

    Type interaction      Input                     Output          Delta
    proper              503.99794560          503.99710880       -0.00083680
    coulomb 1-4      -11453.22971840       -11453.22511600        0.00460240
    coulomb total   -110746.21943280      -110750.16062123        0.15731840
    vdw 1-4              128.97054480         128.97054480        0.0
    vdw total         13044.57604400        13044.20013721       -0.41630800

Now:

    Type interaction      Input              Output           Delta
    proper              503.99794560       503.99794560       0 
    coulomb 1-4      -11453.22971840    -11453.22971840       0
    coulomb total   -110746.21943280   -110746.22026960      -0.00083680     
    vdw 1-4             128.97054480       128.97054480       0     
    vdw total         13044.57604400    13044.57646240        0.00041840

Given that the coulomb total and vdw total are different by essentially rounding (0.0001 kcal/mol for vdw total, the last digit of error in AMBER and 0.0002 kcal/mol for coul total), and cutoffs are very hard to get exactly right, I think this is probably good enough.  There MAY be some terms that aren't quite reported in sufficient precision yet, but this is much better. 

NOTE: this also includes the box vector fix in #935 in order to get this degree of agreement.  I will properly merge them once that one is taken care of.


mrshirts:
Hmm, looks like a fail in [fail] 1.94% test_format_conversions.TestAmberToGromacs.test_benzene_cyclohexane: 14.7641s

ctk3b:
The failure is a regression test that compares the files written by gromacs so naturally this fix will require updating that file. 

LGTM % updating the test file and assuming that the Travis timeouts are spurious.

mrshirts:
I added a topology file with higher precision; this was converted with my modified version of ParmEd on the saved comparison file.  Inspection showed they are the same, except for the added precision.

PR #940, opened 2018-01-18
Fix for PyRosetta4 Pose to ParmEd structure conversion

dww100:
I was having an issue where trying to convert PyRosetta4 Pose objects into ParmEd structures was making my system hang. This appears to be caused by calling the `tuple` function on the `pyrosetta.rosetta.numeric.xyzVector_double_t` containing the atomic coordinates. The small change in the code below fixes this.

I also updated the import to the new style (importing from rosetta rather than pyrosetta is marked to become deprecated and produces a warning).

### DIFF
--- a/parmed/rosetta/pose.py
+++ b/parmed/rosetta/pose.py
@@ -12,7 +12,7 @@ from parmed.topologyobjects import Atom, ExtraPoint, Bond
 from parmed.utils.six.moves import range
 
 try:
-    from rosetta import Pose, AtomID
+    from pyrosetta import Pose, AtomID
 except ImportError:
     Pose = AtomID = None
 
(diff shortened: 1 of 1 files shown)


parmed/rosetta/pose.py
```diff
@@ -74,7 +74,7 @@ def load(pose):
                 else:
                     atom = Atom(**params)
 
-                atom.xx, atom.xy, atom.xz = tuple(at.xyz())
+                atom.xx, atom.xy, atom.xz = (at.xyz()[0], at.xyz[1], at.xyz[2])
```
swails: This doesn't look quite right.  why is at.xyz() called on the first, but not on the second and third?

Basically, I would have expected either:

```python
atom.xx, atom.xy, atom.xz = at.xyz()[0], at.xyz()[1], at.xyz()[2]
```
or
```python
atom.xx, atom.xy, atom.xz = at.xyz[0], at.xyz[1], at.xyz[2]
```

Am I missing something? Like a side-effect of at.xyz that turns it from a method to a property after the first call?

dww100:
No, that's me being an idiot at the end of a long day and pushing the wrong thing.

Issue #941, opened 2018-01-20
Writing water models for OpenMM: CHARMM includes H-H bond, OpenMM does not


jchodera:
cc: https://github.com/pandegroup/openmm/issues/1939

@swails: CHARMM parameter files define waters with H-H bonds to make them rigid. For example, for [CHARMM TIP3P](https://github.com/choderalab/openmm-forcefields/blob/master/charmm/toppar/toppar_water_ions.str#L65-L70):
```
RESI TIP3         0.000 ! tip3p water model, generate using noangle nodihedral
GROUP
ATOM OH2  OT     -0.834
ATOM H1   HT      0.417
ATOM H2   HT      0.417
BOND OH2 H1 OH2 H2 H1 H2    ! the last bond is needed for shake
```
On the other hand OpenMM does *not* retain the H-H bond in the OpenMM ffxml templates: specifying a water model with a nonchemical bond is not OpenMM compliant.

I need to create a bugfix for our CHARMM -> OpenMM pipeline to generate OpenMM compliant water residues without the H-H bond, but I'm trying to figure out the correct behavior for reading these parameter sets into ParmEd.

There are a number of possibilities here:
* Keep H-H bonds internally, but remove on writing to OpenMM. (What should Amber do?)
* Remove H-H bonds on CHARMM reading (which might result in writing CHARMM files without constraints)
* Remove H-H bonds on CHARMM reading, always adding them back on CHARMM writing only
* Remove H-H bonds on CHARMM reading, maybe keeping some sort of `rigid_water` flag that can be used if needed on writing

swails:
General guiding principles I've been trying to follow:

1. Read everything
1. Retain everything
1. Be pragmatic: be flexible with the above principles when it makes sense.

If information needs to be shed or a lossy transformation needs to happen, it happens as late as possible (if it's done invisibly to the user so the loss doesn't propagate back up, that's best).

So:

> Keep H-H bonds internally, but remove on writing to OpenMM.

This one does the best job at keeping with the above principles.  Another tweak on this would be to shed the H-H bonds when converting from CHARMM to OpenMM (basically the `pmd.openmm.OpenMMParameterSet.from_parameterset()` call).  If it detects H-O-H with a H-H bond, just drop the H-H bond.

> (What should Amber do?)

[...]

PR #942, opened 2018-01-22
Have OpenMMParameterSet remove non-chemical H-H bonds from water upon creation

jchodera:
This PR partially addresses https://github.com/ParmEd/ParmEd/issues/941

- [x] `OpenMMParameterSet.from_parameterset()` now removes unphysical H-H bonds from waters
- [x] Added an `empirical_chemical_formula` property to `Residue` that returns the Hill notation chemical formula (e.g. `H2O`, `C6H6`), provided the `Atom` objects in the residue have `atomic_number` correctly set. Note that the CHARMM parameter reader does not actually set the `atomic_number` fields, so this must be done defensively by `OpenMMParameterSet.from_parameterset()`. (If it's a bug that the CHARMM reader doesn't set this correctly, we should fix that.)
- [x] Added `delete_bond` method to `Residue`
- [x] Added `element_name` field to `Atom`

Currently, the [`test_write_xml_parameters_charmm`](https://github.com/ParmEd/ParmEd/blob/master/test/test_parmed_openmm.py#L612-L633) will fail because, once H-H bonds are removed, there are now two identical water residue templates in [`toppar_water_ions.str`](https://github.com/ParmEd/ParmEd/blob/master/test/files/toppar_water_ions.str#L65-L87), causing OpenMM's `ForceField` to throw an exception. This means that there is no way for ParmEd to correctly read CHARMM parameter sets and write them to OpenMM if they contain duplicate waters, as the standard `toppar_water_ions.str` does. 

To fix this, I am currently presuming we will need to add the capability to specify which residue templates to omit when reading a CHARMM parameter file (pending response to https://github.com/choderalab/openmm-forcefields/issues/56#issuecomment-359291070).

cc: @peastman
cc: https://github.com/choderalab/openmm-forcefields/issues/56
cc: https://github.com/pandegroup/openmm/issues/1939#issuecomment-359130400

### DIFF
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -10,7 +10,6 @@ import os
 from parmed.residue import AminoAcidResidue, RNAResidue, DNAResidue
 from parmed.structure import Structure
 from parmed.topologyobjects import Atom, Bond, AtomList, TrackedList
-from parmed.periodic_table import AtomicNum, Element
 from parmed.utils.six import iteritems
 from parmed.exceptions import IncompatiblePatchError
 import warnings
@@ -288,8 +287,7 @@ class ResidueTemplate(object):
         # Count number of appearances of each element
         element_count = defaultdict(int)
         for atom in self.atoms:
-            element = Element[atom.atomic_number]
-            element_count[element] += 1
+            element_count[atom.element_name] += 1
         # Pop EPs if they are present, since they are not chemical
         if 'EP' in element_count:
             element_count.pop('EP')
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -157,10 +157,7 @@ class ResidueTemplate(object):
         # Remove all bonds involving this atom
         for bond in list(self.bonds):
             if (bond.atom1 == atom) or (bond.atom2 == atom):
-                # Remove this bond from all atoms it spans
-                bond.delete()
-                # Delete the bond from the residue
-                self.bonds.remove(bond)
+                self.delete_bond(bond)
 
         # Remove all impropers involving this atom
         for impr in list(self._impr):
@@ -229,8 +226,11 @@ class ResidueTemplate(object):
             The bond to be deleted
 
         """
-        bond.delete()
-        self.bonds.remove(bond)
+        if bond in self.bonds:
+            bond.delete()
+            self.bonds.remove(bond)
+        else:
+            raise Exception('The specified bond {} is not present in this residue {}'.format(bond, self))
 
     @classmethod
     def from_residue(cls, residue):
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -230,7 +230,7 @@ class ResidueTemplate(object):
             bond.delete()
             self.bonds.remove(bond)
         else:
-            raise Exception('The specified bond {} is not present in this residue {}'.format(bond, self))
+            raise ValueError('The specified bond {} does not belong to this residue {}'.format(bond, self))
 
     @classmethod
     def from_residue(cls, residue):
--- a/parmed/modeller/residue.py
+++ b/parmed/modeller/residue.py
@@ -3,13 +3,14 @@ This contains the basic residue template and residue building libraries
 typically used in modelling applications
 """
 
-from collections import OrderedDict
+from collections import OrderedDict, defaultdict
 import copy as _copy
 import numpy as np
 import os
 from parmed.residue import AminoAcidResidue, RNAResidue, DNAResidue
 from parmed.structure import Structure
 from parmed.topologyobjects import Atom, Bond, AtomList, TrackedList
+from parmed.periodic_table import AtomicNum, Element
 from parmed.utils.six import iteritems
 from parmed.exceptions import IncompatiblePatchError
 import warnings
@@ -220,6 +221,18 @@ class ResidueTemplate(object):
         if atom1 not in atom2.bond_partners:
             self.bonds.append(Bond(atom1, atom2, order=order))
 
+    def delete_bond(self, bond):
+        """ Delete a bond from this residue template.
+
+        Parameters
+        ----------
+        bond : :class:`Bond`
+            The bond to be deleted
+
+        """
+        bond.delete()
+        self.bonds.remove(bond)
+
     @classmethod
     def from_residue(cls, residue):
         """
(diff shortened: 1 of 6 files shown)


jchodera:
@swails: I could use your guidance on how best to allow us to specify a sensible behavior for directing CHARMM -> OpenMM conversions to exclude certain topology residues to avoid duplicates. Some ideas:
1. Add an initializer kwarg:
```python
        params = CharmmParameterSet(*charmm_files, exclude_residue_templates=['TP3M'])
```
2. Simply use `del` to manually delete problematic residues in tests
```python
# read CHARMM parameters
charmm_params = pmd.charmm.CharmmParameterSet(get_fn('par_all36_prot.prm'),
                                              get_fn('top_all36_prot.rtf'),
                                              get_fn('toppar_water_ions.str'))
# drop the problematic residue to avoid the next step throwing an exception
del charmm_params.residues['TP3M']
# convert to OpenMM
openmm_params = openmm.OpenMMParameterSet.from_parameterset()
```
3. Something else?

swails:
For now I’d say just use del to remove offending residue templates. 

But why not have a separate set of xml files for the different water models?

jchodera:
That's the plan! I just wanted to figure out if we wanted to add additional API support to ParmEd to make it possible to read CHARMM parameter sets that could be exported without further manipulation. I'll use the del approach for now to fix the test and then forge on with the CHARMM conversion scripts.

swails:
Honestly to me the question boils down to whether it makes sense to run the del call inside the function by specifying what to del as an argument to the reader vs. just making the client run the del command themselves. I’d rather not add that complexity to ParmEd.

jchodera:
Hm:
```
ERROR: Tests automatic loading of downloaded Amber mdcrd file
  File "/home/travis/build/ParmEd/ParmEd/test/test_parmed_formats.py", line 2238, in test_download_amber_traj_ascii
    self.assertTrue(amber.AmberMdcrd.id_format(self.url + 'tz2.truncoct.crd'))
  File "/home/travis/.pyenv/versions/pypy-4.0.1/site-packages/parmed/amber/asciicrd.py", line 485, in id_format
    f = genopen(filename, 'r')
  File "/home/travis/.pyenv/versions/pypy-4.0.1/site-packages/parmed/utils/io.py", line 71, in genopen
    raise IOError('Could not open %s: %s' % (name, e))
IOError: Could not open https://github.com/ParmEd/ParmEd/raw/master/test/files/tz2.truncoct.crd: HTTP Error 504: Gateway Time-out
```

jchodera:
Whoa. No idea what this means:
```
ERROR: Test NetCDF and ASCII restart and trajectory reporters (no PBC)
  File "/home/travis/build/ParmEd/ParmEd/test/test_openmm_reporters.py", line 194, in test_reporters
    ntraj = [NetCDFTraj.open_old(get_fn('traj1.nc', written=True)),
  File "/home/travis/miniconda/envs/myenv/lib/python2.7/site-packages/parmed/amber/netcdffiles.py", line 544, in open_old
    inst = cls(fname, 'r')
  File "/home/travis/miniconda/envs/myenv/lib/python2.7/site-packages/parmed/amber/netcdffiles.py", line 406, in __init__
    self._ncfile = NetCDFFile(fname, mode, mmap=False)
  File "/home/travis/miniconda/envs/myenv/lib/python2.7/site-packages/parmed/utils/netcdf.py", line 286, in __init__
    self._read()
  File "/home/travis/miniconda/envs/myenv/lib/python2.7/site-packages/parmed/utils/netcdf.py", line 598, in _read
    self.__dict__['version_byte'] = fromstring(self.fp.read(1), '>b')[0]
DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead
```

jchodera:
@swails : These failures seem unrelated to the changes I've made. Perhaps travis just needs to be restarted?

jchodera:
Maybe this is a change in some upstream dependency?
```
ERROR: Parsing Amber netcdf file that does not have box info
  File "/home/travis/build/ParmEd/ParmEd/test/test_parmed_amber.py", line 1689, in test_parsing_netcdf_file_without_unitcells
    parm = pmd.load_file(tn, fn)
  File "/home/travis/miniconda/envs/myenv/lib/python3.6/site-packages/parmed/formats/registry.py", line 197, in load_file
    return cls.parse(filename, *args, **kwargs)
  File "/home/travis/miniconda/envs/myenv/lib/python3.6/site-packages/parmed/amber/amberformat.py", line 359, in parse
    return LoadParm(filename, *args, **kwargs)
  File "/home/travis/miniconda/envs/myenv/lib/python3.6/site-packages/parmed/amber/readparm.py", line 73, in LoadParm
    f = load_file(xyz)
  File "/home/travis/miniconda/envs/myenv/lib/python3.6/site-packages/parmed/formats/registry.py", line 176, in load_file
    if cls.id_format(filename):
  File "/home/travis/miniconda/envs/myenv/lib/python3.6/site-packages/parmed/amber/netcdffiles.py", line 57, in id_format
    f = NetCDFFile(filename, 'r', mmap=False)
  File "/home/travis/miniconda/envs/myenv/lib/python3.6/site-packages/parmed/utils/netcdf.py", line 286, in __init__
    self._read()
  File "/home/travis/miniconda/envs/myenv/lib/python3.6/site-packages/parmed/utils/netcdf.py", line 598, in _read
    self.__dict__['version_byte'] = fromstring(self.fp.read(1), '>b')[0]
DeprecationWarning: The binary mode of fromstring is deprecated, as it behaves surprisingly on unicode inputs. Use frombuffer instead
```

swails:
> Maybe this is a change in some upstream dependency?

No, that's Python 3.6 deprecating something that `netcdf.py` uses :(

parmed/modeller/residue.py
```diff
@@ -220,6 +220,18 @@ def add_bond(self, atom1, atom2, order=1.0):
...
+        """ Delete a bond from this residue template.
+
+        Parameters
+        ----------
+        bond : :class:`Bond`
+            The bond to be deleted
+
+        """
+        bond.delete()
+        self.bonds.remove(bond)
```
swails: This should be done IFF `bond` is in `self.bonds`, yes?

parmed/openmm/parameters.py
```diff
@@ -91,6 +91,32 @@ def __init__(self, *filenames):
...
+        for atom in residue.atoms:
+            if atom.type not in params.atom_types_str:
+                raise Exception('Residue {} contains atom type {} not found in parameter set'.format(residue.name, atom.type))
+            atom.atomic_number = params.atom_types_str[atom.type].atomic_number
+
+        # Check waters
+        if residue.empirical_chemical_formula == 'H2O':
+            # Remove any H-H bonds if they are present
+            for bond in residue.bonds:
+                if (bond.atom1.element_name == 'H') and (bond.atom2.element_name == 'H'):
```
swails: Is `element_name` an attribute?  I don't ever remember that being an attribute.  Why not use `atomic_number`?

parmed/openmm/parameters.py
```diff
@@ -91,6 +91,32 @@ def __init__(self, *filenames):
...
+                    LOGGER.debug('Deleting H-H bond from water residue {}'.format(residue.name))
+                    residue.delete_bond(bond)
```
swails: You can't do this.  `residue.delete_bond` modifies `residue.bonds` as you're iterating over it.  Python doesn't like that.

You either need to loop over `list(residue.bonds)` or keep a list of bonds to remove from `self.bonds` after the iteration over all bonds is done.

jchodera:
Thanks for the great feedback! Working on this now!

Two quick questions for @swails:
* @rafwiewiora notes that the previous behavior was that `OpenMMParameterSet` would happily import a parameter set where some residue templates had missing atom type definitions, and would just silently not write out those residues when exporting an `ffxml`. Are you happy to keep that behavior?
* Should parameter set readers guaranteed to set their `Residue.atom.atomic_num` fields to the correct atomic numbers, or is this something that is not guaranteed to be defined at any time?

parmed/openmm/parameters.py
jchodera: I just added `element_name` as a convenience property, since this makes the code clearer to understand.

jchodera:
I think I've added the necessary test coverage, and have changed the code to drop residues without atom type definitions when converting to `OpenMMParameterSet`. If that's not the desired behavior, I'm happy to fix!

swails:
> @rafwiewiora notes that the previous behavior was that OpenMMParameterSet would happily import a parameter set where some residue templates had missing atom type definitions, and would just silently not write out those residues when exporting an ffxml. Are you happy to keep that behavior?

Yes, I'm happy to keep that behavior.

> Should parameter set readers guaranteed to set their Residue.atom.atomic_num fields to the correct atomic numbers, or is this something that is not guaranteed to be defined at any time?

This isn't something that can really be guaranteed, as not all file formats specify this information.  If a good guess can be made somewhat unambiguously (either through comparing atomic masses or some heuristic combination of various properties, like bond pattern + name), then that guess is made.  Otherwise, it's just set to `None`.

swails:
Is this ready?

jchodera:
Yes!

PR #945, opened 2018-01-27
Add handling for CHARMM lone pairs

jchodera:
This PR adds support for CHARMM lone pairs.

- [x] Read `LONEPAIR` directive in CHARMM residue template definitions (only `BISEctor` and `RELAtive` are implemented)
- [x] Store lone pair information in a tuple in `ResidueTemplate`
- [x] Write `<VirtualSite>` entries for OpenMM residue templates
- [x] Eliminate bonds between hydrogens and bonds to lone pairs when importing to `OpenMMParameterSet`

cc: #941 

Open questions for @swails:
* Should I extend the [`ResidueTemplate`](https://github.com/ParmEd/ParmEd/blob/master/parmed/modeller/residue.py#L40-L94) class with a `.lonepairs` field? 
* Should `.lonepairs` be a list of `tuple`s, or should I define a `LonePair` type in `topologyobjects.py`? If a special type, what does it inherit from? Does it have to have special properties, like be a `TrackedList`?
* Should I just generalize this and go for `VirtualSite` instead of `LonePair`?


swails:
How is a `LonePair` different from an `ExtraPoint`?

parmed/charmm/parameters.py
```diff
@@ -858,6 +858,21 @@ def read_topology_file(self, tfile):
...
                         elif line[:6].upper() == 'ACCEPT':
                             pass
+                        elif line[:8].upper() == 'LONEPAIR':
+                            # See: https://www.charmm.org/charmm/documentation/by-version/c40b1/params/doc/lonepair/
+                            # TODO: This currently doesn't handle some formats, like Note 3 in the above URL
+                            words = line.split()
+                            lptype_keyword = words[1][0:4].upper()
+                            if lptype_keyword not in ['BISE', 'RELE']:
+                                raise CharmmError('LONEPAIR type {} not supported; only BISEctor and RELEtive supported.'.format(words[1]))
+                            a1, a2, a3, a4 = words[2:6]
```
swails: Improve error catching if the line doesn't have enough words.

parmed/modeller/residue.py
```diff
@@ -81,6 +81,7 @@ class ResidueTemplate(object):
     def __init__(self, name=''):
         self.atoms = AtomList()
         self.bonds = TrackedList()
+        self.lonepairs = list() # TODO: Should this be a TrackedList?
```
swails: This should not be a `TrackedList`.  The `TrackedList` is only for instances where if you change the list (add or delete an item), it means other information needs to be recomputed.

Basically, if any of the atom, residue, bond, ... etc. arrays in an `AmberParm` change, all of the pointer sections need to be recomputed.

jchodera:
> How is a LonePair different from an ExtraPoint?

Hm, I somehow missed the existence of `ExtraPoint` completely. 

It looks like `ExtraPoint` carries [`weights`](https://github.com/ParmEd/ParmEd/blob/master/parmed/topologyobjects.py#L1038-L1039), but it's not clear to me how the atoms that go along with these weights are defined, nor what [`_frame_type`](https://github.com/ParmEd/ParmEd/blob/master/parmed/topologyobjects.py#L1043) is support or how it should be supplied on construction.

I imagine using `ExtraPoint` will involve parsing the CHARMM `LONEPAIR` cards to see which `Atom` objects (which have already been created) should be promoted to `ExtraPoint` objects. 

I'll also have to add support for writing `ExtraPoint` objects to OpenMM, but may need your help in implementing the other kinds of `ExtraPoint` objects that ParmEd currently supports. It's not easy for me to glean what I'd need to support here.

Also, we just realized that the CHARMM PSF reader/writer does not support lone pairs: https://github.com/choderalab/openmm-forcefields/pull/58#issuecomment-361087055

jchodera:
@swails: I think the best way forward is to merge this and find a way to rework it and the PSF reader/writer in separate PRs.

jchodera:
This is ready to merge @swails.

PR #950, opened 2018-02-12
Fix typo in OpenMM ffxml tag name "CMAPTorsionForce"

jchodera:
cc: https://github.com/choderalab/openmm-forcefields/pull/61#issuecomment-365052830

CMAP forces were incorrectly being written to OpenMM ffxml file with the misspelled tag `<CMapTorsionForce>` instead of the correct `<CMAPTorsionForce>` used [here](https://github.com/pandegroup/openmm/blob/master/wrappers/python/simtk/openmm/app/forcefield.py#L2278). 

I'm trying to think of an appropriate test to make sure that all tags are correctly handled. 
A force-by-force breakdown could work, but is difficult to manage.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -704,7 +704,7 @@ class OpenMMParameterSet(ParameterSet):
     @needs_lxml
     def _write_omm_cmaps(self, xml_root, skip_types):
         if not self.cmap_types: return
-        xml_force = etree.SubElement(xml_root, 'CmapTorsionForce')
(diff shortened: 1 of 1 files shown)


ctk3b:
This should probably live in OpenMM but validating against an [XML schema](http://lxml.de/validation.html) that defines the OpenMM .xml format would catch bugs like this.

jchodera:
> This should probably live in OpenMM but validating against an XML schema that defines the OpenMM .xml format would catch bugs like this.

Agreed! Will create an OpenMM issue to suggest we add versioned schemas.

PR #951, opened 2018-02-13
Ensure processing of CHARMM residues and patches in deterministic order

jchodera:
This PR ensures that patches and residues are processed and compared in deterministic order---the same order as read in. 

This allows us to avoid issues like [this](https://github.com/choderalab/openmm-forcefields/pull/61#issuecomment-365117418) where patches are compared against random base residues for determining what should be added/deleted in a patch.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -601,11 +601,6 @@ class OpenMMParameterSet(ParameterSet):
             if (residue.tail is None) and (patched_residue.tail is not None):
                 etree.SubElement(patch_xml, 'AddExternalBond', atomName=patched_residue.tail.name)
 
-            # DEBUG
-            if (patch.name == 'CTER'):
-                print('residue: {}'.format(residue))
-                print('patched_residue: {}'.format(patched_residue))
-
             if write_apply_to_residue:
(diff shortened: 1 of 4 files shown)


parmed/openmm/parameters.py
```diff
@@ -592,6 +601,11 @@ def _write_omm_patches(self, xml_root, valid_residues_for_patch, write_apply_to_
             if (residue.tail is None) and (patched_residue.tail is not None):
                 etree.SubElement(patch_xml, 'AddExternalBond', atomName=patched_residue.tail.name)
 
+            # DEBUG
+            if (patch.name == 'CTER'):
+                print('residue: {}'.format(residue))
+                print('patched_residue: {}'.format(patched_residue))
+
```
swails: Remove the debug code before merging.

jchodera: Whoops! Thanks!

jchodera:
@swails : All travis branches pass, except for [this one](https://travis-ci.org/ParmEd/ParmEd/jobs/341706182) which just timed out.

jchodera:
This is ready to merge @swails. The failure is a travis timeout.

PR #953, opened 2018-02-15
Update OpenMM ffxml LennardJonesForce writer to write out special 1,4 parameters

jchodera:
cc: https://github.com/choderalab/openmm-forcefields/pull/61#issuecomment-365797924

This PR adds support for writing out `epsilon14` and `sigma14` attributes to `<Atom/>` tags in `<LennardJonesForce>` when writing out OpenMM ffxml parameters that use them.

@peastman is adding support for this to the OpenMM `ForceField` class separately.


jchodera:
Support for this feature has now been added to OpenMM via https://github.com/pandegroup/openmm/pull/1993 and is required for correct energy reproduction of CHARMM systems.

This should be ready to merge @swails

swails:
I don't understand how this would work?  How is this new feature supposed to be used?

swails:
And what about adding a test?

jchodera:
We already test by reading in the converted parameters via OpenMM's ForceField, which will raise an exception if this is misformatted. Once the new OpenMM release is cut, this will be automatically tested.

jchodera:
I could add additional tests, but I'm not sure what additional tests you are looking for. 

Ultimately, I'd like to move some of the energy fidelity testing here, but that is going to be a much bigger change for a separate PR.

TL;DR: There's already a test that it is correctly generating ffxml files that runs locally, but the Travis testing will only be updated once the new OpenMM release is posted.

swails:
Ah, I see -- `separate_ljforce=True` is how to avoid the exception and write the desired LJ force field.

Ideally this would happen at the top of the function and automatically set separate_ljforce` to `True` if necessary, but this is fine for now.

jchodera:
Thanks!

> Ideally this would happen at the top of the function and automatically set separate_ljforcetoTrue` if necessary, but this is fine for now.

Agreed! I left a `# TODO` in there to remind us to rework this later!

PR #955, opened 2018-02-18
Compress impropers when writing OpenMM ffxml

jchodera:
This PR compresses impropers with identical atom type compositions since OpenMM's `ForceField` cannot handle multiple impropers with the same composition.

This was suggested by @peastman: https://github.com/choderalab/openmm-forcefields/pull/61#issuecomment-366411167


jchodera:
Hm, my git cherry-pick got screwed up somehow. Let me fix.

Also, https://github.com/ParmEd/ParmEd/pull/915 may be relevant here.

PR #960, opened 2018-02-26
Fix the calculation of dihedral angle

SGenheden:
The function geometry.dihedral() does not return the correct sign as implemented in e.g. MDAnalysis or UCSF Chimera. 

This obvious in e.g. this structure 
```
HETATM    1  C2  TH1     1      -2.569  -0.169  -0.066  1.00  0.00           C  
HETATM    2  C3  TH1     1      -1.117  -0.539   0.261  1.00  0.00           C  
HETATM    3  C4  TH1     1      -0.142   0.462  -0.337  1.00  0.00           C  
HETATM    5  C6  TH1     1       1.323   0.205  -0.469  1.00  0.00           C  
END  
```

where the dihedral angle should be -165 degrees


swails:
Thanks!

PR #962, opened 2018-02-27
Write and read Gromacs itp-files

SGenheden:
This PL adds the options to read/write Gromacs-itp files. That is Gromacs topology files without any definition of parameters or system (e.g. defaults, atomtypes or molecules sections).

I have used this frequently in my research, hopefully it is useful for other people as well.


swails:
Thanks!  I agree this could be helpful.

Issue #959, opened 2018-02-21
The rigidWater=True option in parmed.Structure.CreateSystem is not correctly working 


nividic:
Hi there,

I was trying to create an OpenMM system by using a Parmed structure object. I was selecting the following options:

```new_system = water_structure.createSystem(nonbondedMethod=NoCutoff, rigidWater=True, constraints=None)```

If Parmed wants to mimic OpenMM, the option rigidWater=True should produce a rigid water i.e. bonds and angle constrained during MD runs. However, in the Parmed lines:
https://github.com/ParmEd/ParmEd/blob/641865e3cdda56cce10657a7afedcfce486cf24c/parmed/structure.py#L2127-L2135
the constraints are applied just between the two H-O bonds missing the extra constraints between the two H-H. This is causing the water to be not fully rigid as in the TIP3P water model. I have attached a simple script that you can use to check the problem.
[test_water.zip](https://github.com/ParmEd/ParmEd/files/1742382/test_water.zip)

 @davidlmobley @jchodera

nividic:
I'm going to fix this in a PR request

davidlmobley:
CC @slochower

swails:
Fixed in #969

Issue #958, opened 2018-02-20
The option constraints=HAngles is wrong in parmed.structure.createSystem 


nividic:
Hi there,

I was trying to create an OpenMM system by using a Parmed structure object. I was selecting the following options:
```new_system = water_structure.createSystem(nonbondedMethod=NoCutoff, rigidWater=True, constraints=HAngles)```
The option HAngles should put angle constraints on all the angles of the form H-X-H and H-O-X (where X is an arbitrary atom) and on all bonds. Now, the problem are the lines
https://github.com/ParmEd/ParmEd/blob/641865e3cdda56cce10657a7afedcfce486cf24c/parmed/structure.py#L2140-L2156
In particular the line
https://github.com/ParmEd/ParmEd/blob/641865e3cdda56cce10657a7afedcfce486cf24c/parmed/structure.py#L2155
The bond lengths (l1 and l2) have already been multiplied by the conversion factor and the length between the the two hydrogens does not need to. This is causing to generate  wrong water  geometries. For example the water angle H-O-H was going from ~104 degrees to 9 degrees. 

@davidlmobley @jchodera

davidlmobley:
You're going to do a PR to fix, right?

CC @slochower

jchodera:
That would be good to fix! But we don't recommend using `HAangles` constraints because it can significantly disrupt conformational equilibria.

nividic:
I'm going to work on a PR

swails:
Fixed by #969

PR #969, opened 2018-03-19
[WIP] Fix OpenMM Constraints

nividic:
This is a fix for the issues: [https://github.com/ParmEd/ParmEd/issues/958](https://github.com/ParmEd/ParmEd/issues/958) and [https://github.com/ParmEd/ParmEd/issues/959](https://github.com/ParmEd/ParmEd/issues/959)

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -2130,13 +2130,13 @@ class Structure(object):
         constraint_angle_set = set()
         is_water = _settler(self)
 
-        if constraints == app.AllBonds or constraints == app.HAngles:
+        if constraints is app.AllBonds or constraints is app.HAngles:
             for bond in self.bonds:
                 # Skip all extra points... don't constrain those
                 if isinstance(bond.atom1, ExtraPoint): continue
                 if isinstance(bond.atom2, ExtraPoint): continue
                 constraint_bond_set.add(frozenset((bond.atom1.idx, bond.atom2.idx)))
-        elif constraints == app.HBonds:
+        elif constraints is app.HBonds:
             for bond in self.bonds:
                 if isinstance(bond.atom1, ExtraPoint): continue
                 if isinstance(bond.atom2, ExtraPoint): continue
@@ -2154,7 +2154,7 @@ class Structure(object):
             if frozenset((bond.atom1.idx, bond.atom2.idx)) in constraint_bond_set:
                 system.addConstraint(bond.atom1.idx, bond.atom2.idx, bond.type.req * length_conv)
 
-        if constraints == app.HAngles:
+        if constraints is app.HAngles:
             for angle in self.angles:
                 numH = 0
                 if angle.atom1.element == 1:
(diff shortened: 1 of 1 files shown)


nividic:
@swails I made a fix for the OpenMM Constraints in Parmed and all the tests are passed. The main goal of the patch was to fix the Parmed rigid water option that was not behaving correctly leaving the water partially flexible. In the patch, I still have some concerns related to cases where angle constraints could involve for example virtual sites. I wonder if you can comment on that and overall if you could have a look at the patch and have it merged. Thanks

parmed/structure.py
```diff
@@ -2118,43 +2118,76 @@ def omm_add_constraints(self, system, constraints, rigidWater):
...
         length_conv = u.angstrom.conversion_factor_to(u.nanometer)
-        # Rigid water only
-        if constraints is None:
-            is_water = _settler(self)
+
+        constraint_bond_set = set()
+        constraint_angle_set = set()
+        is_water = _settler(self)
+
+        if constraints == app.AllBonds or constraints == app.HAngles:
```
swails: `app.AllBonds` and `app.HAngles` are better thought of as enums (or singletons).  Compare using `is`, not `==`

parmed/structure.py
```diff
@@ -2118,43 +2118,76 @@ def omm_add_constraints(self, system, constraints, rigidWater):
...
             for bond in self.bonds:
                 # Skip all extra points... don't constrain those
-                if isinstance(bond.atom1, ExtraPoint) or isinstance(bond.atom2, ExtraPoint):
-                    continue
+                if isinstance(bond.atom1, ExtraPoint): continue
+                if isinstance(bond.atom2, ExtraPoint): continue
+                constraint_bond_set.add(frozenset((bond.atom1.idx, bond.atom2.idx)))
+        elif constraints == app.HBonds:
```
swails: `==` -> `is`

parmed/structure.py
```diff
@@ -2118,43 +2118,76 @@ def omm_add_constraints(self, system, constraints, rigidWater):
...
-                    # Law of cosines to find the constraint distance
-                    if l1 is None or l2 is None:
-                        continue # no bonds found...
-                    cost = math.cos(angle.type.theteq*DEG_TO_RAD)
-                    length = math.sqrt(l1*l1 + l2*l2 - 2*l1*l2*cost) * length_conv
-                    system.addConstraint(angle.atom1.idx, angle.atom3.idx, length)
+            if frozenset((bond.atom1.idx, bond.atom2.idx)) in constraint_bond_set:
+                system.addConstraint(bond.atom1.idx, bond.atom2.idx, bond.type.req * length_conv)
 
+        if constraints == app.HAngles:
```
swails: `==` -> `is`

parmed/structure.py
```diff
@@ -2118,43 +2118,76 @@ def omm_add_constraints(self, system, constraints, rigidWater):
...
+                if angle.atom3.element == 1:
+                    numH += 1
+                if numH == 2 or (numH == 1 and angle.atom2.element == 8):
+                    constraint_angle_set.add(frozenset((angle.atom1.idx,
+                                                        angle.atom2.idx,
+                                                        angle.atom3.idx)))
+        if rigidWater:
+            for angle in self.angles:
+                if is_water[angle.atom1.residue.idx]:
+                    constraint_angle_set.add(frozenset((angle.atom1.idx,
```
swails: Order matters here, doesn't it?  So you can't use an unordered collection like set or frozenset.  My guess is that it will _usually_ work, but there may be instances where it'll fail.

swails:
Thanks!  Looks good!

My one last comment is that the angle ordering now won't match the equivalent, but reversed ordering.  However, since this one method constructs the set of constraints to add, and then adds them, from the same `angles` list, the reverse order will never be encountered.  So it's still correct.

I'll merge this as-is.

nividic:
Thank you so much,
Gaetano

nividic:
@swails I'm sorry to bother you again, we would like to make a new conda package in the omnia repository. Would be possible for you to update the Parmed point release to 2.7.4. Thanks a lot

davidlmobley:
To be specific, we're using this on a cloud computing platform where (for practical purposes) we can't use software that isn't available as a conda-installable point release, and having a bug where we accidentally end up with semi-flexible water is obviously not a good way for us to successfully run our calculations. So we can't proceed until we have a new point release (we can update the conda recipe to reflect the new release).

swails:
Done and tagged.  I think all that needs to be done is someone has to update the conda-recipes repo where the recipe for ParmEd lives.

swails:
It's now pip-installable as version 3.0.1 as well, which should help workaround any delays in conda-availability

PR #972, opened 2018-03-23
Bugfix/pdb reading

swails:
Try to fix large PDB file parsing that doesn't follow VMD's or Amber's conventions (like Phenix).

@dacase -- Can you check out these changes and see if it works?  I recall you sending me a file as an example, but I can't seem to find that message anywhere.

Alternatively, send me the file again (last time!) and I'll verify that it works.

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -440,10 +440,16 @@ class PDBFile(object):
                             else:
                                 try:
                                     atnum = int(atnum, 16)
-                                    atom_hex = True
                                 except ValueError:
                                     atnum = last_atom_added.number + 1
                                     atom_other = True
+                                else:
+                                    if atnum == last_atom_added.number + 1:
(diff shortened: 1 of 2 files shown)


dacase:
I'll give this a try.  Note that phenix is not using base32, but something like base36 (10 digits plus 26 uppercase letters).

swails:
> Note that phenix is not using base32, but something like base36 (10 digits plus 26 uppercase letters).

Right -- my head just jumps to powers-of-2 :).  I'm beginning to think, though, that trying so hard to remember numbers for PDB files that overflow is more a wasted effort than anything else.

The main time it's useful to remember numbers in PDB files is for "official" PDB files that, by definition, don't blow past the atom and residue limits and occasionally skip gaps in the sequence. Most software (like VMD) just numbers incrementally when printing "custom" PDB files.  The main thing the residue field should be used for in overflow cases, then, is to identify when a new residue starts.

PR #978, opened 2018-04-16
Bugfix necessary to allow openmoltools to work with recent ParmEd

### LINKED ISSUE #977: AmberParameterSet.from_leaprc produces no residues
Tagging @jchodera 

Somewhere between parmed 2.7.3 and 3.0.1, this forcefield generator in openmoltools stopped working

`openmoltools.forcefield_generators.generateForceFieldFromMolecules`
https://github.com/choderalab/openmoltools/blob/master/openmoltools/forcefield_generators.py#L346

Stepping through it with the debugger, I can see that the line 
`params = parmed.amber.AmberParameterSet.from_leaprc(leaprc)`
ended up producing a  parameter set without any residues defined. 

Below is a minimal example that can capture the problem for me. Note that this feature of openmoltools depends on openeye, so a license is required to run it.

```py
from openmoltools import forcefield_generators, openeye
from lxml import etree


def check_num_elements(xml, elements: list=None) -> int:
    """Check for elements in an xml file"""
    tree = etree.fromstring(xml)

    results = dict()
    for key in elements:
        results[key] =  len(tree.xpath('/ForceField/Residues/{}'.format(key))[:])

    return results


mol = openeye.iupac_to_oemol("2-(Ethylamino)-1,2-diphenylethanol")

ffxml=forcefield_generators.generateForceFieldFromMolecules([mol], ignoreFailures=False, normalize=False) 
print(ffxml)
results = check_num_elements(ffxml, ["Residue/Atom", "Residue/Bond", "Residue"])
assert results["Residue"] == 1, "There should be 1 Residue."
assert results['Residue/Atom'] >0, "No atoms were found, this template is broken"
assert results['Residue/Bond'] >0, "No bonds were found, this template is broken"
```

The forcefield template only has these contents at the end.
```xml
<ForceField>
  <Info>
    <DateGenerated>2018-04-13</DateGenerated>
  </Info>
  <PeriodicTorsionForce>
    <Improper k1="4.6024" periodicity1="2" phase1="3.141592653589793" type1="ca" type2="ca" type3="ca" type4="ha"/>
  </PeriodicTorsionForce>
</ForceField>
```

jchodera:
Fixes #977 and https://github.com/pandegroup/openmm/issues/2047

This PR contains three bugfixes:
* Prune duplicate patches when converting to `OpenMMParameterSet` (https://github.com/pandegroup/openmm/issues/2047)
* The addition of a `remediate_residues` flag to `parmed.openmm.OpenMMParameterSet.from_parameterset()` that allows `openmoltools` to skip the step where residues without defined atom types are discarded (#977)
* Minor performance improvement via deleting unnecessary code

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -198,14 +198,24 @@ class OpenMMParameterSet(ParameterSet):
             if isinstance(residue, ResidueTemplate):
                 if (not remediate_residues) or OpenMMParameterSet._remediate_residue_template(new_params, residue):
                     remediated_residues.append(residue)
-
-        new_params.residues = OrderedDict()
         for residue in remediated_residues:
             new_params.residues[residue.name] = residue
 
+        # Only add unique patches
+        def patch_is_equivalent(patch1, patch2):
+            """Return True if patches are equivalent to OpenMM."""
+            return (set(patch1.add_bonds)==set(patch2.add_bonds)) and (set(patch1.delete_atoms)==set(patch2.delete_atoms))
+
         for name, patch in iteritems(params.patches):
             if isinstance(patch, PatchTemplate):
-                new_params.patches[name] = patch
+                patch_is_unique = True
+                for existing_patch in new_params.patches.values():
+                    if patch_is_equivalent(patch, existing_patch):
+                        warnings.warn('Patch {} discarded because OpenMM considers it identical to {}'.format(patch, existing_patch))
+                        patch_is_unique = False
+                        break
+                if patch_is_unique:
+                    new_params.patches[name] = patch
 
         return new_params
 
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -103,6 +103,14 @@ class OpenMMParameterSet(ParameterSet):
           foreign parameter sets (e.g., CHARMM) into OpenMM parameter sets,
           and not restored on conversion from OpenMM to other formats
 
+        Parameters
+        ----------
+        params : :class:`parmed.parameters.ParameterSet`
+            ParameterSet containing the list of parameters to be converted to a
+            OpenMM-compatible parameter set
+        residue : :class:`parmed.modeller.Residue`
+            The residue to remediate
+
         """
         # Populate atomic numbers in residue template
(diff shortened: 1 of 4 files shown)


parmed/openmm/parameters.py
```diff
@@ -179,16 +190,19 @@ def from_parameterset(cls, params, copy=False):
...
         new_params.default_scee = params.default_scee
         new_params.default_scnb = params.default_scnb
+
         # Add only ResidueTemplate instances (no ResidueTemplateContainers)
         # Maintain original residue ordering
         remediated_residues = list()
         for name, residue in iteritems(params.residues):
             if isinstance(residue, ResidueTemplate):
-                if OpenMMParameterSet._remediate_residue_template(new_params, residue):
+                if (not remediate_residues) or OpenMMParameterSet._remediate_residue_template(new_params, residue, drop_residues_without_parameters=drop_residues_without_parameters):
```
bas-rustenburg: Caught by pyflakes: `drop_residues_without_parameters` is undefined.

jchodera:
Whoops, can you just remove that? It must have someone been checked in during a git accident.

bas-rustenburg:
I don't have access to your fork. Could you fix the bug?

jchodera:
Fixed!

jchodera:
This should be ready to go when tests pass @swails

test/test_parmed_openmm.py
```diff
@@ -531,6 +531,23 @@ def test_write_xml_parameters_gaff(self):
...
+                pmd.amber.AmberParameterSet.from_leaprc(leaprc),
+                remediate_residues=False
+        )
+        ffxml_filename = get_fn('residue.xml', written=True)
+        params.write(ffxml_filename)
+        try:
+            forcefield = app.ForceField(ffxml_filename)
+        except KeyError:
+            # A KeyError is expected
+            pass
```
swails: If a KeyError is expected, should you do an “assert False” in an else clause to trigger a failure if this unexpectedly works?

jchodera:
Added as requested.

jchodera:
@swails : Looks like all test pass! Ready to merge?

PR #981, opened 2018-05-04
Bugfix/do not sort charmm impropers

swails:
Fixes #915

### DIFF
--- a/parmed/charmm/parameters.py
+++ b/parmed/charmm/parameters.py
@@ -13,6 +13,7 @@ import re
 import warnings
 from copy import copy as _copy
 from collections import OrderedDict
+from itertools import combinations
 
 from ..constants import TINY
 from ..exceptions import CharmmError, ParameterWarning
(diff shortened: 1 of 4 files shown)


jvermaas:
This definitely fixes my issue. I just needed a non-dictionary sort. It also looks much cleaner now. :+1:

swails:
It's still failing a lot of the tests because it can't find the improper torsion parameters... :(.  Blech.

swails:
And see this is why I did things the way I did originally.  I followed the rules they laid out in their documentation, but I still get missing parameters on systems I *know* CHARMM works on.

So whatever ruleset they use for assigning improper parameters within CHARMM isn't actually documented *correctly* anywhere I can find it.

jvermaas:
Which check is actually failing? In going through the test cases, I couldn't find one that actually loaded parameter sets in a useful way so that I could recapitulate the failure and use my typical "print" debugging techniques. The way it was done originally was flat out wrong for merging together parameter files, since it would change the physics encoded in the parameter file when written out again.

PR #985, opened 2018-05-25
Added 'cgnr' charge group property to Atom class

jotelha:
…and store accordingly when writing to Gromacs .top topology file.

We have been using ParmEd to work on GROMOS-like force field topologies, which need correct charge groups to be assigned within the topology file. Without knowing much about ParmEd coding, I added minimal changes here in order to be able to set an Atom.cgnr property (integer) and write it into the according column when storing a Gromacs .top file. Previously, each atom's index has been put into this column, resulting in one charge group per atom.

File input methods not changed.

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -1805,7 +1805,7 @@ class GromacsTopologyFile(Structure):
                 dest.write('%5d %10s %6d %6s %6s %6d %10.8f %10.6f   ; '
                            'qtot %.6f\n' % (atom.idx+1, atom.type,
                             residue.idx+1, residue.name, atom.name,
-                            atom.idx+1, atom.charge, atom.mass, runchg))
+                            atom.cgnr, atom.charge, atom.mass, runchg))
         dest.write('\n')
         # Do valence terms now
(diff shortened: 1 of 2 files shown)


swails:
Thanks for the contribution!

Thinking about this a little bit, I think a good approach here would be to create a new `TrackedList` `charge_groups` attached to the `Structure` item that is just a list of `Atom` sets with every atom in that charge group.

The `Atom` could hold a reference to the `ChargeGroup` instance it's a part of, and the standard deletion APIs would remove each atom from its charge group.  The `ChargeGroup` numbers would be auto-tracked like they are for everything else in `Structure`.

That would make it easier to shuffle around atoms within charge groups -- but I suspect that's probably not done very often?  Probably best to just get this in for now and get to my proposal later if it provides any value.

A couple requests:

* Can you add a test case for this?
* Can you look into the test failure?  This is what Travis is saying:

```
FAIL: Test converting binary liquid from Amber prmtop to Gromacs top
  File "/home/travis/build/ParmEd/ParmEd/test/test_format_conversions.py", line 44, in test_benzene_cyclohexane
    self.assertTrue(diff_files(saved, written))
AssertionError: False is not true
```

swails:
run jenkins

PR #988, opened 2018-06-08
Show velocities and box only when available

panxl:


swails:

parmed/amber/netcdffiles.py
```diff
@@ -262,14 +262,18 @@ def coordinates(self, stuff):
...
+            return (vels.reshape((-1, self.atom, 3)) * self.velocity_scale)
+        else:
+            return None
 
     @velocities.setter
     def velocities(self, stuff):
         self._ncfile.variables['velocities'][:] = \
                     np.reshape(stuff, (self.atom, 3)) / self.velocity_scale
         self.flush()
+        self.hasvels = True
```
swails: The `velocities` setter here only gets called when *writing* NetCDF files.  So if you're reading NetCDF files, `hasvels` will *always* be `False`, even if it actually has velocities.

parmed/amber/netcdffiles.py
```diff
@@ -291,13 +295,17 @@ def cell_angles(self, stuff):
...
+            leng, ang = self.cell_lengths, self.cell_angles
+            return np.concatenate((leng, ang))
+        else:
+            return None
 
     @box.setter
     def box(self, stuff):
         self.cell_lengths = stuff[:3]
         self.cell_angles = stuff[3:]
+        self.hasbox = True
```
swails: Same comment here as in velocities.

parmed/amber/netcdffiles.py
```diff
@@ -262,14 +262,18 @@ def coordinates(self, stuff):
 
     @property
     def velocities(self):
-        vels = self._ncfile.variables['velocities'][:]
-        return (vels.reshape((-1, self.atom, 3)) * self.velocity_scale)
+        if self.hasvels:
```
swails: Replace this with:

```python
if 'velocities' in self._ncfile.variables:
```

Alternatively, you can add a `hasvels` property like so:

```python
@property
def hasvels(self):
    return 'velocities' in self._ncfile.variables
```

parmed/amber/netcdffiles.py
```diff
@@ -291,13 +295,17 @@ def cell_angles(self, stuff):
 
     @property
     def box(self):
-        leng, ang = self.cell_lengths, self.cell_angles
-        return np.concatenate((leng, ang))
+        if self.hasbox:
```
swails: `box` here is just a composite value.  Instead, put the protection I described for velocities in for `cell_lengths` and `cell_angles`.  That way this fix gets applied to those descriptors as well.

hainm:
It would be great if test is added too.

parmed/amber/netcdffiles.py
panxl: I think the first way makes things easier for now. If `hasvels ` is a property, we will probably need a setter and something like `_hasvels` since `hasvels` is set as an attribute in `open_new` method.

parmed/amber/netcdffiles.py
panxl: I think we still need to make sure `self.cell_lengths` and `self.cell_angles` are not `None` here.

swails: agreed

parmed/amber/netcdffiles.py
swails: hasvels shouldn't be an attribute in `open_new`, probably (is it even *used* anywhere?).  But I agree -- this is the easiest way to go, I think.

swails:
Thanks!

PR #994, opened 2018-07-13
Fix a regex to accommodate trailing spaces.

### LINKED ISSUE #993: Issue parsing AMBER off file
I encountered an issue when trying to load `$AMBERHOME/datt/leap/lib/phos_amino94.lib` from my AmberTools 17 install.
I get the error:

> RuntimeError: Expected atoms table not found

The issue appears to be trailing spaces in the residue name lines in the initial `!!index array str` section. As the lib file is from official sources it appears that the regex used to determine the section is too strict, or is this a design decision?

swails:
Fixes #993


swails:
run jenkins

PR #995, opened 2018-07-20
Add flag to optionally write HETATM tags in PDBs

kristofarkas:
HETATM tags in PDB can cause problems for certain use cases. For example NAMD doesn't recognise them and raises errors. Adding an option to `write_pdb` to optionally use `ATOM  ` instead of `HETATM` fixes this problem. The default is still the same as was before.

### DIFF
--- a/test/test_parmed_formats.py
+++ b/test/test_parmed_formats.py
@@ -1018,10 +1018,10 @@ ATOM      5  SG  CYX H   2      36.833  15.443  15.640  1.00 15.60           S
         """Tests HETATM/ATOM tag writing"""
         structure = Structure()
         a = Atom(name='CA', atomic_number=6)
-        structure.add_atom(a, 'ASH', 2, 'A')
-        structure.add_atom(a, 'DG', 2, 'A')
-        structure.add_atom(a, 'T', 2, 'A')
-        structure.add_atom(a, 'MOL', 2, 'A')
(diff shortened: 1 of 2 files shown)


swails:

kristofarkas:
@swails added a test

test/test_parmed_formats.py
```diff
@@ -1014,6 +1014,28 @@ def test_pdb_write_standard_names_water(self):
...
+        a = Atom(name='CA', atomic_number=6)
+        structure.add_atom(a, 'ASH', 2, 'A')
+        structure.add_atom(a, 'DG', 2, 'A')
+        structure.add_atom(a, 'T', 2, 'A')
+        structure.add_atom(a, 'MOL', 2, 'A')
+
+        coordinates = np.zeros((len(structure.atoms), 3))
+
+        structure.write_pdb("hetatom.pdb", use_hetatoms=True, coordinates=coordinates)
+        structure.write_pdb("no_hetatom.pdb", use_hetatoms=False, coordinates=coordinates)
```
hainm: you can write to StringIO to avoid writing to disk.

kristofarkas:
@hainm @swails any other comments?

swails:
Can you update the or with the latest master branch?

swails:
After that one final comment, this is good to go and I'll merge.

Thanks for bearing with me!

kristofarkas:
@swails requested changes made!

PR #997, opened 2018-08-05
Fix unused variable warning in rdparm.cpp

multiplemonomials:
I know this is pretty minor but it always bothers me when I see it in the build output.  Now `m` has a use in the python 3 version of the code so it won't trigger a warning.


swails:

swails:
I'll merge this one after #995 (which will require this one be updated, which I can do from the web interface if I'm not on mobile).

PR #1001, opened 2018-08-11
Fix logic for handling CHARMM impropers and warn if improper parameters are missing when writing OpenMM ffxml

jchodera:
When writing an OpenMM ffxml from a `CharmmParameterSet`, the logic for matching CHARMM impropers to expand generics found in residue templates into specific impropers differed from the way `CharmmParameterSet` would actually match torsions. This PR uses the same logic as in `CharmmParameterSet`, and also adds a `warning` if an improper isn't found.

This addresses [this issue](https://github.com/pandegroup/openmm/issues/2139#issuecomment-411503059) where the converted forcefield suddenly lost a lot of impropers.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -24,6 +24,9 @@ from parmed.utils.six.moves import range
 import warnings
 from parmed.exceptions import ParameterWarning, IncompatiblePatchError
 import itertools
+from itertools import product, combinations
+from ..topologyobjects import (DihedralType, ImproperType)
+
 
 from collections import OrderedDict
 
@@ -191,6 +194,10 @@ class OpenMMParameterSet(ParameterSet):
         new_params.default_scee = params.default_scee
         new_params.default_scnb = params.default_scnb
 
+        # Copy CHARMM improper type map, if present, since this is needed for matching impropers
+        if hasattr(params, '_improper_key_map'):
+            new_params._improper_key_map = new_params._improper_key_map
+
         # Add only ResidueTemplate instances (no ResidueTemplateContainers)
         # Maintain original residue ordering
         remediated_residues = list()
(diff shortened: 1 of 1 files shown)


jchodera:
This should be ready to merge!

parmed/openmm/parameters.py
```diff
@@ -358,71 +413,33 @@ def get_types(residue, atomname):
...
-                                        MATCH = True
-                                        key_placeholder = key
-                                        improper_periodic[altkeys1] = self.improper_periodic_types[key]
-                    if not MATCH:
-                        warnings.warn("No improper parameter found for {}".format(altkeys1), ParameterWarning)
+                        raise Exception('Something went wrong with improper type for {} returning an unexpected object {}'.format(key, improper))
+
+                # Warn if no improper was found
+                if not improper_found:
+                    warnings.warn('No improper found for improper {} in residue {} (types were {})'.format(impr, name, types))
```
swails: Should this have an explicit warning type other than `UserWarning` so it can be turned into a fatal error if a missing improper type is detected? What’s the use case for this not just raising an exception?

jchodera: Fixed.

jchodera:
@swails : Anything else you want me to do here?

swails:
> Do you recall what that was?

Oh, I misread what was happening.  You copied the recent changes I made to `CharmmParameterSet` over to `OpenMMParameterSet`.

I raised a PR in jchodera/ParmEd#7 to eliminate the duplication and, while I was at it, fixed #1003 (it was just a couple lines -- thanks @rafwiewiora for pinning down the problem!)

Have a look at that PR and let me know if you see anything that won't work.

jchodera:
@swails : 
```
  File "<string>", line 1, in <module>
  File "parmed/__init__.py", line 25, in <module>
    from parmed import amber, charmm, gromacs, namd, openmm, rosetta, tinker
  File "parmed/openmm/__init__.py", line 14, in <module>
    from parmed.openmm.parameters import OpenMMParameterSet
  File "parmed/openmm/parameters.py", line 54, in <module>
    class OpenMMParameterSet(ParameterSet, CharmmImproperMatchingMixin):
  File "parmed/utils/six.py", line 794, in wrapper
    return metaclass(cls.__name__, cls.__bases__, orig_vars)
TypeError: metaclass conflict: the metaclass of a derived class must be a (non-strict) subclass of the metaclasses of all its bases
```

swails:
Poop. Go ahead and remove the ABCMeta attribution to the mixin class I made.

jchodera:
Done.

PR #1005, opened 2018-08-21
Update OpenMM ffxml template hasher to not use atom names

### LINKED ISSUE #1004: Duplicate residue patches can be written when writing OpenMM ffxml files
cc: https://github.com/choderalab/openmm-forcefields/pull/66#issuecomment-414827923

When I updated the OpenMM residue template hasher in #1000, I accidentally included the atom names (which OpenMM ignores in determining whether residue templates are equivalent) in the hash.

I'll fix this in a PR, and figure out how to add a test that will catch this issue.

jchodera:
Fixes #1004.

This is still a WIP because I'm still fiddling with tests.


jchodera:
Oh, dear. It turns out many of the tests were not running because `$AMBERHOME` was not set:
https://travis-ci.org/ParmEd/ParmEd/jobs/418910206#L2080-L2086

jchodera:
Weird. Even the `lxml` install is failing:
```
    building 'lxml.etree' extension
    creating build/temp.linux-x86_64-2.7
    creating build/temp.linux-x86_64-2.7/src
    creating build/temp.linux-x86_64-2.7/src/lxml
    cc -O2 -fPIC -Wimplicit -DCYTHON_CLINE_IN_TRACEBACK=0 -I/usr/include/libxml2 -Isrc -Isrc/lxml/includes -I/home/travis/.pyenv/versions/pypy-4.0.1/include -c src/lxml/etree.c -o build/temp.linux-x86_64-2.7/src/lxml/etree.o -w
    src/lxml/etree.c:1337:3: error: unknown type name ‘PyBaseExceptionObject’
       PyBaseExceptionObject __pyx_base;
       ^
    Compile failed: command 'cc' failed with exit status 1
    creating tmp
    cc -O2 -fPIC -Wimplicit -I/usr/include/libxml2 -I/usr/include/libxml2 -c /tmp/xmlXPathInitL2BNk4.c -o tmp/xmlXPathInitL2BNk4.o
    /tmp/xmlXPathInitL2BNk4.c:2:1: warning: return type defaults to ‘int’ [-Wimplicit-int]
     main (int argc, char **argv) {
     ^
    cc tmp/xmlXPathInitL2BNk4.o -lxml2 -o a.out
    error: command 'cc' failed with exit status 1
    
    ----------------------------------------
    Command "/home/travis/.pyenv/versions/pypy-4.0.1/bin/pypy -c "import setuptools, tokenize;__file__='/tmp/pip-build-xy6Mf8/lxml/setup.py';exec(compile(getattr(tokenize, 'open', open)(__file__).read().replace('\r\n', '\n'), __file__, 'exec'))" install --record /tmp/pip-9uYe5h-record/install-record.txt --single-version-externally-managed --compile" failed with error code 1 in /tmp/pip-build-xy6Mf8/lxml
```

jchodera:
OK, we're finally seeing the Amber test failures in travis now. I'll fix these over the next day or so.

@swails : The tests do take longer now, but would it be OK to include testing with AmberTools going forward?

swails:
I raised a PR into your branch to fix the pypy issues.

jchodera:
I've run into a small issue:

The `OpenMMParameterSet.write()` method has a `write_unused` flag that defaults to `True` to ensure that residue templates can be written out separately from parameters (e.g. for small molecule conversions when `gaff.dat` has already been converted to `gaff.xml`). 

Unfortunately, the code we added to [remediate residues to comply with OpenMM requirements on parameter set import](https://github.com/ParmEd/ParmEd/blob/c67c3263eb9360f171a75c13a3ca9de974ae536a/parmed/openmm/parameters.py#L100-L137) also prunes residues for which no parameters are available. This is done to ensure we can populate the `atomic_number` field for each atom type so we can delete virtual bonds between water hydrogens.

What's the best way to fix this? 
1. Remediate residues only on write, modifying the `OpenMMParameterSet` then?
2. Drop unparameterized residues only on write?
3. Something else? Maybe there's a better way to drop the virtual water bonds?

swails:
> Remediate residues only on write, modifying the OpenMMParameterSet then?

Writing the parameter set should never have the side-effect of modifying it.  It would be OK to create a copy, modify that, write it, and then discard it after the write.  But that side-effect would be completely unexpected and a possible source of very subtle and confusing bugs.

> Something else? Maybe there's a better way to drop the virtual water bonds?
> View changes

Why not just `return True` [here](https://github.com/ParmEd/ParmEd/blob/c67c3263eb9360f171a75c13a3ca9de974ae536a/parmed/openmm/parameters.py#L123)?  That's basically equivalent to remediating residues for which you have enough information to do so, but refusing to *drop* any residues for which the necessary information is lacking.

Or is there something I'm not understanding?

jchodera:
> Why not just return True here? That's basically equivalent to remediating residues for which you have enough information to do so, but refusing to drop any residues for which the necessary information is lacking.
> Or is there something I'm not understanding?

@swails: If we did this, the converted CHARMM and AMBER forcefields would absolutely *not* work when read into OpenMM because there would be residue templates that made use of undefined types.

The behavior we want is:
* When converting a whole forcefield, drop any residue templates for which any parameters are not defined
* When converting just a residue for something like GAFF (where we suppose we have already converted the parameters), do *not* drop residue templates for which parameters are not defined

I can try to pass the `write_unused` argument to `write()` around to do that.

jchodera:
Hm, you're right---we already have `_find_unused_residues` in `write()`.

jchodera:
@swails: I could use your help fixing [this test](https://github.com/ParmEd/ParmEd/blob/c1351aed58dce16b4c2c90ede3eb90cf979de74f/test/test_parmed_openmm.py#L534-L551) since I don't believe the behavior it is testing for is valid anymore. The current behavior is that, by default, `write_unused=True` and the residue will be written even if parameters are absent. Setting `write_unused=False` will cause the residue to be dropped. Neither behavior should give a `KeyError`.

Sorry for all of these test issues cropping up at once. I hadn't realized that the Amber tests were simply not being run by travis all this time.

jchodera:
@swails : Just pinging again in case you had some advice as to what the correct behavior should be.

jchodera:
OK, I think I've fixed these issues, but now I'm getting [`numpy` import errors](https://travis-ci.org/ParmEd/ParmEd/jobs/423651588#L920-L933):
```
$ python -c "import parmed; print(parmed.__version__)"
  File "<string>", line 1, in <module>
  File "parmed/__init__.py", line 22, in <module>
    from parmed import unit, utils
  File "parmed/unit/__init__.py", line 27, in <module>
    from parmed.unit.unit import Unit, is_unit
  File "parmed/unit/unit.py", line 35, in <module>
    from parmed.utils.six import iterkeys
  File "parmed/utils/__init__.py", line 3, in <module>
    from parmed.utils.pairlist import find_atom_pairs
  File "parmed/utils/pairlist.py", line 5, in <module>
    import numpy as np
ImportError: No module named numpy
```

jchodera:
Woohoo! Tests pass! 

Should be ready to merge @swails.

jchodera:
@swails : Tests pass and this should be ready to merge!

PR #1009, opened 2018-09-02
Add atom reordering functionality for GromacsGroFile.write 

cc-a:
This patch is designed to allow a user to always write Gro files that are compatible with a gromacs topology derived from the same structure. When writing a topology reordering of atoms is sometimes performed to meet the requirement that contiguously bonded groups of atoms be written as a single molecule type. These commits provide the functionality to perform the same reordering when writing a  Gro file from the same structure. This ensures that a gro and top file suitable for further use with gromacs can be easily written.

The default behaviour of GromacsGroFile.write has not been changed. Two keyword arguments have been added. The first (match_topology) determines whether or not to perform atom reordering, default False. The second (combine) mimics the combine argument provided to GromacsTopologyFile.write as this influences reordering behaviour.

A test for this functionality using an artificially constructed Structure object has been added to test_parmed_gromacs.py.


swails:

parmed/gromacs/gromacsgro.py
```diff
@@ -221,7 +222,33 @@ def write(struct, dest, precision=3, nobox=False):
             is True, no box will be written. If False, the periodic box will be
             defined to enclose the solute with 0.5 nm clearance on all sides. If
             periodic box dimensions *are* defined, this variable has no effect.
+        match_topology : bool, optional
```
swails: What's the use case for `match_topology=False` (which is the default)?  Such a file does not seem useful to me.  I would suggest either:

1. Remove the `match_topology` argument altogether and just keep the logic for the `True` case
1. Change the default to `True`.

I'd prefer 1 because then it could simplify the code and ease maintenance.  But if there's a compelling reason, choice 2 is OK.

cc-a: The purpose of ``match_topology=False`` was mainly to preserve the default behaviour for backwards compatibility. I have no specific use cases in mind for it so happy to remove.

swails: Then I’d vote to remove the option and simplify.

cc-a:
Apologies for the long delay. Made life difficult for myself by extending the test to cover some more challenging cases and had to entirely reimplement everything I'd done to cover them. Unfortunately the code came out much more convoluted than previously so I'm open to any suggestions.

parmed/gromacs/gromacsgro.py
```diff
@@ -232,39 +254,76 @@ def write(struct, dest, precision=3, nobox=False):
...
+                                atid += 1
+                                _write_atom_line(
+                                    original_atom, atid % 100000,
+                                    resid % 100000, has_vels, dest, precision)
+                                new_molecule.add(original_atom)
+                                last_found_atom = original_atom
+                                unused_atoms.remove(original_atom)
+                                break
+                        else:
+                            raise Exception("Could not find %s" % atom)
```
swails: Never raise an `Exception`, since then it needs to catch all exceptions in order to catch this one.

`RuntimeError` is probably sensible here.

PR #1010, opened 2018-09-20
Fix up some imports and add a Link class.

swails:


swails:
@dacase -- This PR (finally!) adds support for reading and writing LINK records for PDB files.

As far as I could tell, though, PDBx/mmCIF files don't have LINK records?  At least the PDB files I was using as an example don't have LINK records in the corresponding CIF file.

Does this satisfy your use case?

dacase:
Thanks for doing this: I want to use LINK records to figure out how to add bonds between protein atoms and certain cofactors.  A common use case is this: the `FS4` residue needs to make four covalent bonds between its four Fe atoms and four CYS residues in the protein.  This information is encoded in LINK records, at least in PDB files from RCSB.

I'll give this new capability a workout and report back.  But may take a while to get to it.....

PR #1018, opened 2018-11-12
Adapt parmed to deal with residues that are simultaneously pH and redox active

vwcruzeiro:
With these modifications, one is now able to generate a new file called CPEIN that contains all the information necessary for C(pH,E)MD. This generalized format allows a given residue to be pH and redox active at the same time. This is important mainly in situations in which pH and redox effects are strongly coupled, and allows MC moves that change both number of protons and electrons. (Changes in pmemd and sander to deal with this new format are ready and should be submitted once this pull request is processed)

These modifications do not break back compatibility, as it is still possible to generate CPIN and CEIN files as before.


vwcruzeiro:
@swails : all tests are passing. May you take a look at these changes please? Thank you.

parmed/amber/titratable_residues.py
```diff
@@ -159,37 +160,33 @@ def _str_refenes(self, solvent=False, igb=2, dielc=1.0):
...
+                           range(len(self.states))]) + '\n'
+        )
         for i, atom in enumerate(self.atom_list):
             ret_str += ('%8s' % atom +
                         ''.join(['%12.4f' % (state.charges[i])
                                 for state in self.states]) + '\n'
             )
         ret_str += '-' * (8 + 12 * len(self.states)) + '\n'
-        if (self.typ=="ph"):
+        if (self.typ=="ph" or self.typ=="phredox"):
```
swails: Put spaces around the `==`. This needs to be done in lots of places.

parmed/amber/titratable_residues.py
```diff
@@ -482,7 +519,7 @@ def write_cpin(self, output, igb=2, intdiel=1.0, oldfmt=False, typ="ph", coions=
 
 AS4 = TitratableResidue('AS4', ['N', 'H', 'CA', 'HA', 'CB', 'HB2', 'HB3', 'CG',
                         'OD1', 'OD2', 'HD21', 'C', 'O', 'HD22', 'HD11', 'HD12'],
-                        pka=4.0, typ="ph")
```
swails: Where does the pka adjustment happen if you got rid of this parameter but didn’t change the reference energy?

I’m nervous that this could break cpinutil in subtle ways.

parmed/amber/titratable_residues.py
vwcruzeiro: I fixed this.

parmed/amber/titratable_residues.py
vwcruzeiro: Good question @swails . This variable pka was only used in two places: during printing when someone types "cpinutil.py --describe" or as an identifier when someone uses the minpka or maxpka options in cpinutil.py (these options were removed). As you can see in the code, the set_pKa calls are still in place, so the pka adjustment when the old format is requested is sill going to work fine. You may check my amber branch vwdc_phe-merge and you will see that the cpinutil.py tests are passing (and there are tests with the old format).

I had to remove this pka variable because when we move to the more general cases, more than one pka values might be associated to the same residue, or, with these new adaptations, multiple pka and Eo values could be associated with the same residue.

Please let me know what you think.

swails: That was already the case with Histidine. I just took the macroscopic pKa for the side-chain in general. I think those options were helpful for doing, for instance, a base-range or acid-range titration. Users didn’t need to know *which* residues were titratable at which standard pH ranges. They could just say what residues they wanted to be titratable by giving a general range of solvent conditions. 

I think those are options worth preserving.

vwcruzeiro: Ok. I will bring these features back.

vwcruzeiro:
run jenkins

swails:
Don’t worry about Jenkins for ParmEd.

vwcruzeiro:
Hello @swails . I brought the minpka and maxpka functionalities back to cpinutil.py. Thank you.

vwcruzeiro:
@swails : all tests are passing.

PR #1021, opened 2018-11-26
Fix typo in writing nbfix in GROMACS topology files

mattwthompson:
Looks like I goofed #923 

We're working on some systems and got epsilon values of 20 kJ/mol, oops


swails:
=]

hainm:
@swails may be split the very slow tests into multiple tests and run the tests in parallel.

swails:
Something was unusual in that run.  Some of the longest tests took twice as long in that environment compared to other Travis runs.

This really shouldn't be taking anywhere near 50 minutes.  The main issue with running tests in parallel is how test files are written.  I didn't know about `tempfile` when I wrote the code that managed writing files as part of tests.  What I came up with instead is not safe to run in parallel.  If you ran the tests in parallel, many would fail as they trampled over each other in the filesystem.

Ideally I should switch to something based on tempfile, but that will take development time.

mattwthompson:
Anything else you'd like (in this scope of this bugfix)? @swails

swails:
Nope, this is the next Pr I’ll merge.

PR #1022, opened 2018-11-27
allow LINKR records in PDB files

dacase:
PDB files created by refmac often have LINKR records, which is a variant of the "official" PDB LINK record.  Since refmac files are very common
(e.g. all of pdb_redo uses this program), parmed should not reject a
file as PDB format just because it has a LINKR record.  This change would be in
line with other non-standard record types that are already allowed.


swails:
To expand on this, the id_format method is only ever used in the automatic format detection routine (`parmed.load_file`). And it only processes the first few lines. 

For the actual parsing it uses the `parse` method.

swails:
I looked through the code -- the `parse` method ignores any record it doesn't recognize, so this is good to go.

PR #1028, opened 2019-01-09
Change in the addLJType function

vwcruzeiro:
Make the addLJType function work even when an atomic selection is not made. This action adds a new LJ type into the prmtop file, but doesn't change the atomic LJ mapping. This functionality is going to be necessary in some new implementations I am working on.

### DIFF
--- a/parmed/tools/addljtype.py
+++ b/parmed/tools/addljtype.py
@@ -3,21 +3,23 @@
 from math import sqrt
 from parmed.constants import NTYPES
 from parmed.utils.six.moves import range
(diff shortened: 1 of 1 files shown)


vwcruzeiro:
Good morning @swails . It seems that the test failures are unrelated to the changes in this PR.

swails:
Can you merge master?

I have to wrap my head around this (and then fix all the PDB downloading, since it looks like the PDB blacklisted Travis from submitting requests :frowning:)

swails:
All of the test cases have been fixed in the master branch.  If you merge master, then the tests should work.

PR #1031, opened 2019-02-15
Option to keep duplicate templates when writing OpenMM ffXML

tristanic:
There are cases where it makes sense for a ffXML file to contain multiple templates that will be considered identical according to OpenMM's topology-matching approach. The most obvious use-case is stereoisomers - sugars in particular become a huge problem under the default scheme. Another is metal ions of the same element but different valency (for those brave enough to actually use them).

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -23,7 +23,7 @@ from ..utils.io import genopen
 from ..utils.six import add_metaclass, string_types, iteritems
 from ..utils.six.moves import range
 import warnings
-from ..exceptions import ParameterWarning 
+from ..exceptions import ParameterWarning
 from itertools import product
 from ..topologyobjects import (DihedralType, ImproperType)
 
@@ -239,7 +239,8 @@ class OpenMMParameterSet(ParameterSet, CharmmImproperMatchingMixin):
         return new_params
 
     @needs_lxml
-    def write(self, dest, provenance=None, write_unused=True, separate_ljforce=False,
+    def write(self, dest, provenance=None, write_unused=True, skip_duplicates=True,
+              separate_ljforce=False,
               improper_dihedrals_ordering='default', charmm_imp=False):
         """ Write the parameter set to an XML file for use with OpenMM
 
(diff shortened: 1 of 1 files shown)


parmed/openmm/parameters.py
```diff
@@ -239,7 +239,8 @@ def from_parameterset(cls, params, copy=False, remediate_residues=True):
         return new_params
 
     @needs_lxml
-    def write(self, dest, provenance=None, write_unused=True, separate_ljforce=False,
+    def write(self, dest, provenance=None, write_unused=True, skip_duplicates=True,
+              separate_ljforce=False,
```
swails: Put `skip_duplicates` at the end to avoid breaking backwards compatibility.

Any arguments (except those after `*`) can be either keyword args or positional args.  If a script uses positional args, adding new arguments in the middle of the call signature will change the behavior of that function call.

That's why it's always best practice to add them at the end (unless you specify keyword-only arguments).

parmed/openmm/parameters.py
```diff
@@ -531,7 +542,7 @@ def _write_omm_atom_types(self, xml_root, skip_types):
                 etree.SubElement(xml_section, 'Type', element=str(element), **properties)
 
     @needs_lxml
-    def _write_omm_residues(self, xml_root, skip_residues, valid_patches_for_residue=None):
+    def _write_omm_residues(self, xml_root, skip_residues, skip_duplicates=False, valid_patches_for_residue=None):
```
swails: `_write_omm_residues` is a private method.  No need to supply a default for `skip_duplicates` (especially not one that differs from the `write` default which is the only place this function is currently called).

I'd just make `skip_duplicates` a required argument, since intent should be explicit here.

tristanic:
OpenMM doesn't mind if the XML file has two matching templates at load time. It only throws an error in `ForceField.createSystem() ` if a residue matches more than one template **and** the template to use hasn't been specified in the optional `residueTemplates` argument (a simple {residue index: template name} dict).

parmed/openmm/parameters.py
tristanic: See new commit (sorry if there's a better way of doing this - I'm still a bit fresh when it comes to working with GitHub).

parmed/openmm/parameters.py
tristanic: See new commit.

swails:
Thanks!

PR #1037, opened 2019-03-17
add alias read_pdb, read_cif, download_dpb, download_cif

hainm:
Hi,

I have not used parmed for awhile and here is the issue I met.
I want to use `read_PDB` function to create a Structure from StringIO (`load_file` does not offer that).
I used `read_pdb` at first and realized that it did not exists. Can we make an alias here (I guess it's no harm).

### DIFF
--- a/parmed/__init__.py
+++ b/parmed/__init__.py
@@ -32,12 +32,16 @@ from parmed.utils.decorators import deprecated as _deprecated
 load_file = formats.load_file
 read_PDB = formats.PDBFile.parse
 read_CIF = formats.CIFFile.parse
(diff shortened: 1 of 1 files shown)


swails:
API additions should clear a higher barrier than “no harm” and actually be positive. 

That said, I agree the name should have been the ones you proposed here originally. I used uppercase because pdb is an acronym, but other acronyms (like url) are traditionally lowercased in accordance with pep8 style guidelines. 

Your change introduced minimal churn/maintenance and gives a more consistent experience. So I’m +1 on this. 

Thanks!

PR #1039, opened 2019-03-23
Complete overhaul of the PDB parsing routines.

### LINKED ISSUE #1013: pdb parser doesn't correctly populate the other_locations data structure
This is a stub, pending looking at the code in detail.  But I'm posting it in case it rings some bells.

Alternate location indicators in PDB files come in (at least) two varieties:

1. Alternating "A" and "B" flags, e.g.
```
ATOM    162  N  ACYS A  12      11.777  -1.274   0.601  0.55  4.97           N
ATOM    163  N  BCYS A  12      11.503  -1.503   0.812  0.45  6.18           N
ATOM    164  CA ACYS A  12      11.811  -0.298  -0.455  0.55  4.86           C
ATOM    165  CA BCYS A  12      11.623  -0.484  -0.262  0.45  5.73           C
ATOM    166  C  ACYS A  12      10.387   0.251  -0.721  0.55  5.05           C
ATOM    167  C  BCYS A  12      10.271   0.199  -0.506  0.45  5.40           C
ATOM    168  O  ACYS A  12       9.398  -0.462  -0.917  0.55  4.07           O
ATOM    169  O  BCYS A  12       9.280  -0.529  -0.459  0.45  4.20           O
ATOM    170  CB ACYS A  12      12.294  -0.762  -1.828  0.55  4.66           C
ATOM    171  CB BCYS A  12      12.078  -1.246  -1.516  0.45  6.36           C
ATOM    172  SG ACYS A  12      13.992  -0.380  -2.264  0.55  5.98           S
ATOM    173  SG BCYS A  12      13.700  -2.026  -1.379  0.45  8.67           S
```

2. All "A" atoms before all "B" atoms:

```
ATOM    162  N  ACYS A  12      11.777  -1.274   0.601  0.55  4.97           N
ATOM    164  CA ACYS A  12      11.811  -0.298  -0.455  0.55  4.86           C
ATOM    166  C  ACYS A  12      10.387   0.251  -0.721  0.55  5.05           C
ATOM    168  O  ACYS A  12       9.398  -0.462  -0.917  0.55  4.07           O
ATOM    170  CB ACYS A  12      12.294  -0.762  -1.828  0.55  4.66           C
ATOM    172  SG ACYS A  12      13.992  -0.380  -2.264  0.55  5.98           S
ATOM    163  N  BCYS A  12      11.503  -1.503   0.812  0.45  6.18           N
ATOM    165  CA BCYS A  12      11.623  -0.484  -0.262  0.45  5.73           C
ATOM    167  C  BCYS A  12      10.271   0.199  -0.506  0.45  5.40           C
ATOM    169  O  BCYS A  12       9.280  -0.529  -0.459  0.45  4.20           O
ATOM    171  CB BCYS A  12      12.078  -1.246  -1.516  0.45  6.36           C
ATOM    173  SG BCYS A  12      13.700  -2.026  -1.379  0.45  8.67           S
```
(see, e.g., the `1gwe` PDB entry, where each of the above orders appears at different points in the file.)

We are using code roughly like this:  first create a Structure via `parm = parmed.load_file(pdb_fn)`.  Next, we have a loop like this:
```
  holder = {}
  for atom in parm.atoms:
    if atom.other_locations:
      holder[atom.residue.idx+1] = len(atom.other_locations)
```

However, the `atom.other_locations` information is correct for alternative 1 above, but not for alternative 2.  (For alternative 2, the code doesn't find any atoms with "other_locations".)

...more later as a study what is going on.  But maybe a fix (or a statement that I'm confused) will be obvious to Jason or Hai.....

swails:
It now essentially discards any efforts made to conform to the PDB
standard while parsing PDB files that very clearly break that standard
(e.g., by overflowing the atom and residue index fields).

No attempt is made anymore to decode differently-encoded residue or atom
numbers (all attempts were always sequential, anyway).

Nobody should be negatively impacted by this change, and it dramatically
simplifies the routine and fixes other deficiencies. That is, it is now
a more-compliant parser than it was for PDB files that obey the
standard.

Now it's time to spend a bajillion years debugging :-/

Fixes #1013

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -4,7 +4,7 @@ PDBx/mmCIF files.
 """
 from __future__ import division, print_function, absolute_import
 
-from collections import OrderedDict
+from collections import OrderedDict, namedtuple
 from contextlib import closing
 try:
     from string import ascii_letters
@@ -29,6 +29,8 @@ import re
 import warnings
 
 #++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+
+_ascii_letters_set = set(ascii_letters)
 
 def _compare_atoms(old_atom, new_atom, resname, resid, chain, segid, inscode):
     """
@@ -104,6 +106,11 @@ def _number_truncated_to_n_digits(num, digits):
 class PDBFile(object):
     """ Standard PDB file format parser and writer """
     #===================================================
+
+    AtomLookupKey = namedtuple(
+        'AtomLookupKey', ('name', 'number', 'residue_name', 'residue_number', 'chain',
+                          'insertion_code', 'segment_id', 'alternate_location')
+    )
 
     @staticmethod
     def id_format(filename):
@@ -205,6 +212,7 @@ class PDBFile(object):
         self._model_open = True
         self._insertion_codes = set() # Only permitted for compliant PDB files
         self._last_atom = None
+        self._last_residue = None
         # Some writers use additional fields to hold extra digits for the residue number if it goes
         # more than 4 digits
         self._residue_number_field_extended_by = 0
@@ -595,18 +603,18 @@ class PDBFile(object):
         except ValueError:
             return self._last_atom.number + 1 if self._last_atom is not None else 1
 
-    @staticmethod
-    def _make_atom_key_from_parts(atom_parts, all_parts=False):
+    @classmethod
+    def _make_atom_key_from_parts(cls, atom_parts, all_parts=False):
         """ If all_parts is False, only key from the values that determine alternate locations """
-        return (
-            atom_parts['name'],
-            atom_parts['number'] if all_parts else None,
-            atom_parts['residue_name'],
-            atom_parts['residue_number'],
-            atom_parts['chain'],
-            atom_parts['insertion_code'],
-            atom_parts.get('segment_id', ''),
-            atom_parts['alternate_location'] if all_parts else None,
+        return cls.AtomLookupKey(
+            name=atom_parts['name'],
+            number=atom_parts['number'] if all_parts else None,
+            residue_name=atom_parts['residue_name'],
+            residue_number=atom_parts['residue_number'],
+            chain=atom_parts['chain'],
+            insertion_code=atom_parts['insertion_code'],
+            segment_id=atom_parts.get('segment_id', ''),
+            alternate_location=atom_parts['alternate_location'] if all_parts else None,
         )
 
     def _parse_atom_record(self, line):
@@ -626,7 +634,7 @@ class PDBFile(object):
         attribute_key = self._make_atom_key_from_parts(atom_parts)
         all_attribute_key = self._make_atom_key_from_parts(atom_parts, all_parts=True)
         current_atom = self._atom_map_from_attributes.get(attribute_key, None)
-        if (current_atom is not None and atom_parts['alternate_location'] in ascii_letters and
+        if (current_atom is not None and atom_parts['alternate_location'] in _ascii_letters_set and
             not self._atom_indices_overflow and not self._residue_indices_overflow):
             if self._current_model_number == 1:
                 current_atom.other_locations[atom_parts['alternate_location']] = atom
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -200,6 +200,7 @@ class PDBFile(object):
         self._model_atom_counts = [0]
         self._anisou_records = dict()
         self._atom_map_from_atom_number = dict()
+        self._model1_atoms_in_structure = set()
         self._model_open = True
         self._insertion_codes = set() # Only permitted for compliant PDB files
         # Some writers use insertion code to hold an extra digit for the residue number if it
@@ -437,7 +438,7 @@ class PDBFile(object):
         self.struct.keywords += '%s,' % line[10:].strip()
 
     def _parse_ter_record(self, *args):
-        next(reversed(self._atom_map_from_attributes)).residue.ter = True
+        self._find_last_atom_read().residue.ter = True
 
     def _parse_experimental(self, line):
         if self.struct.experimental:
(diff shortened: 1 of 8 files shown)


dacase:
> Prevent altloc atoms from being added to the main Structure 

Not sure what the "main Structure" is here: It looks like the `other_locations` data structure (which is what I use) is unchanged, and that `all_attributes` (which I don't use) is modified.  Is that correct?

Basically, if you think that the `other_locations` info will be changed with this PR, let me know, and I'll spend more time looking at/testing the code to see its consequences.

...thx

swails:
> Not sure what the "main Structure" is here: It looks like the `other_locations` data structure (which is what I use) is unchanged, and that `all_attributes` (which I don't use) is modified. Is that correct?

"main Structure" in this case refers to the `Structure.atoms` data structure.  This comment was about fixing a regression in the new parser, not about changing behavior.

The intention of this refactoring is that it should yield the same behavior in every case that people care about.  It will also make the parser more compliant for cases where atomic attributes (like anisotropy tensors and alternate locations) don't appear consecutively in the PDB file.

The notable places where it will differ all involve reading non-compliant PDB files -- specifically those used to represent more than 99999 atoms or 9999 residues.  In those cases, the old parser would attempt to convert/remember the atom/residue numbers regardless of how the file attempted to handle overflows.  This allowed "proper" support of `ANISOU` records for these files, alternate locations, `CONECT` records, etc.  PDB parser bugs/improvements were getting increasingly difficult to make, so it was ultimately easier to fix #1013 by refactoring the entire PDB parsing routine than somehow adding support to the existing parser without breaking things.

The new parser abandons efforts to retain anything outside of the atom name, residue name, and coordinates from `ATOM` lines for any non-compliant PDB file (but continues to process as much information as the old parser for compliant PDBs).

My strong suspicion is that this won't impact anybody.  For anybody it does impact, they can use ParmEd 3.1.0 until they migrate their workflows over to use PDBx/mmCIF files.

dacase:
Thanks for the explanation!

> For anybody it does impact, they can use ParmEd 3.1.0 until they migrate their workflows over to use PDBx/mmCIF files.

I wish I had one of your neat animated gif's to respond to the notion that legacy PDB formats will disappear before climate change kills us all....

swails:
temporary

swails:
@dacase -- I plan on merging this unless you'd rather I hold off.  This should fix #1013 as well.

swails:
I'm also curious to see if it resolves #1038.  CHARMM uses the extra blank space after the residue name so it can support 4-letter residue names, and since no other dialect uses that spot for anything I didn't see the harm in always reading in that character (and stripping off the trailing space).

dacase:
> I'm also curious to see if it resolves #1038. CHARMM uses the extra blank space after the residue name so it can support 4-letter residue names, and since no other dialect uses that spot for anything I didn't see the harm in always reading in that character (and stripping off the trailing space).

Go ahead and merge: I'll see if any of my problems pop up again.  (Were you stripping the any trailing space before?)

p.s. I think I've figured what was probably meant in the python statements that issued warnings about bad ANISOU cards.  Maybe this is already fixed in the new rework; otherwise I'll chat with you about this at Tampa.

dacase:
@swails  Do you have an opinion about whether #1039  should be included in the upcoming AmberTools19 distribution?

swails:
> p.s. I think I've figured what was probably meant in the python statements that issued warnings about bad ANISOU cards. Maybe this is already fixed in the new rework; otherwise I'll chat with you about this at Tampa.

There are two possible things that could go wrong with processing ANISOU records:

1. corrupt ANISOU record where the anisotropic factors are either not in the right columns or not numeric and casting to an integer fails. [This results in a warning and that ANISOU record is ignored](https://github.com/ParmEd/ParmEd/blob/master/parmed/formats/pdb.py#L590)
1. The atom referenced by the ANISOU record doesn't match up with any atom in the PDB file. [This results in an warning where the missing set of atom attributes are printed to the screen](https://github.com/ParmEd/ParmEd/blob/master/parmed/formats/pdb.py#L769). The PDB is still processed, and ANISOU records for non-existing atoms are ignored and dropped.

swails:
> @swails Do you have an opinion about whether #1039 should be included in the upcoming AmberTools19 distribution?

I go back and forth.  If you think you (or someone else) will be able to exercise it sufficiently over the next few months, it may be worth releasing it as a bugfix later.  It's the main parsing engine behind `pdb4amber` if I'm not mistaken, which means the potential impact is significant (? - how popular is pdb4amber these days?)

That said, there is a pretty big array of tests in ParmEd that exercises basically all of the common use-cases pretty extensively on over a dozen PDB files (and each bug in the old parser spawned a new test to prevent regressions).

Another thing we can do is run the new parser and old parser through a large sample of PDBs from RCSB as validation (perhaps comparing the before- and after- outputs from pdb4amber applied to each of them).  That'll give us more confidence.

If we're not going to get much more testing/exercise from the ParmEd PDB parser, I'd just as soon include it.

We can also punt on this decision until we discuss in Tampa.

dacase:
> how popular is pdb4amber these days?

 Very widely used, I think.  We should be pushing more Amber users to routinely run input pdb files through pdb4amber before going further.  And pdb4amber should get better about identifying what is in PDB files, relying less on atom and residue "names" in the files, and more on what the structures actually are.  I have a slide on this in my Tampa presentation.

Of course, the ideal above could be implemented somewhere else besides pdb4amber, but it (pdb4amber) seems like a logical place for this.  Phenix makes heavy use of pab4amber in its work, and we've run more than 15,000 PDB files through it. (Obviously, that was before #1039.  )

dacase:
About warnings in the PDB parser, even after this change:

Some warnings look like this:
```
warnings.warn('Corrupt link line: %s' % line, PDBWarning)
```

Which is helpful in telling the user which CONECT record it doesn't like.

But the ANISOU warnings all look like this:
```
warnings.warn('ANISOU record does not match previous '
                                      'atom', PDBWarning)
```
There  are two problems here: first is that there is no info about which ANISOU record is faulty; but second, it doesn't even print what it looks like it ought to print: the string "atom" is repeated twice in the output.  Is it OK if I prepare suggestions for better warning messages?

swails:
> But the ANISOU warnings all look like this:

That's the old parser.  The new one will print out the atom lookup key which will include enough information to narrow down the line (atom name, residue name, atom number, residue number, etc.).  The new parser does, however, print no information that will help you identify the offending line if it fails to convert one of the ANISOU values into a float.  But if it fails to match an atom, debugging information will be printed.

That said, I added debugging information to #1040 -- you can see the changes in `pdb.py` in that PR to see what I'm talking about.  This should hopefully suffice for users that like to dig into problems.

hainm:
>> how popular is pdb4amber these days?

> Very widely used, I think. 

> Phenix makes heavy use of pab4amber in its work, and we've run more than 15,000 PDB files through it.

and thus `parmed` is very widely used under the hood. ;)

PR #1041, opened 2019-03-27
Break when parsing unsupported lonepairs

Eigenstate:
Even though the residue would be marked to skip, upon encountering the COLINEAR lonepair type the parser tries to keep going and fails.

Now top_all36_cgenff.rtf can be parsed for chamber actions.


swails:

swails:
:smiley:

PR #1044, opened 2019-03-31
Take absolute value of CHARMM epsilons, since those are always negative

### LINKED ISSUE #1012: Flip sign for epsilon values read from CHARMM .prm files
I believe the sign for epsilon values read from CHARMM .prm is incorrect. The parameter file specifies negative epsilon values ([see here for example](https://www.ks.uiuc.edu/Training/Tutorials/namd/namd-tutorial-unix-html/node25.html)), causing flipped values to be written upon conversion.

swails:
Fixes #1012


bradakstx:
Just passing through the PR list and saw this. If I recall the historical perspective correctly (maybe I'm making this up?):

A negative epsilon was supposed to flag geometric mixing (since the product of two negative numbers is positive) whereas a positive epsilon was supposed to flag arithmetic mixing. I have no idea if this was ever actually supported beyond CHARMM itself (or someone's local version of it). NAMD, for example, ignored this convention outright and had its own global flag for mixing rules.

PR #1047, opened 2019-04-08
Add RAN and DTN to nucleic residues

dwhswenson:
I'm using ParmEd's residue names as a basis for adding support for recognizing nucleic acids to MDTraj's atom selection language. In the process, I found that it looks like you're missing a couple. Here they are.

By the way -- I'm curious about some of the more obscure names you have (e.g., YYG for RNA guanine). Is there a canonical list for these somewhere?

### DIFF
--- a/parmed/residue.py
+++ b/parmed/residue.py
@@ -269,14 +269,14 @@ class RNAResidue(DNAResidue):
 DG = DNAResidue('Guanine', 'DG', ['GUA', 'DG5', 'DG3', 'DGN'])
 DC = DNAResidue('Cytosine', 'DC', ['CYT', 'DC5', 'DC3', 'DCN', 'DCP'])
 DA = DNAResidue('Adenine', 'DA', ['ADE', 'DA5', 'DA3', 'DAN', 'DAP'])
-DT = DNAResidue('Thymine', 'DT', ['THY', 'DT5', 'DT3'])
+DT = DNAResidue('Thymine', 'DT', ['THY', 'DT5', 'DT3', 'DTN'])
 G = RNAResidue('Guanine', 'G', ['GUA', 'G5', 'G3', 'GN', 'RG', 'RG3', 'RG5',
                                 'RGN', 'GF2', 'M2G', 'YYG', '7MG', 'OMG',
                                 '2MG',])
 C = RNAResidue('Cytosine', 'C', ['CYT', 'CP', 'C5', 'C3', 'CN', 'RC', 'RC5',
                                  'RC3', 'RCN', 'CFZ', '5MC', 'OMC',])
 A = RNAResidue('Adenine', 'A', ['ADE', 'AP', 'A5', 'A3', 'AN',
-                                'RA', 'RA3', 'RA5', 'AF2', '1MA'])
+                                'RA', 'RA3', 'RA5', 'RAN', 'AF2', '1MA'])
 U = RNAResidue('Uracil', 'U', ['URA', 'U3', 'U5', 'UN', 'RU', 'RU3', 'RU5',
                                'RUN', 'UFT', '5MU', 'H2U', 'PSU',])
 T = RNAResidue('Thymine', 'T', ['THY', 'T3', 'T5', 'TN',


swails:

dacase:
I think this is likely to be a bad idea (TM) in the long run.  DTN is pdb-lingo is dithionite, and RAN
has nothing to do with RNA.  I think you may be seeing some seldom-used non-standard residue names from an Amber library(?). Amber has been better than some competitors, but by no means perfect, in trying to follow "correct" PDB nomenclature.

The "canonical" description of names is the components.cif file, available at http://www.wwpdb.org/data/ccd  This says that YYG is "4-(3-[5-O-PHOSPHONORIBOFURANOSYL]-4,6-DIMETHYL-8-OXO-4,8-DIHYDRO-3H-1,3,4,5,7A-PENTAAZA-S-INDACEN-YLAMINO-BUTYRIC ACID METHYL ESTER" aka
"MODIFIED GUANOSINE-5'-PHOSPHATE".  But I'm not sure what the modification is here.  And there doesn't seem to "standard" names for other 5'-phosphate residues.....

dwhswenson:
There's still a consistency problem, though. ParmEd does not include the non-canonical CCD-conflicting names DTN and RAN. But it does include other analogous conflicting names [DAN](https://www.ebi.ac.uk/pdbe-srv/pdbechem/chemicalCompound/show/DAN), [DGN](https://www.ebi.ac.uk/pdbe-srv/pdbechem/chemicalCompound/show/DGN), [DCN](https://www.ebi.ac.uk/pdbe-srv/pdbechem/chemicalCompound/show/DCN), and [RUN](https://www.ebi.ac.uk/pdbe-srv/pdbechem/chemicalCompound/show/RUN), as well as the remaining non-conflicting variants on the same naming style. Not to mention the various D\*5/D\*3 and RNA analogues, some of which have conflicts. (My favorite is [this PDB entry](http://www.rcsb.org/structure/367D) which has CCD DA5 complexed to DNA -- fortunately, not with a 5' adenine!)

Is the better solution to remove all the residue names that conflict with CCD entries?

swails:
I'm torn here.  I want to do what's the most *useful* moreso than what's most correct.

The main issues with mischaracterizing residues as RNA or DNA when it's some other component are:

1. PDB files get written with `ATOM` records instead of `HETATM`
1. Automatic bond perception may add bonds that are not there (although I think existing bonds will get properly added assuming they are close enough for ParmEd to consider them 'bonded').

These are mostly issues with PDB file reading and writing, and neither is going to be overly serious given current functionality.  However, this may limit ParmEd's utility to bioinformatics applications.

It's probably best to just remove non-standard names utilized by Amber except for those very commonly used still (e.g., ASH, GLH, CYX, and the like).

hainm:
may be adding an API to define a residue as `standard`?

swails:
> may be adding an API to define a residue as `standard`?

Another API that nobody will use? :smiley:

And there is an API -- edit the standard residue lists and maps directly.

dacase:
> Is the better solution to remove all the residue names that conflict with CCD entries?

Not in my view.  It's probably best to keep conflicting names that are in common usage.  But none of the specific residues mentioned here fall into that category: `xxN` refer to isolated molecules, not to residues that could be part of an RNA or DNA chain.  These would be probably be rarely used, so the balance would move toward leaving them out of the parmed list.  They would certainly be `HETATM` records, if that makes a difference.

Of course, Jason gets to decide here....

dwhswenson:
I'm mainly aiming to improve the nucleic acid side of things here. There are two types of mischaracterization to address:

* residue names that come from non-CCD sources and may represent something completely different in the CCD (e.g., `DAN`)
* residue names that are correct from the CCD, but are not really equivalent to the standard nucleic residues (e.g., `YYG`)

For the first problem, which ones do you see as widely-enough used to keep? Existing aliases like this mainly fall into the categories `xxN`, `xx5`, `xx3`, `DxP`, and names like `GUA`, `THY`, etc.

For the second one, if the main purposes are writing PDBs or recognizing bonds, I expect these should be excluded. For PDB writing, these *should* be HETATM entries, right? And I'm not sure how much it helps you to make the bonds correctly if you say that [YYG](https://www.ebi.ac.uk/pdbe-srv/pdbechem/chemicalCompound/show/YYG) is guanosine.

On the other hand, if you're using these to identify which residues (could) belong to nucleic acid chains, you might want these (and perhaps a lot more). I wrote a little notebook that extracts and organizes all the [RD]NA linking molecules from the CCD (and discusses this issue more). I think this creates the canonical list I asked about, although the list is a lot longer than I expected. The notebook is [in a gist](https://gist.github.com/dwhswenson/68c994d1c8aed6d0ce91c0f11b7ba3ef), if you're curious.

PR #1054, opened 2019-05-20
Fix DeprecationWarning when importing ABCs

mkcor:
I'm suggesting this change because I'm getting the following
```sh
<...>/venv/lib/python3.7/site-packages/parmed/amber/parameters.py:10: DeprecationWarning:
  
  Using or importing the ABCs from 'collections' instead of from 'collections.abc' is deprecated, and in 3.8 it will stop working
```
whenever I'm running our test suite over at https://github.com/plotly/dash-bio ;)

### DIFF
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -7,7 +7,12 @@ Date: Aug. 11, 2015
 """
 from __future__ import division, print_function
 
-from collections import defaultdict, Sequence
+import six
+if six.PY2:
+    from collections import Sequence
+else:
+    from collections.abc import Sequence
+from collections import defaultdict
 from contextlib import closing
 import math
 import os
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -7,15 +7,12 @@ Date: Aug. 11, 2015
 """
 from __future__ import division, print_function
 
-import six
-if six.PY2:
-    from collections import Sequence
-else:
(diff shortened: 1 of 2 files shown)


swails:

swails:
You don’t need to require `six` — it’s already bundled inside `parmed.utils`. 

Just use that one instead of adding a dependency.

mkcor:
@swails aaah, I was wondering where you could be hiding all of your dependencies... done.

mkcor:
Oh, timeout error: I probably cannot rerun the CI workflow myself (exploring the Travis CI interface...), right?

swails:
> @swails aaah, I was wondering where you could be hiding all of your dependencies... done.

Thanks!  I had actually commented the day you submitted the PR, but I never submitted the review (I was using the mobile interface, which is far less user-friendly than the computer).  Sorry for the appearance of being ignored.

All of the hard dependencies _except numpy_ are bundled in `parmed.utils` (just `six` and `netcdf` from scipy, I think).

I should probably put `numpy` in the `setup.py` file the way you had put `six`...

All of the other dependencies are optional.  If you try to use functionality that requires a package you don't have installed, you'll get hit with an `ImportError` (like building an OpenMM `System` without having OpenMM installed).

> Oh, timeout error: I probably cannot rerun the CI workflow myself (exploring the Travis CI interface...), right?

I restarted the one build that failed.

Thanks again for the PR!

PR #1072, opened 2019-11-25
Fall back to stdlib xml if lxml not available

tristanic:
Allows writing of FFXML files to go ahead using the standard Python XML library if lxml is not installed. Probably slightly slower for very big forcefields, but the result is functionally and cosmetically essentially identical apart from the insertion of `<?xml version="1.0" ?>` as the first line (which could easily be stripped, of course - but it appears harmless).

Note: this seems to make the @needs_lxml decorator superfluous.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -30,10 +30,25 @@ from ..topologyobjects import (DihedralType, ImproperType)
 
 from collections import OrderedDict
 
+_have_lxml = False
 try:
     from lxml import etree
+    _have_lxml = True
 except ImportError:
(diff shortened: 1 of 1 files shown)


tristanic:
> It’s pretty easy to install lxml. Do you or others have the trouble? Cheers.

To expand on what @swails said, this is to make it easier to use ParmEd in the scope of a bigger distribution like Phenix, which tries to keep its dependencies to a minimum for maintainability. Not (directly) for my own use, but for the end user who doesn't necessarily know (nor want to know) how to add `lxml` to their local installation.

Issue #1080, opened 2020-03-02
CHARMM Drude force field


peastman:
I'd like to use ParmEd to convert the CHARMM Drude force field to OpenMM.  It uses a lot of features not found in standard Amber force fields:

- Drude particles
- Lone pairs
- NBFIX terms
- Urey-Bradley terms
- CMAP terms

I'm not sure which of these are currently supported by ParmEd and which aren't.  What would be involved in adding support for them?

peastman:
It also uses a harmonic form for improper torsions.

peastman:
Can anyone offer advice on this?  I'm willing to help implement the needed features, I just don't know what needs to be done or how it should be approached.

swails:
Sorry, I never saw the email about this.

Drude particles and lone pairs are probably not in ParmEd yet (although it has some concept of virtual sites, which may suffice).  The rest are there.

You can look at `topologyobjects.py`, which contains all of these entities.

You'd basically need to add the parameter types for those new terms and then update the CHARMM parameter file parsers to read those sections and add those terms into the parameterset dict.

peastman:
Thanks!  I'll spend a while looking over the code, then come back with more specific questions.

The `ExtraPoint` class should mostly work for lone pairs, except that CHARMM supports more complicated methods of specifying positions (https://www.charmm.org/charmm/documentation/by-version/c40b1/params/doc/lonepair/).  For Drude particles we basically need to store a bunch of other properties for certain atoms

peastman:
For each polar atom, the force field defines the following attributes that we'll need to store:

- Alpha, the polarizability.  The presence of this attribute is what signals to add a Drude particle.
- The thole damping factor.  This is optional, but if it's missing that just means to use the default value of 1.3.
- The atom type to use for the Drude particle.  This is optional, but if it's missing that just means to use the default type DRUD.

In addition, some atoms have anisotropic polarizability.  For those, we need to store references to three other atoms that are used to define the coordinate frame, plus two scale factors.

We could add some of this information to the `Atom` class, but it might be cleaner to put all of it in a separate `Drude` class?  Or perhaps the simple parameters should go in `Atom`, but the coordinate frame definition would go in an `AnisotropicDrude` class?

I looked at the classes for AMOEBA, since those need to store very similar information, but they look incomplete.  `MultipoleFrame` stores a set of indices rather than atoms?  And it looks like they never actually get created or associated with an atom.

[...]

PR #1082, opened 2020-03-20
Add support for Drude particles

peastman:
Initial implementation of #1080.  I've added support for loading Drude particles from CHARMM files and writing them to OpenMM force fields.  I'm still working on testing and making some required changes to OpenMM, but this is basically complete and functional.

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -110,6 +110,7 @@ class OpenMMParameterSet(ParameterSet, CharmmImproperMatchingMixin):
         super(OpenMMParameterSet, self).__init__()
         if filenames:
             raise NotImplementedError('Cannot yet read OpenMM Parameter sets')
+        self.unique_atom_types = False
 
     @classmethod
     def _remediate_residue_template(cls, params, residue):
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -256,6 +256,16 @@ class OpenMMParameterSet(ParameterSet, CharmmImproperMatchingMixin):
         if (n_discarded_patches > 0):
             warnings.warn('{} patches discarded, {} retained'.format(n_discarded_patches, len(new_params.patches)))
 
+        # CHARMM Drude force field lists all lone pairs as being hydrogens???  Fix them.
+        lp_types = set()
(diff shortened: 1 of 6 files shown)


peastman:
If the changes look ok, this should be ready to merge.  To work correctly with OpenMM, it needs the changes from https://github.com/openmm/openmm/pull/2623.  That's why a few test cases are failing.

parmed/openmm/parameters.py
```diff
@@ -213,6 +224,7 @@ def from_parameterset(cls, params, copy=False, remediate_residues=True):
         new_params._combining_rule = params.combining_rule
         new_params.default_scee = params.default_scee
         new_params.default_scnb = params.default_scnb
+        new_params.unique_atom_types = unique_atom_types
```
swails, line 227: `unique_atom_types` is only set in this constructor, but is used like it's always present below (unless I'm missing something).  It should probably be set to `False` in `__init__` as well to make sure the attribute is always present.

peastman:
Is it ok if I have it use the latest dev build of OpenMM?  After the next release comes out, we can switch it back.

swails:
Latest dev is fine in general.

There were issues with one version awhile ago which is why I switched it to
a pinned version.

But things are stable enough these days that we can just use the dev build.

peastman:
All of the Python 3 builds are passing.  OpenMM no longer supports Python 2, so that build is failing.  Are you still trying to support it, or should I just remove it from the test matrix?

swails:
>  Are you still trying to support it

I'm still trying to support it for now. There are still some straggler platforms out there that need to use ParmEd that are still on Python 2.7 (the Phenix refinement package is the major one, I think).

This support will likely end after AmberTools 21 is released next year.  But OpenMM is an optional dependency, so I'm fine just making the Python 2.7 build a minimal build that doesn't test with all of the optional dependencies.

peastman:
> I'm fine just making the Python 2.7 build a minimal build that doesn't test with all of the optional dependencies.

Can I leave it to you to make whatever changes you want to the travis files?

swails:
Done.

peastman:
Thanks!  All tests are passing now.

swails:
Done, thanks!

PR #1087, opened 2020-05-05
Optionally don't de-duplicate matching atom types in OpenMM conversion

mattwthompson:
In https://github.com/openforcefield/openforcefield/issues/580 (well, everywhere using the SMIRNOFF specification, but it's not always fatal), uniqueness of valence terms based on the atom types they are built off of is not guaranteed. This is not _usually_ a problem since you typically only need the relevant parameters and atom types as part of the valence term (i.e. the length and force constant of a bond and the atom indices - but not atom types! - of the two atoms involved). But some of the bookkeepping does rely on that; we were encountering an issue with atom types being stored as dictionary keys. One approach would be to change that behavior, but this sort of thing is probably going to crop up elsewhere. A not-very-performant but catch-all solution is to make each atom have its own atom type object.

### DIFF
--- a/parmed/openmm/topsystem.py
+++ b/parmed/openmm/topsystem.py
@@ -27,7 +27,7 @@ __all__ = ['load_topology']
 
 
 @needs_openmm
-def load_topology(topology, system=None, xyz=None, box=None, unique_atom_types=True):
+def load_topology(topology, system=None, xyz=None, box=None, condense_atom_types=True):
     """
     Creates a :class:`parmed.structure.Structure` instance from an OpenMM
     Topology, optionally filling in parameters from a System
@@ -47,10 +47,10 @@ def load_topology(topology, system=None, xyz=None, box=None, unique_atom_types=T
         information unless ``box`` (below) is also specified
     box : array of 6 floats
         Unit cell dimensions
-    unique_atom_Types : bool, default=True
-        If True, create unique atom types based on de-duplcating properties. If False,
+    condense_atom_types : bool, default=True
+        If True, create unique atom types based on de-duplicating properties. If False,
         create one atom type for each atom in the system, even if its properties match
-        an existin atom type.
+        an existing atom type.
 
     Returns
     -------
@@ -183,7 +183,7 @@ def load_topology(topology, system=None, xyz=None, box=None, unique_atom_types=T
         elif isinstance(force, mm.CMAPTorsionForce):
             _process_cmap(struct, force)
         elif isinstance(force, mm.NonbondedForce):
-            _process_nonbonded(struct, force, unique_atom_types)
+            _process_nonbonded(struct, force, condense_atom_types)
         elif isinstance(force, ignored_forces):
             continue
         else:
@@ -376,7 +376,7 @@ def _process_cmap(struct, force):
             struct.cmap_types.append(cmap_type)
     struct.cmap_types.claim()
 
-def _process_nonbonded(struct, force, unique_atom_types):
+def _process_nonbonded(struct, force, condense_atom_types):
     """ Adds nonbonded parameters to the structure """
     typemap = dict()
     element_typemap = defaultdict(int)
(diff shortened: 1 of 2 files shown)


parmed/openmm/topsystem.py
```diff
@@ -47,6 +47,10 @@ def load_topology(topology, system=None, xyz=None, box=None):
         information unless ``box`` (below) is also specified
     box : array of 6 floats
         Unit cell dimensions
+    unique_atom_Types : bool, default=True
+        If True, create unique atom types based on de-duplcating properties. If False,
+        create one atom type for each atom in the system, even if its properties match
+        an existin atom type.
```
j-wags, line 53:
```suggestion
        an existing atom type.
```

parmed/openmm/topsystem.py
```diff
@@ -47,6 +47,10 @@ def load_topology(topology, system=None, xyz=None, box=None):
         information unless ``box`` (below) is also specified
     box : array of 6 floats
         Unit cell dimensions
+    unique_atom_Types : bool, default=True
+        If True, create unique atom types based on de-duplcating properties. If False,
```
swails, line 51: de-deuplcating -> de-duplicating

parmed/openmm/topsystem.py
```diff
@@ -47,6 +47,10 @@ def load_topology(topology, system=None, xyz=None, box=None):
         information unless ``box`` (below) is also specified
     box : array of 6 floats
         Unit cell dimensions
+    unique_atom_Types : bool, default=True
```
swails, line 50: `unique_atom_Types` -> `unique_atom_types` (capitalization).

I'll admit to being confused by the behavior here.  I read `unique_atom_types` as "assign all atom types to be unique", when what this really means is that atom types should be condensed into as few atom types as possible.

I think this parameter is better named `condense_atom_types=True` to be more reflective of what it really does.

swails:
It would also be good to have a test case for this scenario.

mattwthompson:
One day I'll submit patches without typos 😅

I've implemented the above suggestions and added a unit test.

One thing I may wish to change in the future is a sort of partial collapsing that resolves the above issue but doesn't carry atom types that are duplicated in all non-bonded and bonded terms. But that will be a little trickier to do, so unless it's needed down the line, i.e. if these redundant atom types cause memory issues, this should satisfy our needs.

test/test_parmed_openmm.py
```diff
@@ -93,6 +93,22 @@ def test_load_topology(self):
...
+        structure_condensed = openmm.load_topology(ommparm.topology, system, condense_atom_types=True)
+        structure_uncondensed = openmm.load_topology(ommparm.topology, system, condense_atom_types=False)
+
+        num_unique_atom_types_condensed = len(set([a.atom_type for a in structure_condensed]))
+        num_unique_atom_types_uncondensed = len(set([a.atom_type for a in structure_uncondensed]))
+
+        assert num_unique_atom_types_uncondensed > num_unique_atom_types_condensed
+
+        # This may change if this or a similar flag does a partial condensing, see PR #1087
+        assert num_unique_atom_types_uncondensed == len([*structure_uncondensed.atoms])
```
swails, line 110: This is causing a SyntaxError in the pypy test:

```
  File "/home/travis/.pyenv/versions/pypy2.7-6.0.0/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/.pyenv/versions/pypy2.7-6.0.0/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/.pyenv/versions/pypy2.7-6.0.0/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/ParmEd/ParmEd/test/test_parmed_openmm.py", line 110
    assert num_unique_atom_types_uncondensed == len([*structure_uncondensed.atoms])
```

Why not just use `len(structure_uncondensed.atoms)`?  It's a `list`-derived type.

PR #1091, opened 2020-06-03
More fixes to support CHARMM Drude force field

peastman:


parmed/openmm/parameters.py
```diff
@@ -458,11 +463,13 @@ def get_types(residue, atomname):
...
         # Iterate over all residues
-        # TODO: Do we have to iterate over all patched residues too?
-        for name, residue in iteritems(self.residues):
+        for name, residue in chain(iteritems(self.residues), iteritems(self.patches)):
             for impr in residue._impr:
                 # Get the list of types involved in this improper
-                types = [ get_types(residue, atomname) for atomname in impr ]
+                try:
+                    types = [ get_types(residue, atomname) for atomname in impr ]
+                except:
```
swails, line 471: We should catch a narrower exception here.  What do we expect can be thrown?

parmed/openmm/parameters.py
```diff
@@ -159,9 +165,8 @@ def _remediate_residue_template(cls, params, residue):
                     # Remove nonphysical H-H bonds
                     LOGGER.debug('Deleting H-H bond from water residue {}'.format(residue.name))
                     residue.delete_bond(bond)
-                elif (bond.atom1.atomic_number == 0) or (bond.atom2.atomic_number == 0):
-                    LOGGER.debug('Deleting bonds to virtual sites in residue {}'.format(residue.name))
-                    residue.delete_bond(bond)
+                else:
+                    LOGGER.debug('keeping %s to %s %s' %(str(bond.atom1), str(bond.atom2), bond.atom2.element_name))
```
swails, line 169: Minor nit, but you don't need the `str()` around the atoms, and you should replace the `%` operator with a comma:

```python
LOGGER.debug('keeping %s to %s %s', bond.atom1, bond.atom2, bond.atom2.element_name)
```

parmed/openmm/parameters.py
```diff
@@ -149,7 +149,13 @@ def _remediate_residue_template(cls, params, residue):
         for lonepair in residue.lonepairs:
             lp_atom = lonepair[1]
             params.atom_types[types[lp_atom]].atomic_number = 0
-            atom.atomic_number = 0
+            residue[lp_atom].atomic_number = 0
+
+        # CHARMM Drude force field includes bonds to lone pairs.  Delete them.
+        for bond in list(residue.bonds):
```
swails, line 155: You shouldn't need to coerce to a list here -- just iterate over the bonds directly.

parmed/openmm/parameters.py
```diff
@@ -149,7 +149,13 @@ def _remediate_residue_template(cls, params, residue):
...
+            if (bond.atom1.atomic_number == 0) or (bond.atom2.atomic_number == 0):
```
swails, line 156: I think `isinstance(bond.atom1, ExtraPoint)` (imported from `topologyobjects` is a better way of determining whether something is a virtual site or lone pair or something than checking if the atomic number is 0.

parmed/openmm/parameters.py
peastman: Fixed!

parmed/openmm/parameters.py
peastman: Fixed!

parmed/openmm/parameters.py
peastman: This is needed for a different reason.  If you modify the list of bonds at the same time you're iterating over it, things get confused and some bonds get missed.  The call to `list()` means we're iterating over an unchanging copy.

parmed/openmm/parameters.py
peastman: I tried but it doesn't work.  The atoms are all instances of `Atom` or `DrudeAtom`, not `ExtraPoint`.

parmed/openmm/parameters.py
swails: Oh of course. I’ve done this before. Can you add a comment to this effect?

peastman: Sure.  (This isn't my code though.  It was already there.  I just moved it to a different point in the method.)

peastman:
> Do you have a test case or two you can add here?

Not anything simple.  My "test" is converting the latest version of the CHARMM Drude force field.  OpenMM will have a test case for the converted force field to check that a particular system matches the energy computed by CHARMM.

jchodera:
@swails : Once this is merged, is there any possibility it could get pulled into an AmberTools patch update sometime soon?

PR #1092, opened 2020-06-04
Allow 'strbnd' to be optional for AMOEBA Parameters

mquevill:
For some very small molecules, `'strbnd'` may not exist in parameter files, so allow fall-through if this is the case.


swails:
Thanks!

PR #1097, opened 2020-06-28
Fix nonbond_params parsing and test

### LINKED ISSUE #1094: Wrong output for gromacs nonbond_params
Hello ParmEd Team,

in "parmed/gromacs/gromacstop.py" there seems to be a mistake while writing gromacs topologies in "nonbond_params" sections. See lines 1484-1485. 

`sig = val[0];
eps = val[1]`

shoulde be

`sig = val[1];
eps = val[0]`

when compared to how they are assigned in lines 387-388. With this change I get similar values when reading a topology and then writing it.

Cheers,

Daniel

mattwthompson:
Fixes #1094 

Pushing as separate commits to ensure the patch actually fixes the bug as it does locally (test-driven development and all)


swails:
Thanks!

PR #1102, opened 2020-08-13
Convert AMBER input files to DLPOLY input files

vwcruzeiro:
This PR adds to ParmEd the capability to convert AMBER input files to DLPOLY input files. The opposite process is impossible because the topology file in DLPOLY does not contain information about residues, only about "molecules" (in DLPOLY, one "molecule" cannot have bonds with any other "molecule").

This PR also adds tests for the new functionality.


parmed/amber/netcdffiles.py
```diff
@@ -264,7 +264,7 @@ def coordinates(self, stuff):
     def velocities(self):
         if 'velocities' in self._ncfile.variables:
             vels = self._ncfile.variables['velocities'][:]
-            return (vels.reshape((-1, self.atom, 3)) * self.velocity_scale)
+            return (vels.reshape((1, self.atom, 3)) * self.velocity_scale)
```
swails, line 267: This should be `-1` because there may be more than 1 frame.  The `-1` indicates that the dimension should take whatever value necessary to give a shape that accounts for the full length of the array.  For NetCDF trajectories that have more than 1 frame, this will fail.

vwcruzeiro: @swails , makes sense. Thank you for carefully checking on this.

`pyflakes` is detecting failures and for this reason it is killing the Travis CI build. What should I do to fix this? Just remove the "imported but unused" stuff?

Thanks!

swails: Yes, resolve the pyflakes issues.

parmed/dlpoly/__init__.py
```diff
@@ -0,0 +1,46 @@
+"""
+Contains classes for parsing DLPOLY topology and parameter files
+"""
+import os as _os
+from parmed.utils import which as _which
```
hainm, line 5: hi, what's the aim of this?

parmed/amber/netcdffiles.py
vwcruzeiro: @swails : continuing on the first comment you made, I noticed that a few lines above ( parmed/amber/netcdffiles.py#255 ) there is an equivalent function for the coordinates, and there is no `-1` inside the `reshape` function:
```
            return coords.reshape((1, self.atom, 3))
```
This is why I made the change you commented in the `velocities` function. Shall I place a `-1` in this line as well?

parmed/dlpoly/__init__.py
vwcruzeiro: I implemented the functionalities on this PR starting from what was already in ParmEd for Gromacs. I am not sure why these `as` choices were employed.

hainm: I see.

swails: It prevents them from being imported with `from parmed.dlpoly import *`

parmed/amber/netcdffiles.py
swails: Yes, I think that's a good idea.

vwcruzeiro:
@swails : apparently the tests that failed were complaining about the following: _Could not find netCDF4 module_. How can we fix that?

vwcruzeiro:
Hello @swails , just making sure my last comment did not get lost. I am unsure how to proceed regarding the errors in Travis CI. These appear to be unrelated to my changes in this PR. Thanks.

swails:
> Hello @swails , just making sure my last comment did not get lost. I am unsure how to proceed regarding the errors in Travis CI. These appear to be unrelated to my changes in this PR. Thanks.

Yea, I saw it.  The problem is that the netcdf bundled with ParmEd (taken from scipy) uses deprecated features.  I updated to the latest version of that library in #1104.  As long as that passes the tests (i.e., it doesn't break Python 2.7 or something else), then we can merge that and update your branch with the fix.

swails:
Finally fixed the NetCDF/Travis issues.

swails:
> The opposite process is impossible because the topology file in DLPOLY does not contain information about residues, only about "molecules" (in DLPOLY, one "molecule" cannot have bonds with any other "molecule").

There's no reason we can't map a molecule from DLPOLY to a residue in Amber.  The main target functionality here is to get the same energies and forces in both programs.  If the informatics portion doesn't map between the two programs I don't really care.  For actual informatics applications, neither DLPOLY or AMBER input files will be used.

parmed/dlpoly/__init__.py
```diff
@@ -0,0 +1,46 @@
...
+__all__ = ['DLPOLY_TOPDIR', 'DlpolyFieldFile', 'DlpolyConfigFile']
+
+DLPOLY_TOPDIR = None
+
+if _os.getenv('GMXDATA') is not None and _os.path.isdir(
+        _os.path.join(_os.getenv('GMXDATA'), 'top')):
+    DLPOLY_TOPDIR = _os.path.join(_os.getenv('GMXDATA'), 'top')
+elif _os.getenv('GMXBIN') is not None and _os.path.isdir(
+        _os.path.join(_os.getenv('GMXBIN'), '..', 'share', 'dlpoly', 'top')):
+    DLPOLY_TOPDIR = _os.path.join(_os.getenv('GMXBIN'), '..', 'share',
```
swails, line 16: Does this really use the GROMACS environment variables?

parmed/dlpoly/__init__.py
swails, line 5: Make this a relative import:

```python
from ..utils import which as _which
```

parmed/dlpoly/dlpolyconfig.py
```diff
@@ -0,0 +1,145 @@
...
+(coordinate) files and building a stripped-down Structure from it
+"""
+from __future__ import print_function, division, absolute_import
+
+from parmed.constants import TINY
+from parmed.formats.registry import FileFormatType
+from parmed.geometry import (box_lengths_and_angles_to_vectors,
+                             reduce_box_vectors)
+from parmed.utils.io import genopen
+from parmed.utils.six import add_metaclass, string_types
```
swails, line 12: Relative imports

parmed/dlpoly/dlpolyfield.py
```diff
@@ -0,0 +1,588 @@
...
+
+from parmed.exceptions import DlpolyError
+from parmed.formats.registry import FileFormatType
+from parmed.parameters import ParameterSet
+from parmed.structure import Structure
+from parmed.topologyobjects import (ExtraPoint, DihedralType, DihedralTypeList,
+                                    UnassignedAtomType)
+from parmed.utils.io import genopen
+from parmed.utils.six import add_metaclass, string_types, iteritems
+from parmed.utils.six.moves import range
```
swails, line 18: Relative imports.

swails (changes requested):
Relative imports are preferable because they permit easier relocation/subpackaging of ParmEd.

parmed/dlpoly/__init__.py
vwcruzeiro: Ops... I have removed these unnecessary calls. Good catch!

vwcruzeiro:
Thank you very much for your comments @swails . I have made the changes you suggested, and I also pushed your recent changes. Let's hope all tests will pass now.

vwcruzeiro:
All test are passing now 👍

vwcruzeiro:
Is this ready to go? :D

vwcruzeiro:
Hello @swails . Just a friendly reminder about this pull request. Happy Thanksgiving!

swails:
I'm closing this in favor of #1146.  I created a branch from this PR, merged master, and refactored to get rid of the code duplication from GROMACS.

@vwcruzeiro, this refactor continues to pass the tests you added.  Can you give it a quick look and I'll merge when you give the go-ahead?

vwcruzeiro:
@swails : I confirm that it is working. Thank you!

PR #1115, opened 2020-10-10
Add cphmd/cemd limits section to cpin/cein

swails:
- [x] Tests need to be updated for this (they will fail)
- [x] We need to make sure a cpin file generated from here runs in "old" versions of Amber
- [x] We need to make sure this functionality "works"


swails:
@vwcruzeiro - proposed changes here

vwcruzeiro:
> We need to make sure a cpin file generated from here runs in "old" versions of Amber

I have tested this. The new namelist will be ignored in older versions of Amber, exactly what we want.

vwcruzeiro:
@swails I have tested your changes, and all is good. Please change line 381 in titratable_residues to:
`if (typ != "redox"): limit_buf.add_word(' maxh=%d,' % max(len(r.states) for r in self))`
We do not need maxh for CEIN files.

Thanks!

swails:
If it doesn't hurt as-is, let's just keep it. The fewer branches we can get away with the better.

vwcruzeiro:
> If it doesn't hurt as-is, let's just keep it. The fewer branches we can get away with the better.

Ok. If we keep it, the code will still run. maxh will be read by pmemd/sander, but will be ignored at constante.F90

swails:
It needs to be in the name list in for conste. Otherwise reading it will fail.

vwcruzeiro:
maxh is in the namelist at constante.F90, so we are good.

swails:
Finally :/

PR #1122, opened 2020-11-09
Fix the order of assignment of the NBFIX terms.

HanatoK:
I used the topology converted from PSF to the gromacs format but the
simulation crashed, and after checking the topology file I found the
epsilon and sigma were wrong. It seems PR #1097 only partially changed
the order of assignment of NBFIX. This commit should fix the issue.


swails:
Thanks!

PR #1138, opened 2021-02-04
Major cleanup and remove Python 2 support

swails:
Python 2 support has been dropped by most major projects, and I am dropping support here.  ParmEd will require Python 3.6+ after this PR is merged.

It also removes functionality that has been deprecated for years.  The next release will be ParmEd 4.0.


swails:
image

Cut down 4000 lines of code!

hainm:
> remove Python 2 support

I will do for pytraj too.

PR #1139, opened 2021-02-04
Check to make sure CHARMM patch doesn't create empty residue

jchodera:
This PR makes a minor modification to how CHARMM patch compatibility is checked, and raises an exception if a patch results in all atoms and bonds being removed from a residue template.


swails (approved):
LGTM!

PR #1141, opened 2021-02-10
Bug in Atom.umass

Andrew-AbiMansour:
**Bug**: 
   `self._mass` seems to be undefined in the [Atom](https://github.com/ParmEd/ParmEd/blob/master/parmed/topologyobjects.py#L232) class. Therefore, calling [Atom.umass](https://github.com/ParmEd/ParmEd/blob/master/parmed/topologyobjects.py#L727) raises an AttributeError.

**Fix**: 
   Replace `self._mass` with `self.mass`

### DIFF
--- a/parmed/topologyobjects.py
+++ b/parmed/topologyobjects.py
@@ -725,7 +725,7 @@ class Atom(_ListItem):
 
     @property
     def umass(self):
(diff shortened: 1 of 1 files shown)


swails:
Thanks!!

Issue #1136, opened 2021-01-26
Support modified scee/scnb in OpenMM export


jchodera:
It would be useful if we could add the ability to export OpenMM systems with modified `scee` and `scnb` nonbonded scaling parameters. Attempting to convert the AMBER GLYCAM06j-1 glycan force field yields the following error:
```
Traceback (most recent call last):
  File "convert_amber.py", line 1277, in <module>
    main()
  File "convert_amber.py", line 107, in main
    convert_yaml(args.input, ffxml_dir='ffxml/')
  File "convert_amber.py", line 420, in convert_yaml
    ffxml_name = convert_leaprc(files, ffxml_dir=ffxml_dir, ignore=ignore,
  File "convert_amber.py", line 212, in convert_leaprc
    params.write(ffxml_name, provenance=provenance, write_unused=write_unused, improper_dihedrals_ordering='amber')
  File "/Users/choderaj/miniconda/envs/openmmforcefields/lib/python3.8/site-packages/parmed/openmm/parameters.py", line 65, in wrapper
    return func(*args, **kwargs)
  File "/Users/choderaj/miniconda/envs/openmmforcefields/lib/python3.8/site-packages/parmed/openmm/parameters.py", line 368, in write
    self._write_omm_nonbonded(root, skip_types, separate_ljforce)
  File "/Users/choderaj/miniconda/envs/openmmforcefields/lib/python3.8/site-packages/parmed/openmm/parameters.py", line 65, in wrapper
    return func(*args, **kwargs)
  File "/Users/choderaj/miniconda/envs/openmmforcefields/lib/python3.8/site-packages/parmed/openmm/parameters.py", line 886, in _write_omm_nonbonded
    raise NotImplementedError('Cannot currently handle mixed 1-4 '
NotImplementedError: Cannot currently handle mixed 1-4 scaling: Elec. Scaling factors 1.0, 1.2 detected
```
cc @peastman @ivy-zhang @glass-w

swails:
My understanding is that this needs support from OpenMM.  If that support is already there, then the OpenMMParameterSet writer needs to be tweaked to write it.

peastman:
When a ForceField creates a NonbondedForce, it calls `createExceptionsFromBonds()` on it, passing in the specified scale factors.  There's no way to use different scale factors for different molecules in the system.  To do that, we would need to build the exceptions directly.

jchodera:
[...]

PR #1149, opened 2021-02-17
Allow some 1-4 interactions to be unscaled

peastman:
This partially address #1136.  This code has only been minimally tested!  To fully convert a force field and do proper tests, we need to also update the openforcefields conversion script, and possibly add a feature to ParmEd so you can specify a prefix for atom types.


swails:
Looks like master needs to be merged first?

peastman:
What do you mean?

swails:
image

peastman:
I don't get that message!  It just tells me merging is restricted to authorized users.  I've hopefully done the merge it wants.

swails:
@jchodera - do you have a test for this?

jchodera:
@peastman: I'm a bit confused by the implementation. It looks like you go through the effort of tracking which atom types have nonstandard scee/scnb and modifying the 1-4 interactions for those atoms, but _only_ for those that set scee/scnb to 1 exactly. Anything else still throws an exception. Am I missing something, or wouldn't it be straightforward to extend this to track any nonstandard scee/scnb and apply those nonstandard values in the script?

> @jchodera - do you have a test for this?

@swails: I'm guessing that the easiest way to properly test this is to convert `leaprc.` and `leaprc.GLYCAM_06j-1`, export these to OpenMM ffxml file, and then try to parameterize a short glycosylated peptide such as [1KYJ](https://www.rcsb.org/structure/1KYJ). If we prepare Amber prmtop/inpcrd files with LEaP and compare the potential energy for OpenMM-parameterized peptide and the Amber prmtop/inpcrd system, we should be able to detect errors.

The difficulty here is that preparing even simple glycosylated peptide systems in Amber is an incredible pain. It would be much easier if there was a pre-prepped system ready to go. 

**Update:** I _think_ I may have found a pre-prepared system in `amber20_src/AmberTools/test/leap/glycam/06j/Chains.{parm7,rst7}.save`, which I can convert to a (someone noncompliant) PDB with `CONECT` records with
```bash
ambpdb -p Chains.parm7.save -c Chains.rst7.save -conect > Chains.pdb
```

I can work on this test, but I'll have to open a separate PR since this is contributed from @peastman's branch. 

Note that the GLYCAM force field conversion may be a problem in general since they inexplicably seem to have made the choice to use a different definition of what constitutes a "residue" than the PDB, so while we can go from AMBER PDB -> OpenMM System, we won't be able to go from RCSB PDB -> OpenMM System with these definitions.

Really looking forward to OpenFF adding support for glycans so all of this will "just work"...

jchodera:
@peastman : I realize I'm running really short on time, and don't want to hold this PR up further, so here is a ZIP archive of a directory you can use in this test:
[glycan.zip](https://github.com/ParmEd/ParmEd/files/6014005/glycan.zip)

* This `glycan` test directory should be added to `ParmEd/test/files/`.
* It looks like `ambertools` is included in the test environment, so you shouldn't have to include any force field files---just load `oldff/leaprc.ff10` and `oldff/leaprc.GLYCAM_06j_10` as parameter sets, write them to OpenMM ffxml files, and then use them to parameterize `Glycoprotein.pdb` and create a `System` object.
* This can be compared with a `System` object created with the OpenMM app `AmberPrmtop/Inpcrd` loaders from `Glycoprotein.rst7.save` (inpcrd) and `Glycoprotein.parm7.save` (prmtop)

peastman:
> Am I missing something, or wouldn't it be straightforward to extend this to track any nonstandard scee/scnb and apply those nonstandard values in the script?

See the discussion at https://github.com/ParmEd/ParmEd/issues/1136#issuecomment-780237656.

jchodera:
@peastman: I'm not sure I quite follow. Couldn't we throw an exception _only_ if we run into an ambiguous case, such as those cases you enumerate? This PR seems to do the opposite, and _only_ handles the special case where the scee/scnb are exactly unity.

swails:
> This PR seems to do the opposite, and _only_ handles the special case where the scee/scnb are exactly unity.

This PR is quite narrow in that it supports any combination of force fields where there is only one non-unity set of scaling factors supported.  While narrow, it does provide support for all Amber force fields and combinations out there: GLYCAM uses unity scaling factors and the protein, RNA, and Lipid force fields all have the same scaling factors (1/1.2 electrostatic and 1/2 Lennard-Jones).

So this PR should support GLYCAM + any/all of the other Amber force fields, since it will support the single scaling factor from the protein, RNA, and Lipid force fields along with the unity (unscaled) GLYCAM force field.

jchodera:
OK, let's run with it, then.

@peastman: Here are the files you can use in a test!
https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-782546139

jchodera:
@peastman: Could you add the test that @swails requested using the test input files I provided in #1149? 

I'm unfortunately swamped by grant proposals right now, so wouldn't be able to get to this for a couple of weeks.

peastman:
I'll try to get to it within the next few days.  I've been busy with other things.

jchodera:
Thanks! Once this is done, I can quickly convert the GLYCAM force field, which will really help our group prepare a bunch of the new SARS-CoV-2 mutants and get them on Folding@home ASAP so we can assess their impact on ACE2 binding! The lack of GLYCAM is blocking them.

peastman:
I'm not sure what I'm supposed to do with those files?  There's a PDB file, and a bash script that just sets a single variable, and then some other files of types I'm not familiar with.

I'm attaching the converted force field.  I edited the `unscaled_types` array by hand to add the `glycan-` prefix to all the types, but otherwise it's exactly what you get from this code and your convert_amber.py script.

[GLYCAM_06j-1.xml.txt](https://github.com/ParmEd/ParmEd/files/6053395/GLYCAM_06j-1.xml.txt)

jchodera:
> I'm not sure what I'm supposed to do with those files? There's a PDB file, and a bash script that just sets a single variable, and then some other files of types I'm not familiar with.

@peastman: Apologies for the miscommunication. Here's a summary:
* @swails asked that a test be included in your PR
* I suggested a simple test could be to convert the GLYCAM force field and then have OpenMM `ForceField` parameterize a PDB file that contains a few amino acids and a few glycans, computing the energy both with the `ForceField` parameterized `System` and a `System` created from the `prmtop/inpcrd` provided by AMBER.
* I found a GLYCAM test  included with the AmberTools20 distribution that includes a short peptide with a few glycans, and ZIPped up a directory containing the `prmtop/inpcrd` provided by AMBER and a PDB file with `CONECT` records.
* I posted this here with a description of the test suggestion since I'm swamped with grant proposals to fund COVID-19 drug discovery work, and asked if you could add the test to your PR in the meantime.

@peastman: Can you also review the important details in https://github.com/openmm/openmmforcefields/pull/156 where my group attempted to use this PR? There are two main issues:
* The GLYCAM force field uses some Amber protein atom types, so the prefix `glycam-` will not work. We need to unify the namespace of types or classes with the protein force field. This requires we make some decisions regarding the still-open issue https://github.com/openmm/openmmforcefields/issues/99
* Given that you had to add the prefix by hand, it seems we will also want to extend the ParmEd ffxml parameter writing method to allow us to _specify_ the prefix so that we don't have to reprocess the XML afterwards. This will significantly simplify things.

Thanks so much for your help in wrapping this up, @peastman---the group is trying to get these simulations of SARS-CoV-2 mutants set up ASAP so we can help our experimental collaborators determine whether we need to worry about emerging variants in developing therapies.

peastman:
> and ZIPped up a directory containing the prmtop/inpcrd provided by AMBER and a PDB file with CONECT records.

There are no prmtop or inpcrd files in your zip.  Can you check that you included the right set of files?

> Can you also review the important details in openmm/openmmforcefields#156 where my group attempted to use this PR? There are two main issues:

You left out the biggest problem.  Quoting from that issue,

> Currently, the primary showstopper for the test is that ffxml/GLYCAM_06j-1.xml does not contain residues that are actually part of glycans. The only residues that are present are the modified protein residues to which glycans can bond (e.g. NLN, which is the modified version of ASN).

I can confirm that only those residues are present in GLYCAM_amino_06j_12SB.lib.  I don't know enough about the force field to comment on whether others should be present as well, or where it should get them from.

> Given that you had to add the prefix by hand, it seems we will also want to extend the ParmEd ffxml parameter writing method to allow us to specify the prefix so that we don't have to reprocess the XML afterwards.

Possibly so.  Alternatively it could be done with a modification to convert_amber.py.  See my comment at the top of this thread.  That's a separate feature though.

jchodera:
> There are no prmtop or inpcrd files in your zip. Can you check that you included the right set of files?

Sorry about the information being spread out in multiple comments---I mentioned this [here](https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-782546139). Older versions of amber sometimes use different file naming conventions, so `Glycoprotein.parm7.save` is an Amber 7 format `prmtop` file (the current generation of `prmtop` files we read, I believe), and `Glycoprotein.rst7.save` is an `inpcrd` file.

> You left out the biggest problem. Quoting from that issue,
> I can confirm that only those residues are present in GLYCAM_amino_06j_12SB.lib. I don't know enough about the force field to comment on whether others should be present as well, or where it should get them from.

I think I can fix that problem!

The residue definitions are loaded by `leaprc.GLYCAM_06j-1` from a PREP file:
```
##
#
#       load all prep files for polysaccharides
#
loadamberprep GLYCAM_06j-1.prep
```
The issue appears to be that ParmEd does not implement the `loadamberprep` command for LEaP---something I hadn't realized before. It must assume that all old-style `.prep` files have been converted to `lib` files, but that has only happened with the modified amino acids for GLYCAM.

The GLYCAM leaprc files seem to be the only files in the current AmberTools distribution that use `loadamberprep`.

It appears I can manually create a `.lib` file from this `.prep` file using a leap script such as this:
```
loadamberprep GLYCAM_06j-1.prep
saveoff { UYB 4YB VMB 2MA 0YB 0FA ... } GLYCAM_06j-1.lib
```
where I have to list the whole set of 100+ residue names in `{...}`. 

It looks like I can do this and include the modified `lib` file and parameters in `openmmforcefields`, so I should be able to solve that problem there.

@swails: How would we get such a change eventually made upstream in the AmberTools distribution?

> Possibly so. Alternatively it could be done with a modification to convert_amber.py. See my comment at the top of this thread. That's a separate feature though.

I think it would simplify things greatly to add this feature to ParmEd so we don't have to hack XML files.

swails:
I think we can simply do a regression test here.  Convert a GLYCAM force field into an OpenMM XML file and then just compare the generated XML with what this version of the code produces.

> @swails: How would we get such a change eventually made upstream in the AmberTools distribution?

It's not really fair to try and force the GLYCAM maintainers to abandon PREP files to make this particular task easier, although they may be amenable if it increases adoption of GLYCAM.

It may be time to simply bite the bullet and add a PREP file parser (but no writer, since I think Amber is actively trying to kill off that format).  I'll also need to break down and actually write code (i.e., copy code from somewhere) to interconvert between internal and cartesian coordinates.

jchodera:
> It's not really fair to try and force the GLYCAM maintainers to abandon PREP files to make this particular task easier, although they may be amenable if it increases adoption of GLYCAM.

The `.prep` format is 27 years old, isn't it? A pre-LEaP remnant.

drroe:
> The .prep format is 27 years old, isn't it? A pre-LEaP remnant.

The PDB format is over 40 years old... 🤷‍♂️

jchodera:
> The PDB format is over 40 years old... 🤷‍♂️

That's why the RCSB was sensible enough to [formally deprecate it](https://journals.iucr.org/d/issues/2019/04/00/rr5177/index.html) and force everyone to use a more modern, flexible format. :)

Supporting old formats should be "best effort", but can't drive development into a corner.

drroe:
> Supporting old formats should be "best effort", but can't drive development into a corner.

Absolutely. If everyone stopped using the PDB format tomorrow I would clap my hands and jump for joy. My point however is that as of right now it is still one of the most widely-used formats, so supporting it remains a necessary evil. 

If this issue can be solved without adding support for the PREP format, that would be great. However, if supporting PREP is the only (or most straightforward) way to address this issue, then as Jason said it might be time to bite the bullet. I think the next step is to reach out to the Glycam devs directly and get their take. If they are willing to abandon the PREP format, then this becomes a non-issue and that old format can finally be put out to pasture...

jchodera:
> If this issue can be solved without adding support for the PREP format, that would be great. However, if supporting PREP is the only (or most straightforward) way to address this issue, then as Jason said it might be time to bite the bullet. I think the next step is to reach out to the Glycam devs directly and get their take. If they are willing to abandon the PREP format, then this becomes a non-issue and that old format can finally be put out to pasture...

My understanding is that a two-line LEaP script is all that is needed to generate the requisite `lib` file for inclusion in AmberTools, but that this step hasn't been undertaken because it was not recognized as being useful. Now, we have a clear use case, and all we need to do is persuade the AmberTools/GLYCAM devs to add this two-line LEaP script and run tleap with each new release. Even better would be to also make a one-line change in the GLYCAM `leaprc` files to use the lib file instead.

In the meantime, I can run the two-line LEaP script manually and bundle the files with `openmmforcefields`, and eliminate those files once AmberTools is updated. I already do this with all `gaff*.dat` versions,  since AmberTools does not bundle them all (but should).

swails:
I'm just not 100% confident that OFF and PREP files have feature parity.  GLYCAM pushes tleap to its absolute limits in terms of modeling, so it's entirely possible that converting to OFF files will simply break GLYCAM leaprc files (especially if they're internal-coordinate prep files instead of cartesian).

jchodera:
The internal coordinate representations are presumably useful for building glycans onto amino acids, but (1) none of the approaches mentioned in the Amber20 manual uses this feature, and (2) it's irrelevant to the application of parameters described in LEaP in the GLYCAM instructions and examples.

If there is concern that some part of the potential energy specification is missing, an energy test using the existing examples should catch that.

swails:
I was merely pointing that out as a reason that the GLYCAM developers may resist switching from prep to OFF.  Certainly for Amber -> OpenMM XML there's no harm in converting to OFF.

jchodera:
@swails @peastman how do you want to handle finishing this PR?

jchodera:
I've just had a chat with @zhang-ivy, who points out four different issues we need to solve before this can be construed as working as intended. See her attempt to use this PR to convert the glycam force field at https://github.com/openmm/openmmforcefields/pull/156 :

1. Converting all of the residues found in the glycam `.prep` file (after conversion to `.lib`) ends up generating multiple templates for identical bond-element graphs. Glycam apparently considers different stereoisomers as different residues---presumably only for the purpose of building on glycans using the internal geometries in the residue definitions, since it does not appear to be the case that the parameters change. So we need a way of filtering the residue templates upon writing to only write unique residue definitions.
2. When @zhang-ivy converted a small subset of unique residues, she ran into the issue where the a glycosylated protein test system did not match energies between ParmEd (via LEaP) and OpenMM ForceField. There are potentially multiple causes, but the first readily observable issue is that [the post processing script generated on ffxml export](https://github.com/openmm/openmmforcefields/blob/4b0a7b3710da1ace6ec3b5c088ba2ce0c8defacf/amber/ffxml/GLYCAM_06j-1_trunc.xml#L1470-L1488) contains protein atom types as well as glycan atom types, meaning 1-4 interactions within the protein will be incorrectly altered as well:
```python
import simtk.openmm as mm
unscaled_types = ["Cp", "H2", "H", "CX", "NT", "O2", "Ck", "Hc", "Sm", "2C", "Cy", "Ng",
    "Oy", "C", "Ha", "S", "Oh", "Cg", "Ho", "CT", "Hp", "NH", "H1", "P",
    "O", "3C", "N", "Os", "N3", "Cj"]
```
Note the `C`, `CT`, `N`, `O` types, etc. I think this can be fixed by using the atom types defined in the parameter set, rather than the atom types used in the templates.
3. The [conversion script](https://github.com/openmm/openmmforcefields/blob/4b0a7b3710da1ace6ec3b5c088ba2ce0c8defacf/amber/ffxml/GLYCAM_06j-1_trunc.xml#L1470-L1488) also only unscales the charges, but not the Lennard-Jones `epsilon`:
```pyton
      if p1 in unscaled_atoms and p2 in unscaled_atoms and chargeProd._value != 0:
        force.setExceptionParameters(i, p1, p2, atom_charges[p1]*atom_charges[p2], sigma, epsilon)
```
If I'm reading the prmtop sections right, both scee and scnb are reset to 1 for glycam:
```
%FLAG SCEE_SCALE_FACTOR                                                         
%FORMAT(5E16.8)                                                                 
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.20000000E+00  1.20000000E+00
  1.20000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.20000000E+00
  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00
  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00
  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00
  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00  1.20000000E+00
  1.20000000E+00  1.20000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  0.00000000E+00  0.00000000E+00  0.00000000E+00
%FLAG SCNB_SCALE_FACTOR                                                         
%FORMAT(5E16.8)                                                                 
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  2.00000000E+00  2.00000000E+00
  2.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  2.00000000E+00
  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00
  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00
  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00
  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00  2.00000000E+00
  2.00000000E+00  2.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00  1.00000000E+00
  1.00000000E+00  0.00000000E+00  0.00000000E+00  0.00000000E+00
```
4. Exporting to OpenMM ffxml omits external bonds records for the glycosidic linkages. This appears to be due to the fact that ParmEd only has a concept of "head" and "tail" for external bonds. @swails : can you think of a straightforward way to enable additional atoms to be perceived as appropriate linkages? @zhang-ivy wrote a quick hack in https://github.com/openmm/openmmforcefields/pull/156#issuecomment-791754672, using some atom naming and valence matches, but I'm uncertain of the most robust way to do this.

jchodera:
Also, my grant proposal was submitted today, so I'm happy to finish this up if @peastman @swails can offer pointers about the best way to proceed!

peastman:
This PR isn't intended to be a complete fix for all problems in converting glycam.  It addresses one specific problem, recognizing that other changes will be needed both to ParmEd and to the conversion script.  So to focus on just this PR for the moment, what changes are required to make unscaled 1-4 interactions work?

zhang-ivy:
@peastman In John's comment, items 2 and 3 are the ones most relevant to this PR. 
Additionally, @jchodera  Were you able to figure out (from looking at the one of the prmtops) whether protein-glycan 1,4s should be left unscaled?

peastman:
> @peastman In John's comment, items 2 and 3 are the ones most relevant to this PR.

Those are the problems that need to be solved, but how do you want them addressed?  What are the specific changes you need to this PR?

zhang-ivy:
I've opened a PR in @peastman's fork with my attempt to address the problems outlined by item 3. I'm not sure how to fix item 2, as some of the protein atoms in `unscaled_types` are in fact involved in glycan ff specific residues, e.g. NLN, which is an ASN that is modified such that it can bond to a glycan residue.

peastman:
I had the script decide whether to apply scaling with the line

```python
if p1 in unscaled_atoms and p2 in unscaled_atoms and chargeProd._value != 0:
```

It sounds like it should actually be `or` rather than `and`?  When a glycam atom interacts with a standard atom, the scaling is removed?

How then do I determine which are the glycam atom types?  ParmEd reports scaling factors for individual dihedrals, not by atom type.  One of the [assumptions I made](https://github.com/ParmEd/ParmEd/issues/1136#issuecomment-780189609) is, "If an atom type uses particular scaling factors for one 1-4 interaction, it will use those same factors for every 1-4 interaction."  It sounds like that isn't actually correct.  How can we determine whether an atom type should be unscaled?

peastman:
Does anyone have insight on the questions above?

zhang-ivy:
>Does anyone have insight on the questions above?

I think we need @jchodera's insight here

jchodera:
@peastman @zhang-ivy @swails : Apologies for the delay in investigating this further---I hadn't had time to deeply dive into the details until now.

**SCEE/SCNB:**
I've carefully inspected the 1-4 exceptions in the AMBER-prepared peptide:glycan test systems, and found---to-my surprise---that the 1-4 exceptions that bridge peptide and glycan have SCEE=1, but the intra-peptide and intra-glycan exceptions were the default SCEE=1.2. On re-reading the Amber20 manual on why this might be the case, I found this excerpt, which notes that it's more complicated than this: Amber11 and later added support to `frcmod` files for specifying non-default SCEE/SCNB for individual torsion types:

image

The `GLYCAM_06j.frcmod` file is filled with a variety of SCEE/SCNBs:
```
Cy-Cg-Cg-Cg   1    0.45          0.0             1.         SCEE=1.0 SCNB=1.0 Butane - for sialic acid only!
Cy-Cg-Cg-Hc   1    0.10          0.0             3.         SCEE=1.0 SCNB=1.0 Propane - for sialic acid only!
Cy-Cg-Cg-H1   1    0.15          0.0             3.         SCEE=1.0 SCNB=1.0 2-Dimethylbutanol - for sialic acid only!
Cy-Cg-Cg-Oh   1    0.10          0.0             3.         SCEE=1.0 SCNB=1.0 2,3-Dimethylbutanol - for sialic acid only!
H -Ng-C -O    1    2.50        180.0            -2.         SCEE=1.2 SCNB=2.0 Copy of Parm99 JCC,7,(1986),230 J.C.cistrans-NMA DE
              1    2.00          0.0             1.         SCEE=1.2 SCNB=2.0
HC-CT-3C-Os   1    0.000         0.0            -3.         Copied from ff12SB/parm10 HC-CT-3C-OS/OH, leaving scaling factors as-is
              1    0.250         0.0             1.         Copied from ff12SB/parm10 HC-CT-3C-OS/OH, leaving scaling factors as-is
```

I should have read the ParmEd code more carefully, since the SCEE/SCNB values are stored for each torsion.

The better solution here would be to consider extending the OpenMM ffxml format to enable torsions to specify non-default SCEE/SCNB scaling on a per-torsion basis. 

A quick fix for now that is clearly an improvement would be to modify this PR to store a list of quartets for the torsion atom types, and use that list to identify torsions that should have their 1-4 terms recomputed because they are non-default. From examining the files that use this, it looks like we can get away with just the (1,4) pairs of atom types because the central types don't even impact the choice.

**Removing duplicate templates:** I believe we still need to adapt the writing of OpenMM ffxml files to remove duplicate templates.

@peastman : Would you be able to adapt your PR for these changes?

**Atoms with external bonds:** Automatically identifying and then storing connection points is a much harder problem to solve, because (1) ParmEd does not have a place to store these, and (2) the GLYCAM prep files do not have any way of identifying these linkage atoms. Either we need to compile a (long) list of specific link atoms for each residue manually, or we adapt @zhang-ivy's clever strategy of identifying open-valence atoms on a restricted domain of atom types.

@swails: Any ideas here? Should we just use a variant of @zhang-ivy's [clever open valence identification scheme](https://github.com/peastman/ParmEd/pull/1/files#diff-94168a681b41dcbfd3a141aee5743e5784000e5f7e71f643a4acbf41edbdbd88R617-R648) for now?

peastman:
To be completely clear: I should change it so instead of storing a list of atom types that don't get scaled, it should store a list of atom type pairs.  A pair will not be scaled if and only if the two interacting atoms match one of the type pairs.  Correct?

jchodera:
> To be completely clear: I should change it so instead of storing a list of atom types that don't get scaled, it should store a list of atom type pairs. A pair will not be scaled if and only if the two interacting atoms match one of the type pairs. Correct?

Correct!

We also need to:
1. Add an option (perhaps enabled by default?) to remove duplicate templates during writing
2. Integrate a version of @zhang-ivy's [clever open valence identification scheme](https://github.com/peastman/ParmEd/pull/1/files#diff-94168a681b41dcbfd3a141aee5743e5784000e5f7e71f643a4acbf41edbdbd88R617-R648) to perceive external bonds, which should probably be disabled by default, so that we can perceive the connections between residues

swails:
> The `GLYCAM_06j.frcmod` file is filled with a variety of SCEE/SCNBs:

Not really -- it has two different ones.  1.0 for any torsion whose end-groups are pure GLYCAM types, and 1.2 for any torsion whose end-groups have at least 1 "canonical" Amber FF type (protein or NA atom type).  "Bridging" between GLYCAM and protein/NA groups were decided to adopt the Amber FF scaling factors.

Before mixed-1-4 scaling was added to Amber, the advice from the GLYCAM folks when doing a mixed system was to use the protein/NA scaling factors.  I think that because what the scaling factors "should" have been became undefined, the choice was made to keep them consistent with the previous incarnations of GLYCAM+AmberFF.  That also would've been how these parameters were fitted.

Encoding nonbonded exception parameters in the torsion list is a hack, but it's how it's always been done in Amber.

If you just want to get GLYCAM + AmberFF working here, it's enough to keep a set of unsorted atom type pairs parsed from the torsion list whose scaling factors are 1.0, and then simply apply the "canonical" scaling factors (1/1.2 EEL, 1/2.0 VDW) for all other pairs.  That scheme would only work for Amber + GLYCAM, but I don't know of any other force fields that are "mixable" with this issue.

peastman:
I pushed updates that should hopefully fix the problems.

jchodera:
@peastman: Thanks so much! We're working on testing this.

zhang-ivy:
I've tested @peastman 's fix using my openmmforcefields branch (`zhang-ivy:convert-glycans`, PR: https://github.com/openmm/openmmforcefields/pull/156) and I'm getting the following error upon creating an OpenMM system:
```python
  File "/lila/home/zhangi/choderalab/openmmforcefields/amber/convert_amber.py", line 632, in assert_energies2
    system_omm = ff.createSystem(parm_amber.topology)
  File "/home/zhangi/miniconda3/envs/glycan-ff/lib/python3.9/site-packages/simtk/openmm/app/forcefield.py", line 1341, in createSystem
    exec(script, locals())
  File "<string>", line 51, in <module>
TypeError: Quantity.__float__ returned non-float (type Quantity)
```

peastman:
> Here are the amber and omm system xmls before and after peter's fix:

Thanks!  I'll take a look and see what's causing the difference.

peastman:
Could you also provide a serialized State (or a PDB file, or anything else from which I can get coordinates)?  I need that to evaluate energies.

peastman:
It looks to me like the `omm_after.xml` version doesn't contain any bonded terms inside the NLN residue.  For example, the following triples of atoms have harmonic angle terms in `amber_after.xml`, but not in `omm_after.xml`.

> {(155, 157, 166), (163, 162, 164), (160, 159, 161), (157, 159, 160), (159, 162, 164), (158, 157, 166), (166, 168, 169), (155, 157, 159), (156, 155, 157), (161, 159, 162), (167, 166, 168), (157, 166, 168), (157, 159, 162), (158, 157, 159), (159, 162, 163), (153, 155, 157), (155, 157, 158), (162, 164, 165), (160, 159, 162), (157, 166, 167), (159, 157, 166), (157, 159, 161), (153, 155, 156), (166, 168, 170)}

jchodera:
That suggests @zhang-ivy's PDB file is missing `CONECT` records?

@zhang-ivy is this a file you prepared, or the test peptides I passed long here?
https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-782546139

peastman:
test.pdb does include CONECT records.  If it didn't, it wouldn't match the template.  More likely this indicates a mismatch in the atom type names.  When it looks for parameters for a bond/angle/torsion, it doesn't find any.

zhang-ivy:
>@zhang-ivy is this a file you prepared, or the test peptides I passed long here?

test.pdb is a file the glycosylated RBD that I prepared. The test system you sent will not work because:
1) There is a glycan residue, OLP, that is present in the system, but not present in the GLYCAM_06j-10.prep file
3) Additionally, if you visualize the PDB in pymol, it looks very broken:
Screen Shot 2021-03-30 at 7 48 14 PM

zhang-ivy:
>It looks to me like the omm_after.xml version doesn't contain any bonded terms inside the NLN residue. 
Interesting... were you able to check whether the same is true for the glycan residues (e.g. 2MA, 0fA, UYB, etc)?
The NLN residue is a modified ASN residue such that it can be glycosylated and is present in the GLYCAM forcefield.

jchodera:
Just a quick clarification on this!

>  The test system you sent will not work because:
> There is a glycan residue, OLP, that is present in the system, but not present in the GLYCAM_06j-10.prep file

From the Amber20 manual, page 246:
image

These are sourced in the `leaprc.GLYCAM_06j-1` file:
```
# for attaching glycans to proteins
loadOff GLYCAM_amino_06j_12SB.lib
loadOff GLYCAM_aminoct_06j_12SB.lib
loadOff GLYCAM_aminont_06j_12SB.lib
## add residue maps for linking glycans to proteins
## Note! use of these requires sourcing the ff12SB leaprc and loading the amino 
##      libraries commented out in the load libs section 
addPdbResMap { 
        { 0 "OLS" "NOLS" } { 1 "OLS" "COLS" }
        { 0 "OLT" "NOLT" } { 1 "OLT" "COLT" }
        { 0 "OLP" "NOLP" } { 1 "OLP" "COLP" }
        { 0 "HYP" "NHYP" } { 1 "HYP" "CHYP" }
        { 0 "NLN" "NNLN" } { 1 "NLN" "CNLN" }
} 
```
and appear in these files:
```
GLYCAM_amino_06j_12SB.lib
GLYCAM_aminoct_06j_12SB.lib
GLYCAM_aminont_06j_12SB.lib
```
I think these are the modifications of the amino acids required for using GLYCAM---from pages 44-45 of the Amber20 manual:
image

image


> Additionally, if you visualize the PDB in pymol, it looks very broken:

There are multiple peptides, causing pymol to get confused. If you just tell it `hide all; show lines`, you can see the system clearly:
glycan

zhang-ivy:
@jchodera  Thanks! It seems that there are some complexities I missed when trying to re-write the test to use your test system. Since the most immediate use case of the converted Glycam forcefield is the RBD:ACE2 system, I'm going to stick with my RBD test for now, but I can try to re-write the test with your system later on.

jchodera:
> @jchodera Thanks! It seems that there are some complexities I missed when trying to re-write the test to use your test system. Since the most immediate use case of the converted Glycam forcefield is the RBD:ACE2 system, I'm going to stick with my RBD test for now, but I can try to re-write the test with your system later on.

That's fine! I just wanted to clarify the failure, since it seems like the inclusion of the modified amino acids is a required part of using GLYCAM!

jchodera:
@peastman @zhang-ivy : I wanted to return to this to see if we could resolve it this PR and get the glycam force field converted.

@peastman: Any idea why OpenMM's `ForceField` isn't assigning angle terms?
https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-810582336

peastman:
Like I said above, it probably indicates a mismatch in atom type names.

jchodera:
Great! Can you help figure out why, @peastman? Given the importance of glycans to SARS-CoV-2 research, it would be great to get glycam support wrapped up and released in OpenMM. I'm running super low on time I can devote to this (just having spent several days writing for $600K more funding for our QML work).

peastman:
I'm trying to remember exactly where we were on this.  I believe @zhang-ivy created a force field XML file and used it to generate a System, which was missing bonded energy terms between atoms in the NLN residue?  Did you post your force field file?  I can't find it in the discussion above.  If you post it, I can see why the bonded terms are getting left out.

jchodera:
I think @zhang-ivy left instructions here: https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-810386829

jchodera:
I also provided a separate set of test files you could use: https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-782546139

peastman:
That's the conversion script.  It will be a lot easier for me if you can directly provide the converted force field.

zhang-ivy:
Sorry for the delay! Here is the converted ff I used to generate the system xmls: 
[GLYCAM_06j-1_trunc.xml.zip](https://github.com/ParmEd/ParmEd/files/6611235/GLYCAM_06j-1_trunc.xml.zip)
Note that this does not contain all glycan residues, only `NLN UYB 4YB VMB 2MA 0YB 0fA 0LB` because I ran into an template matching error when trying to convert all glycan residues to an openmm ffxml.

peastman:
When I try to load the force field from that file

```python
ff = ForceField('amber14-all.xml', 'amber14/tip3pfb.xml', 'GLYCAM_06j-1_trunc.xml')
```

I get an exception

```
  File "/Users/peastman/miniconda3/envs/openmm/lib/python3.9/site-packages/openmm/app/forcefield.py", line 310, in loadFile
    template.atoms.append(ForceField._TemplateAtomData(atomName, typeName, self._atomTypes[typeName].element, params))
KeyError: 'N'
```

Looking at the XML file, I see it using types that don't exist:

```xml
    <Residue name="NLN" override="1">
      <Atom name="N" type="N" charge="-0.4157"/>
      <Atom name="H" type="glycan-H" charge="0.2719"/>
      <Atom name="CA" type="glycan-CX" charge="0.0143"/>
      <Atom name="HA" type="glycan-H1" charge="0.1048"/>
      <Atom name="CB" type="glycan-2C" charge="-0.2041"/>
      <Atom name="HB2" type="HC" charge="0.0797"/>
      <Atom name="HB3" type="HC" charge="0.0797"/>
      <Atom name="CG" type="glycan-C" charge="0.713"/>
      <Atom name="OD1" type="glycan-O" charge="-0.5931"/>
      <Atom name="ND2" type="N" charge="-0.6935"/>
      <Atom name="HD21" type="glycan-H" charge="0.4196"/>
      <Atom name="C" type="glycan-C" charge="0.5973"/>
      <Atom name="O" type="glycan-O" charge="-0.5679"/>
```

I assume those are meant to be `"protein-N"`, etc.?  How did you handle this when you generated the system posted above?

zhang-ivy:
@peastman  Yes, this is the first issue I encountered when trying to convert the forcefields. 
For the short term (testing purposes), I've been generating my system using `ffxml/ff14SB.xml` [here](https://github.com/openmm/openmmforcefields/blob/master/amber/ffxml/ff14SB.xml)

For the long term, we'll have to make a decision on whether we should enforce protein atom names in ffxml/GLYCAM_06j-1.xml to contain the `protein-` prefix (to be used with protein.ff14SB)? Or if we should leave ffxml/GLYCAM_06j-1.xml as is and use ff14SB as the protein ff

peastman:
The problem seems to be from missing definitions.  I examined one particular bond that is present in `amber_after.xml` but not in `omm_after.xml`: the bond between atom 155 (atom N of NLN) and atom 157 (atom CA of NLN).  The types of those atoms are `N` (which has class `N`) and `glycan-CX` (which has class `CX`).  Looking at the `<HarmonicBondForce>` tag, there are no definitions at all involving class `CX`.

> For the long term, we'll have to make a decision on whether we should enforce protein atom names in ffxml/GLYCAM_06j-1.xml to contain the protein- prefix (to be used with protein.ff14SB)?

The prefix is essential.  The force field won't work correctly without it.  See my comments at https://github.com/ParmEd/ParmEd/issues/1136#issuecomment-780222890.

zhang-ivy:
>The prefix is essential.

Isn't the prefix only essential when using `protein.ff14SB.xml`? As I mentioned above, I was using `ff14SB.xml` to parametrize the above systems, so the atom types do not contain `protein-` prefixes

peastman:
Looking at GLYCAM_06j.dat, I see that it also contains no parameters for bonds involving `CX` atoms.  Where are those parameters coming from?

peastman:
The prefix is essential when combining multiple types of molecules.  The protein and DNA force fields each define a type called "C", for example.  Without the prefixes, OpenMM will report an error when you try to combine them, since you aren't allowed to define two different types with the same name.

zhang-ivy:
At the top of `GLYCAM_06j.dat`, `CX` is defined as:
```
CX 12.01                             protein C-alpha (new to ff10)  copied from parm10
```
I also don't see bond parameters defined for CX, but I do see angle and torsion parameters. Not sure why this is. Perhaps we `CX`s need to be interpreted as `CA`s?

peastman:
That's defining the atom type.  Scroll down to the next section which defines bond parameters.  There are no definitions involving CX.

peastman:
Sorry, missed the second half of your post.  I don't think Amber supports any kind of subclassing for atom types?  If an atom has type CX, so far as I know it will only use definitions for that specific type.

zhang-ivy:
>The prefix is essential when combining multiple types of molecules. The protein and DNA force fields each define a type called "C", for example. Without the prefixes, OpenMM will report an error when you try to combine them, since you aren't allowed to define two different types with the same name.

For the test case above, I noted that I am only using ff14SB.xml and GLYCAM_06j-1_trunc.xml, so there won't be any naming clashes for parametrizing this system. 
For long term/general use cases, I can add the `protein-` prefix, to protein atoms in the glycan ff, but I just want to clarify that there should not be any prefix clashes in the test case we are discussing here.

peastman:
I think I see.  The problem is in the prefixes.  CX is *not* a glycam type.  It's the standard type used for protein CA atoms.  So the type of that atom should be `protein-CX`, not `glycan-CX`.

In the definition of the NLN residue, there's a mix of atoms with and without the `glycan-` prefix.  How did you add the prefixes?

zhang-ivy:
> I don't think Amber supports any kind of subclassing for atom types? If an atom has type CX, so far as I know it will only use definitions for that specific type.

That sounds right. I'm not sure how to proceed from here. 
@jchodera: As Peter points out above, the current showstopper for glycan forcefield conversion is that there are no bond parameters for atoms involving `CX` atoms in GLYCAM_06j.dat, hence these parameters are also missing in the converted version of the forcefield.

zhang-ivy:
>How did you add the prefixes?

I used this function in openmmforcefields: https://github.com/openmm/openmmforcefields/blob/master/amber/convert_amber.py#L548

peastman:
Ok, I think that clears it up.  We already knew that method [would need to change](https://github.com/ParmEd/ParmEd/pull/1149#issue-575140642) to support this force field.

zhang-ivy:
Ok, it seems like there are a lot of atom types that appear in both `GLYCAM_06j.dat` and ff14SB.xml. Is the desired behavior: iterate through all atom types in `GLYCAM_06j.dat` and if the atom type is present in ff14SB, add the `protein-` prefix, otherwise add the `glycan-` prefix?

peastman:
I suggest we merge this, then open a separate issue to discuss prefixes.  There are several issues that need to be considered, and it may require changes in both ParmEd and openmmforcefields.

zhang-ivy:
Sure, I'd be happy to open an issue and start the discussion. My only hesitation with merging this PR is that the nonbonded energies are still very discrepant: https://github.com/ParmEd/ParmEd/pull/1149#issuecomment-810430224

peastman:
That's expected.  The code to handle scaling of 1-4 interactions depends on atom types.  Since the prefixes are currently being applied inconsistently, it won't work correctly.

zhang-ivy:
Right, but we should probably wait until the atom type prefixes have been fixed (and then run my test again to confirm that the energy discrepancy is no longer present) before merging this. Otherwisse, we cannot assess whether the changes in this PR are scaling energies appropriately.

zhang-ivy:
@peastman: I did not open a new issue to discuss the prefix issue, but I tagged you in an existing issue in `openmmforcefields` [here](https://github.com/openmm/openmmforcefields/pull/156).

peastman:
This PR wasn't intended to fix all problems with GLYCAM, just one particular one.  Getting it working will take multiple PRs across multiple repositories.  That's going to be hard to do if we don't merge any of them until all of them are finished.

I already verified that this produces correct scaling when given a force field where all atom types are defined correctly.

peastman:
Is this ok to merge?

zhang-ivy:
Yes go ahead!

peastman:
@swails is the one who can merge it.

swails:
Merged.

PR #1157, opened 2021-03-07
Add support for more CHARMM force fields

swails:
* Adds support for Drude force fields
* Adds support for lone pairs (virtual sites) in PSF parsing

These enhancements were made to an earlier version of the source code found in OpenMM, and this PR is aimed at transferring those enhancements so that ParmEd is capable of every feature supported by OpenMM.


swails:
I think I'm close -- I'm basically working to make sure that OpenMM and CHARMM yield the same OpenMM system from the same files (the OpenMM code is an old version of the ParmEd code, so it's not super straightforward to translate the changes there to here).

I just have to find the time to work on it.  Examples help.

PR #1162, opened 2021-03-20
Improving parameterized inferrence

Andrew-AbiMansour:
In [parmed.structure.Structure._ _repr_ _](https://github.com/ParmEd/ParmEd/blob/master/parmed/structure.py#L293), parameterization is inferred from bonds/bond types. However, for a system of non-bonded parameterized atoms, Structure appears unparameterized. An improvement is suggested here based on atom types.

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -307,9 +307,9 @@ class Structure:
                 retstr.append(' (triclinic)')
             else:
                 retstr.append(' (orthogonal)')
-        # Just assume that if the first bond has a defined type, so does
+        # Just assume that if the first atom has a defined type, so does
         # everything else... we don't want __repr__ to be super expensive
-        if len(self.bonds) > 0 and self.bonds[0].type is not None:
+        if self.atoms[0].type:
             retstr.append('; parametrized>')
         else:
             retstr.append('; NOT parametrized>')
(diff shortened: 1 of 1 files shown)


parmed/structure.py
```diff
@@ -307,12 +307,12 @@ def __repr__(self):
                 retstr.append(' (triclinic)')
             else:
                 retstr.append(' (orthogonal)')
-        # Just assume that if the first bond has a defined type, so does
+        # Just assume that if the first atom has a defined type, so does
         # everything else... we don't want __repr__ to be super expensive
-        if len(self.bonds) > 0 and self.bonds[0].type is not None:
-            retstr.append('; parametrized>')
+        if self.atoms[0].type:
```
swails, line 312:
```suggestion
        if self.atoms[0].atom_type is not UnassignedAtomType and self.atoms[0]._epsilon is None:
```

This is probably a better heuristic to use.  It's possible to provide parameters without defining an explicit type string, but you really can't provide parameters and have the atoms have an UnassignedAtomType *and* the nonbonded L-J terms be not set.

Andrew-AbiMansour (commented):
Bump

Andrew-AbiMansour:
I went with the most logical criterion based on your suggestion:

```python
if self.atoms[0].atom_type is UnassignedAtomType and self.atoms[0]._epsilon is None:
    retstr.append('; NOT parameterized>')
else:
    retstr.append('; parameterized>')
```

swails (approved):
Thanks!

PR #1165, opened 2021-04-07
fix wrong dim in box estimation

Andrew-AbiMansour:
In the [GromacsGroFile.write](https://github.com/Andrew-AbiMansour/ParmEd/blob/master/parmed/gromacs/gromacsgro.py#L196) method, there's a logical bug in the way the box is estimated from `struct.coordinates` when it is undefined (incorrect dimension is used).


swails (approved):
Thanks!

PR #1170, opened 2021-04-23
missing check for DihedralType in Dihedral.funct

andrew-abimansour:
The `funct` property in the [Dihedral](https://github.com/ParmEd/ParmEd/blob/master/parmed/topologyobjects.py#L2138) class can return incorrect `self._funct` when `self.type` is `DihedralType`. This ensures `Dihedral.funct` checks for `DihedralType` and subsequently returns 1.


swails (approved):
LGTM, thanks!

PR #1172, opened 2021-05-12
Add explicit hydrogens when loading in SMILES

justinGilmer:
Previously, if a user decided to use `parmed.rdkit.load_smiles` to
generate a `parmed` structure, no explicit hydrogens are added to the
structure, this can be an issue if trying to generate hydrocarbon
systems.

This PR leverages the built-in `rdkit` method `rdkit.Chem.AddHs` to generate explicit
hydrogens for an rdkit mol. An additional boolean flag: `hydrogens=True` has been added to this method. Since this could break others' workflows, the default value is `False` to comply with current functionality.

### DIFF
--- a/parmed/rdkit/rdkit.py
+++ b/parmed/rdkit/rdkit.py
@@ -27,7 +27,7 @@ class RDKit:
         return PDBFile.parse(fh)
 
     @staticmethod
-    def from_smiles(smiles, coordinates=True):
+    def from_smiles(smiles, coordinates=True, hydrogens=False):
         """
         Load smiles string to :class:`Structure`
 
@@ -36,6 +36,8 @@ class RDKit:
         smiles : str, smiles
         coordinates : bool, default True
             if True, use `rdkit.Chem.AllChem.EmbedMultipleConfs to assign coordinates
+        hydrogens : bool, default False
+            if True, use `rdkit.Chem.AddHs` to generate explicit hydrogens
 
         Returns
         -------
@@ -44,6 +46,9 @@ class RDKit:
         from rdkit import Chem
         from rdkit.Chem import AllChem
         mol = Chem.MolFromSmiles(smiles)
+
+        if hydrogens:
+            mol = Chem.AddHs(mol)
 
         if coordinates:
             AllChem.EmbedMultipleConfs(mol, useExpTorsionAnglePrefs=True, useBasicKnowledge=True)
--- a/parmed/rdkit/rdkit.py
+++ b/parmed/rdkit/rdkit.py
@@ -36,7 +36,7 @@ class RDKit:
         smiles : str, smiles
         coordinates : bool, default True
             if True, use `rdkit.Chem.AllChem.EmbedMultipleConfs to assign coordinates
-        hydrogens : bool, default False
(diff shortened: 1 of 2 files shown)


parmed/rdkit/rdkit.py
```diff
@@ -36,6 +36,8 @@ def from_smiles(smiles, coordinates=True):
         smiles : str, smiles
         coordinates : bool, default True
             if True, use `rdkit.Chem.AllChem.EmbedMultipleConfs to assign coordinates
+        hydrogens : bool, default False
+            if True, use `rdkit.Chem.AddHs` to generate explicit hydrogens
```
swails, line 40: IMO `hydrogens=True` is the better default (I know it's not strictly backwards-compatible, but the next release will be a major 4.0 release anyway, so let's get all of the breaking changes in now that we can).

```suggestion
        hydrogens : bool, default True
            if True, use `rdkit.Chem.AddHs` to generate explicit hydrogens
```

parmed/rdkit/rdkit.py
```diff
@@ -27,7 +27,7 @@ def load(rmol):
         return PDBFile.parse(fh)
 
     @staticmethod
-    def from_smiles(smiles, coordinates=True):
+    def from_smiles(smiles, coordinates=True, hydrogens=False):
```
swails, line 30:
```suggestion
    def from_smiles(smiles, coordinates=True, hydrogens=True):
```

test/test_parmed_rdkit.py
```diff
@@ -35,3 +35,13 @@ def test_load_smiles(self):
         # coordinates = True (default)
         parm = pmd.rdkit.from_smiles(smiles)
         np.testing.assert_allclose(parm.coordinates[0], [-1.072,  0.829 ,  0.108])
+
+    def test_load_smiles_explicit_hydrogen(self):
```
swails, line 39: The other tests would need `hydrogens=False` added to the `from_smiles` calls to continue passing with the provided recommendation.

swails (commented):
Thank you! This is a nice addition!

I actually think the functionality you added should be made the default and people that use it will have to deal with the backwards incompatibility (there are several in the upcoming 4.0 release).

test/test_parmed_rdkit.py
justinGilmer: Sounds great! I’ll get those necessary changes in, in the morning.

parmed/rdkit/rdkit.py
justinGilmer: I agree. Personally, when I’m using SMILES strings, they are pretty much always some carbon based structures, and adding explicit hydrogens just makes life so much easier.

justinGilmer:
@swails , I went through and made the necessary changes discussed above and the units tests are passing locally for me. There was very little modification of the codebase to support these changes 🎉 .

Let me know if there is anything else needed on my end! It seems like for builds to run it needs approval from a maintainer: https://github.com/ParmEd/ParmEd/actions/runs/836655356

swails:
Thanks! I approved and will merge when the tests pass.

PR #1174, opened 2021-05-24
Fix/openmm namespace change

mikemhenry:
Like https://github.com/ParmEd/ParmEd/pull/1173 this PR address the namespace changes coming to openMM, but does this in a backwards compatible way. The main issue I ran into using the openMM dev build was the logic on how parmed checks the version/capabilities of openMM.

```
>>> from simtk.openmm.app.internal.unitcell import reducePeriodicBoxVectors
Warning: importing 'simtk.openmm' is deprecated.  Import 'openmm' instead.
  File "<stdin>", line 1, in <module>
ModuleNotFoundError: No module named 'simtk.openmm.app.internal'
>>> from openmm.app.internal.unitcell import reducePeriodicBoxVectors
>>> 
```

I'm not sure when the next parmed release is planned, but this change is backwards compatible with other openmm versions and perhaps is a good stopgap for users that want to use the latest version of both openmm and parmed.


mikemhenry:
had an issue with my git history, fixed in 1176

PR #1177, opened 2021-05-27
Fix/openmm namespace change

mattwthompson:
Seeing if #1176 will run without maintainer approval if I pull it into my fork


mattwthompson:
This managed to get tested started over there - I guess me being in the git history was enough to get through GitHub's checks.

PR #1182, opened 2021-07-02
Base unscaled pairs on all atoms in a torsion

peastman:
When I implemented unscaled 1-4 interactions in #1149, I based it on the assumption that pairs could be classified based just on the types of the two interacting atoms, ignoring the two other atoms that connect them.  This turns out to be wrong.  For example, O-C-CX-H1 is an ordinary backbone torsion that uses standard scaling factors.  By O-C-Cg-H1 is a GLYCAM torsion that should be unscaled.  You need to consider all four atoms to tell the difference.  This changes it to do that.


peastman:
Test results are in https://github.com/openmm/openmmforcefields/pull/156#issuecomment-874248395.  With this change, we get excellent agreement on energies.

swails:
Thanks!

PR #1183, opened 2021-07-06
Recalculate 1-4 nonbonded terms in combining_rules setter

daico007:
Currently, if the `Structure.adjusts` section is pre-calculated, they will not be recalculated when the `Structure.combining_rule` is updated later. This PR add an option in the `combining_rule` setter method to recalculate sigmas in `Structure.adjusts` to match with the new combining rule. Inspired by @mattwthompson patch in https://github.com/mosdef-hub/foyer/pull/433.

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -1589,8 +1589,9 @@ class Structure:
                 combine = lambda sig1, sig2: (sig1 + sig2)/2
             elif thing == "geometric":
                 combine = lambda sig1, sig2: (sig1 * sig2)**2
-
-            for adj in structure.adjusts:
+            if any((not np.isclose(adj.type.sigma, combine(adj.atom1.sigma_14, adj.atom2.sigma_14))) for adj in self.adjusts):
+                raise ValueError("Cannot change the combining rule with modified off-diagonal 1-4 Lennard-Jones parameters")
+            for adj in self.adjusts:
                 adj.type.sigma = combine(adj.atom1.sigma_14, adj.atom2.sigma_14)
 
 
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -1577,7 +1577,18 @@ class Structure:
     def combining_rule(self, thing):
         if thing not in ('lorentz', 'geometric'):
             raise ValueError("combining_rule must be 'lorentz' or 'geometric'")
-        self._combining_rule = thing
+        if self._combining_rule != thing
+            self._combining_rule = thing
+            if thing == 'lorentz':
+                for adj in structure.adjusts:
+                    sig1 = adj.atom1.sigma
+                    sig2 = adj.atom2.sigma
+                    adj.type.sigma = (sig1 + sig2)/2
+            elif thing == 'geometric':
+                for adj in structure.adjusts:
+                    sig1 = adj.atom1.sigma
+                    sig2 = adj.atom2.sigma
+                    adj.type.sigma = (sig1 * sig2) ** 0.5
 
     #===================================================
 
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -1574,11 +1574,11 @@ class Structure:
         return self._combining_rule
 
     @combining_rule.setter
-    def combining_rule(self, thing):
+    def combining_rule(self, thing, recalculate_14=False):
         if thing not in ('lorentz', 'geometric'):
             raise ValueError("combining_rule must be 'lorentz' or 'geometric'")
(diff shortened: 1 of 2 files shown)


parmed/structure.py
```diff
@@ -1574,10 +1574,21 @@ def combining_rule(self):
         return self._combining_rule
 
     @combining_rule.setter
-    def combining_rule(self, thing):
+    def combining_rule(self, thing, recalculate_14=False):
```
swails, line 1577: This is a setter, so there's no way to actually specify `recalculate_14=True` as written.  Also, `Atom` has a `sigma_14` attribute that specifies the 1-4 sigma parameter (which may be different than its "normal" sigma parameter).

Instead, I'd do something like this:

```python
@combining_rule.setter
def combining_rule(self, thing):
    if self._combining_rule == thing:
        # nothing to do
        return
    if self.has_NBFIX():
        raise ValueError("Cannot change the combining rule with modified off-diagonal Lennard-Jones parameters")

    if thing == "lorentz":
        combine = lambda sig1, sig2: (sig1 + sig2) / 2
    elif thing == "geometric":
        combine = lambda sig1, sig2: (sig1 * sig2) ** 0.5
    if any(abs(adj.type.sigma - combine(adj.atom1.sigma_14, adj.atom2.sigma_14)) > TINY):
        raise ValueError("Cannot change the combining rule with modified off-diagonal 1-4 Lennard-Jones parameters")
    for adj in structure.adjusts:
        adj.type.sigma = combine(adj.atom1.sigma_14, adj.atom2.sigma_14)
```

swails (commented):
Good catch!  I made a suggestion for a change that's required for this code to do anything (and a correction for 1-4 interactions).

parmed/structure.py
daico007: Good catch! Thank you for the feedback! I only have one question about this line: 
```python3  
if any(abs(adj.type.sigma - combine(adj.atom1.sigma_14, adj.atom2.sigma_14)) > TINY for adj in self.adjusts):
        raise ValueError("Cannot change the combining rule with modified off-diagonal 1-4 Lennard-Jones parameters")
```
it looks like the `if` statement is checking if the newly calculated sigma would be much different than those currently stored in `adj.type.sigma`, but I am not sure why this will raise the ValueError below.

swails: Basically what's happening here is we are throwing a ValueError if there are any off-diagonal modifications of 1-4 nonbonded interactions.  The Lennard-Jones parameter matrix is built by computing pairwise "combined" sigma and epsilon properties.  NBFIX (nonbond "fixes") allow you flexibility in fitting the Lennard-Jones potential by changing "off-diagonal" elements of the parameter matrix that *doesn't* obey the normal combining rules (CHARMM uses this to tune specific pairs of atomic interactions without affecting all of the other interactions).

However, if you change the combining rules and "recompute" the sigmas on the 1-4 adjustments, you would wipe out any specific pair overrides (NBFIXes) you implemented.  It's unclear exactly what should be done if there is NBFIX implemented:

* Do you keep the pair-specific parameters used to override the combining rules?
* Or do you overwrite them by applying the combining rules to all pairs?
* Or do you throw an error and force the developer to resolve the issue?

There are reasonable arguments to support either of the first two options (leave them alone or overwrite), and large numbers of people could (justifiably) be surprised by either of the first two behaviors.  So the only safe option is to do the third, which will ensure nobody gets surprised, even if some people are slightly inconvenienced.

But if you go through the effort of NBFIXing certain interactions, you aren't likely to change the combining rule so this safety valve probably shouldn't hit anybody (because nobody should really try to change the combining rule *after* NBFIXing).  But just in case, it's worth throwing an exception to make sure surprising (but subtle and hard-to-detect) doesn't slip through unprotected.

daico007: Gotcha. I added that checks to the methods

daico007: Spending some more thought on this, I am still a bit confusing by the the last check (`if any(abs(adj.type.sigma - combine(adj.atom1.sigma_14, adj.atom2.sigma_14)) > TINY for adj in self.adjusts):`). If we are worried about the NBFix implemented by the user, shouldn't it has already captured by the first condition check (`if self.has_NBFIX():`)? 
With the last check being: "Error out if if any of the existing `adj.type.sigma` differ by the newly calculated `sigma`", it kind of defeat the purpose of this PR in updating the sigma if the user update the overall combining rule of the structure.

PR #1184, opened 2021-07-13
Perceive external bonds (necessary for glycam ffxml conversion)

zhang-ivy:
@peastman and I have been working on converting the AMBER glycam forcefield (`GLYCAM_06j-1`) into a ffxml for the OpenMM ecosystem -- we have the conversion pipeline working now (after making a number of changes to Parmed, OpenMM, and soon Openmmforcefields), but there is one more ParmEd fix that I had to use for the conversion pipeline that has not been pushed.

After generating the ffxml and upon trying to create a system for a glycosylated protein, I was getting errors like :
```
ValueError: No template found for residue 197 (UYB). The set of atoms matches UYB, but the bonds are different.
```
This was because the ExternalBonds in the ff templates were not correct. In the converted ffxml, for the UYB residue, I was seeing:
```
     <ExternalBond atomName="C1"/>
     <ExternalBond atomName="O4"/>
```

But there should be a third line for atomName="O6".

Since Parmed was not writing out all of the atoms with external bonds correctly, I added a fix to `parmed/openmm/parameters.py`. This fix involves adding a function called `_get_atoms_with_external_bonds()` that gets atoms with external bonds.

Not sure if this fix should be merged into master as I'm not sure whether it will disrupt the current functionality of `parmed/openmm/parameters.py`, but this is necessary for glycam ffxml conversion, so we should brainstorm how to include it in the repo.

Tagging @jchodera @peastman @swails for feedback

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -276,7 +276,7 @@ class OpenMMParameterSet(ParameterSet, CharmmImproperMatchingMixin, metaclass=Fi
         return atom.type
 
     def write(self, dest, provenance=None, write_unused=True, separate_ljforce=False,
-              improper_dihedrals_ordering='default', charmm_imp=False, skip_duplicates=True):
+              improper_dihedrals_ordering='default', charmm_imp=False, skip_duplicates=True, is_glycam=False):
         """ Write the parameter set to an XML file for use with OpenMM
 
         Parameters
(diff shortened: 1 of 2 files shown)


peastman:
This change looks like it's very specific to GLYCAM.  We should make sure it doesn't break other force fields.  For example, if you run CHARMM36 through it, does this cause it to start inserting external bonds that shouldn't be there?

Possibly this could be controlled by a global option, so you need to specifically enable it when running ParmEd.

zhang-ivy:
The tests were failing for this because I've only allowed external bond perception for `{'C': 4, 'N': 3, 'O': 2, 'H': 1}` -- any other atom types will throw an error. I've checked the glycam ffxml to see if any residues contain S or P and only residue SO3 contains S (no residues contain P). I've included 'SO3' as a special case in the function.

To make sure my fix doesn't affect other conversion of other forcefields, I've also implemeted Peter's suggestion of using a global flag to control whether the external bond perception code is executed.

zhang-ivy:
@peastman @swails Could one of you approve this PR for running test workflows (so that we can check that these changes don't disrupt the behavior for other forcefields)?

peastman:
I'm not a maintainer.  @swails will have to do it.

swails (approved):

PR #1186, opened 2021-07-28
Prototype for heterogenous mass repartitioning

bradakstx:
This is a draft implementation of the two-tier mass repartitioning
strategy recently published by the Sugita group. The main change is that
an additional "alternate" target hydrogen mass is supplied for hydrogens
that are connected to a ring. The complete strategy proposed for AMBER
force fields is:

1. reduce the target mass from 3 x 1.008 to 2.5 x 1.008
2. use an alternate target mass of 2 x 1.008

Changes to the code:

- A new optional keyword `altmass` has been added to the HMassRepartition
action that supplies the alternative mass.
- A quick scheme to produce rdkit mol objects from single-residue ParmEd
molecules has been added so that we can access the IsInRing() method for
determining which atoms are in rings.

Things to improve:

The use of rdkit molecules seems to work fine in practice, but is
obviously terribly inefficient -- a new molecule is created for each and
every atom to check rather than the other way around. I can think of
multiple ways to restructure this, but I'll kick that to the development
community. The current implementation seems to work very robustly
regardless.

### DIFF
--- a/parmed/tools/actions.py
+++ b/parmed/tools/actions.py
@@ -3291,23 +3291,15 @@ class HMassRepartition(Action):
         # coordination of four). This quickly avoids silly cases like
         # accidentally flagging a rigid water as a three-membered ring.
         #
-        num_hyd_partners =\
-                sum((bp.atomic_number == 1) for bp in atom.bond_partners)
+        num_hyd_partners = sum((bp.atomic_number == 1) for bp in atom.bond_partners)
         num_hvy_partners = len(atom.bond_partners) - num_hyd_partners
         if not (num_hvy_partners >= 2 and num_hyd_partners in [1, 2]):
             return False
-        # Use rdkit Atom.IsInRing() method.
-        #
+
         rmol = self._residue_to_rdkit(atom.residue)
-        for patom, ratom in zip(atom.residue.atoms, rmol.GetAtoms()):
-            if patom != atom:
-                continue
-            if ratom.IsInRing():
-                return True
-            else:
-                return False
-        # This should really never be reached...
-        return False
+        for pa, ra in zip(atom.residue.atoms, rmol.GetAtoms()):
+            if pa == atom:
+                return ra.IsInRing()
 
     def execute(self):
         # Back up the masses in case something goes wrong
--- a/parmed/tools/actions.py
+++ b/parmed/tools/actions.py
@@ -2,6 +2,7 @@
 All of the prmtop actions used in PARMED. Each class is a separate action.
 """
 import copy
+import io
 import math
 import os
 import sys
@@ -10,6 +11,7 @@ from collections import Counter, OrderedDict
 from glob import glob
 
 import numpy as np
+import rdkit.Chem as Chem
 
 from .. import gromacs, unit as u, read_PDB
 from ..amber import (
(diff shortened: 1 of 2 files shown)


parmed/tools/actions.py
```diff
@@ -3243,13 +3246,62 @@ class HMassRepartition(Action):
     def init(self, arg_list):
         self.changewater = arg_list.has_key('dowater')
         self.new_h_mass = arg_list.get_next_float(optional=True, default=3.024)
+        self.alt_new_h_mass = arg_list.get_key_float('altmass', default=None)
```
swails, line 3249:
```suggestion
        self.alt_new_h_mass = arg_list.get_key_float('ring_hmass', default=None)
```

parmed/tools/actions.py
```diff
@@ -3243,13 +3246,62 @@ class HMassRepartition(Action):
...
         return retstr + 'Not changing water hydrogen masses.'
 
+    def _residue_to_rdkit(self, residue):
+        """Extract a residue as a standalone rdkit Mol object."""
+        # Hack to make sure that atom types are compliant with rdkit
+        # expectations. That is, rdkit must be able to infer the element based
+        # solely on the atom type, which is not always true for things like
+        # standard GAFF types.
+        #
+        mol = self.parm[':%d'%(residue.number + 1)]
```
swails, line 3264:
```suggestion
        mol = self.parm[residue.id,:]
```

parmed/tools/actions.py
```diff
@@ -3243,13 +3246,62 @@ class HMassRepartition(Action):
...
+        for atom in mol.atoms:
+            atom.type = element_by_name(atom.name)
```
swails, line 3266:
```suggestion
            atom.type = _Element[atom.atomic_number]
```

parmed/tools/actions.py
```diff
@@ -23,6 +25,7 @@
 from ..formats.registry import load_file
 from ..modeller import AmberOFFLibrary, ResidueTemplateContainer
 from ..periodic_table import Element as _Element
+from ..periodic_table import element_by_name
```
swails, line 28:
```suggestion
```

parmed/tools/actions.py
```diff
@@ -3243,13 +3246,62 @@ class HMassRepartition(Action):
...
+        num_hvy_partners = len(atom.bond_partners) - num_hyd_partners
+        if not (num_hvy_partners >= 2 and num_hyd_partners in [1, 2]):
+            return False
+        # Use rdkit Atom.IsInRing() method.
+        # This is obviously really inefficient since each atom in a residue
+        # will cause the _same_ molecule object to be built. It doesn't seem to
+        # be a terrible speed issue for now and it can always be restructured
+        # later.
+        #
+        rmol = self._residue_to_rdkit(atom.residue)
```
swails, line 3294: you could `lru_cache` this (after adding a `__hash__` implementation to `class Residue` that simply returns `id(self)`).

This algorithm is O(N^2) in # of atoms, but adding the cache would drop this to O(N).

swails: Lookup table is already there.

swails:
Can we add a test case so when I finally *do* get around to adding ring perception to ParmEd I can make sure I don't break this?

parmed/tools/actions.py
```diff
@@ -3243,17 +3249,84 @@ class HMassRepartition(Action):
...
+                    if self.is_in_ring(heteroatom):
+                        return self.new_ring_h_mass
+                    else:
+                        return self.new_h_mass
+            else:
+                # Cannot enable heterogeneous partitioning. Issue a warning.
+                warnings.warn(
+                    'Cannot use heterogeneous mass repartitioning without '
+                    'rdkit installed.\nContinuing with standard scheme.'
+                )
```
swails, line 3328:
```suggestion
                raise ImportError("Heterogeneous mass repartitioning requires rdkit.")
```

This warning will not be seen in an automated workflow and will result in broken behavior relative to what is expected.

swails (approved):

PR #1200, opened 2021-10-04
Try to fix osx builds more permanently

### LINKED ISSUE #1193: darwin_major table in setup.py is outdated
In setup.py, the MACOS_DEPLOYMENT_TARGET is determined by the following code in `setup.py`:
```def prepare_env_for_osx():
    """ Prepares the environment for OS X building """
    darwin_major_to_osx_map = {
        '11': '10.7',
        '12': '10.8',
        '13': '10.9',
        '14': '10.10',
        '15': '10.11',
        '16': '10.12',
        '17': '10.13',
        '18': '10.14',
        '19': '10.15',
    }
```
This table is outdated, (and this will be a recurring problem).  My darwin major for OSX 11.5 is "20".  I'm not really sure
what deployment target I should set: `11.5`?

Not setting the deployment target leads to the dreaded "locale not found" error.  Adding a 20 to the table (pointing to 11.5) eliminates that problem.

p.s. for @hainm:  similar problems arise for `pytraj`:  I haven't yet investigated those....

swails:
Fixes #1193

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -3,6 +3,7 @@ import shutil
 import sys
 import versioneer
 import struct
+import warnings
 
 if sys.version_info < (3, 6):
     sys.exit('You must have at least Python 3.6 for ParmEd to work correctly.')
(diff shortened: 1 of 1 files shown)


dacase:
It's not really a "permanent" fix, since the table still needs updating with every Apple release. :-(

Should add `'21': '12.0',` to the table for Monterey.

In the old days, each "minor" release (e.g. 10.14, 10.15) got its own entry, but now it seems (?) that the table needs to be updated just once for each major release.  [Not sure about this.]

PR #1201, opened 2021-10-04
Converted AMOEBA code to work with OpenMM 7.6

peastman:
I copied the implementations of the forces from ForceField and verified that the test cases pass.

### DIFF
--- a/test/test_openmm_amber.py
+++ b/test/test_openmm_amber.py
@@ -1706,7 +1706,6 @@ class TestChamberParm(TestCaseRelative):
                 parm.createSystem(nonbondedMethod=nbmethod)
 
 @unittest.skipUnless(has_openmm, 'Cannot test without OpenMM')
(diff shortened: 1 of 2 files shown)


test/test_openmm_amber.py
```diff
@@ -1713,12 +1713,12 @@ class TestAmoebaParm(TestCaseRelative):
     def test_amoeba_forces(self):
```
swails, line 1713: Can you remove the `@pytest.mark.xfail` line above the `TestAmoebaParm` class?  That line makes failed tests in this test class non-fatal (in the test report, it marks tests that fail with an `x` but still reports a successful run, or with a capital `X` if the test actually passed).

https://docs.pytest.org/en/latest/how-to/skipping.html#xfail-mark-test-functions-as-expected-to-fail

swails:
Thanks!

PR #1206, opened 2021-10-28
Add fixes to external bond perception for glycans

zhang-ivy:
This PR includes the following changes that fix external bond perception for glycans:
- Add exception cases (OME, OLP) to external bond perception function 
- Only designate an atom as containing an external bond if the number of bonds is less than the expected valence
- Split original external bond perception code from glycan code in _write_omm_residues()

### DIFF
--- a/parmed/openmm/parameters.py
+++ b/parmed/openmm/parameters.py
@@ -726,7 +726,6 @@ class OpenMMParameterSet(ParameterSet, CharmmImproperMatchingMixin, metaclass=Fi
                     etree.SubElement(xml_residue, 'ExternalBond', atomName=residue.tail.name)
             else:
                 external_bonds = self._get_atoms_with_external_bonds(name)
-                external_bonds = [external_bonds[0], external_bonds[-1]] + external_bonds[1:-1] if len(external_bonds) > 2 and len(external_bonds[-1].name) <= 2 else external_bonds # Hack to get rid of atom ordering diffs
                 for atom in external_bonds:
                     etree.SubElement(xml_residue, 'ExternalBond', atomName=atom.name)
             if residue.name in valid_patches_for_residue:


zhang-ivy:
@jchodera @peastman : This PR is necessary for https://github.com/openmm/openmmforcefields/pull/180#issuecomment-962498498 (which also needs to be reviewed and merged)
Could you review when possible?

zhang-ivy:
I'd like to note that in order to preserve atom ordering in https://github.com/openmm/openmmforcefields/pull/180#issuecomment-962498498, I had to add the following line (after line 729) to this branch (which I don't think should be included/merged):
```python
external_bonds = [external_bonds[0], external_bonds[-1]] + external_bonds[1:-1] if len(external_bonds) > 2 and len(external_bonds[-1].name) <= 2 else external_bonds # Hack to get rid of atom ordering diffs
```

jchodera:
@zhang-ivy : Can you remind us if this is still relevant and what we need to do to review this? We can get a bugfix out quickly if needed.

zhang-ivy:
@jchodera : Yes this PR is necessary to properly convert amber glycam forcefields to openmm forcefield xmls. Just needs to be reviewed and merged, but no rush on getting a bug fix out quickly, since we already converted the forcefields and opened PRs 
- in openmm:https://github.com/openmm/openmm/pull/3303 
- in openmmforcefields: https://github.com/openmm/openmmforcefields/pull/156 and https://github.com/openmm/openmmforcefields/pull/177 and https://github.com/openmm/openmmforcefields/pull/180

parmed/openmm/parameters.py
```diff
@@ -667,7 +678,7 @@ def _get_atoms_with_external_bonds(self, residue_name):
                 element_name = atom.element_name
 
             # If the number of bonds within the residue does not equal the number of bonds the atom should have, the atom  has an external bond
```
swails, line 680: This comment looks out of date now?  Why this change?

zhang-ivy: Yes, it's out of date. I changed this because it was a mistake in my original code -- we are assuming that any extra bonds are external bonds. If there aren't extra bonds, there are no external bonds.

parmed/openmm/parameters.py
```diff
@@ -706,17 +717,18 @@ def _write_omm_residues(self, xml_root, skip_residues, skip_duplicates, valid_pa
...
+            if not is_glycam:
+                for atom in residue.connections:
+                    etree.SubElement(xml_residue, 'ExternalBond', atomName=atom.name)
+                if residue.head is not None:
+                    etree.SubElement(xml_residue, 'ExternalBond', atomName=residue.head.name)
+                if residue.tail is not None and residue.tail is not residue.head:
+                    etree.SubElement(xml_residue, 'ExternalBond', atomName=residue.tail.name)
+            else:
                 external_bonds = self._get_atoms_with_external_bonds(name)
+                external_bonds = [external_bonds[0], external_bonds[-1]] + external_bonds[1:-1] if len(external_bonds) > 2 and len(external_bonds[-1].name) <= 2 else external_bonds # Hack to get rid of atom ordering diffs
```
swails, line 729: Can you move this to `_get_atoms_with_external_bonds`?  And what is the `len(external_bonds[-1].name)` checking here?

zhang-ivy: Sorry, that line was accidentally swept into the commit. It's a hack I used to preserve atom ordering, see more [here](https://github.com/ParmEd/ParmEd/pull/1206#issuecomment-963235048).

swails:
I'm a little concerned that GLYCAM support may risk being broken by refactors if we don't have a test case for it.  Even a regression test is enough (take a GLYCAM file and check that the generated parameter set matches something computed with this version of the code).

zhang-ivy:
@swails : Yes, understandable -- I need to refresh my memory on the glycam conversion pipeline, but I'm working on this.

zhang-ivy:
@swails : I've written the following test function (based on `openmmforcefields` code) that uses the attached input files:

```python
def check_glycam_conversion():
    """
    Takes leaprc and lib files and uses parmed to write out a ffxml. 
    
    Simplified from openmmtools/amber/convert_amber.py's `convert_leaprc()`
    
    Note that the output of this function is not the final GLYCAM_06j-1.xml in openmmforcefields/amber/ffxml nor 
    openmm/app/data/amber14/. Additional post processing steps are necessary to obtain the final ffxml version. 
    See `modify_glycan_ffxml()` in openmmforcefields/amber/convert_amber.py.
    
    """
    import os 
    import parmed
    from io import StringIO
    import re
    
    _loadoffre = re.compile(r'loadOff (\S*)', re.I)
    _sourcere = re.compile(r'source (\S*)', re.I)
    ignore = {'solvents.lib', 'atomic_ions.lib', 'ions94.lib', 'ions91.lib',
              'phosphoaa10.lib'}
    
    input_file = "glycam/leaprc.GLYCAM_06j-1"
    output_file = "glycam/GLYCAM_06j-1.xml"
    
    # Pre-process leaprc before converting
    with open(input_file, 'rt') as f:
        lines = [ line if '#' not in line else line[:line.index('#')] for line in f.readlines() ]
    new_lines = []
    for line in lines:
        if (ignore is not None and _loadoffre.findall(line) and
        _loadoffre.findall(line)[0] in ignore):
            continue
        new_lines += line
    leaprc = StringIO(''.join(new_lines))
    
    # Convert leaprc into ffxml
    override_level = 1 
    params = parmed.amber.AmberParameterSet.from_leaprc(leaprc)
    params = parmed.openmm.OpenMMParameterSet.from_parameterset(params, remediate_residues=(False))
    if override_level:
        for name, residue in params.residues.items():
            residue.override_level = override_level
    
    params.write(output_file, provenance=None, write_unused=True, improper_dihedrals_ordering='amber', skip_duplicates=False, is_glycam=True)
    
```

We'll want to check that the ffxml written by this function is the same as the attached file (`glycam/GLYCAM_06j-1.xml`). What's the easiest way to do this? Is there an easy way in parmed to load in 2 ffxmls and check that they contain the same parameters? Alternatively, I could look into ways for pythonic ways for checking whether 2 xml files are the same.
[glycam.zip](https://github.com/ParmEd/ParmEd/files/8647774/glycam.zip)

swails:
I'll add this.  Thanks!

PR #1222, opened 2022-02-15
Reading a Gromacs topology fails with insertion code or negative number

Valdes-Tresanco-MS:
Hi Jason. I found a bug in my previous PR. Although rare, negative residues numbers fail with this setup. So I replaced the conditional with a `try-except` clause. According to the tests I did, it should work correctly in all cases.

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -491,10 +491,10 @@ class GromacsTopologyFile(Structure):
             atom = Atom(atomic_number=atomic_number, name=words[4],
                         type=words[1], charge=charge, mass=mass)
         # check for insertion code
-        if words[2].isnumeric():
+        try:
             icode = ''
             resnum = int(words[2])
-        else:
+        except:
             icode = words[2][-1]
             resnum = int(words[2][:-1])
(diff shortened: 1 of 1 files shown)


parmed/gromacs/gromacstop.py
```diff
@@ -491,10 +491,10 @@ def _parse_atoms(self, line, params):
             atom = Atom(atomic_number=atomic_number, name=words[4],
                         type=words[1], charge=charge, mass=mass)
         # check for insertion code
-        if words[2].isnumeric():
+        try:
```
swails, line 494: It looks like the problem is that a leading `-` leads to `isnumeric()` returning `False`.  What about we change line 494 to this instead:

```python
if words[2].isnumeric() or (words[2].startswith('-') and words[2][1:].isnumeric()):
```

Valdes-Tresanco-MS: Done via [a8d0e3f](https://github.com/ParmEd/ParmEd/pull/1222/commits/a8d0e3fed7279bbcaba5f7453a70363325ff24c1)

swails:
Thanks!

PR #1226, opened 2022-02-15
New PBRadii sets

### LINKED ISSUE #1223: Add new PBRadii sets
Hi Jason. In our project, we have reviewed some recent studies where they propose a new set of PBRadii for GAFF and CHARMM. The new PBRadii sets would not overlap with the existing ones, but we would add them as new and with a different name. The purpose of this is that ParmEd has these options in the official repository and that we can use them with [gmx_MMPBSA](https://github.com/Valdes-Tresanco-MS/gmx_MMPBSA). 
We have seen that for CHARMM ff the same values are used as for Amber, which as mentioned in some mailing list threads, has not been extensively tested. We are implementing these features to do some further studies. So far, the results we get are consistent.

I can do a PR if you agree.

Papers:
- https://pubs.acs.org/doi/10.1021/jp970736r
- https://pubs.acs.org/doi/abs/10.1021/jp025852v
- https://pubs.acs.org/doi/10.1021/acs.jcim.1c00177
- https://pubs.acs.org/doi/full/10.1021/acs.jctc.9b00106

Sincerely
Mario S.

Valdes-Tresanco-MS:
Fix #1223

Add `mbondi_pb2`, `mbondi_pb3` and `charmm_radii`.


swails:
This looks like it wasn't prepped against the master branch

Valdes-Tresanco-MS:
Oh, my bad... I did it from the v3.4 branch... I'll push the changes to master as well

PR #1229, opened 2022-02-15
New pb radii sets for master branch

Valdes-Tresanco-MS:
Fix https://github.com/ParmEd/ParmEd/issues/1223

Add mbondi_pb2, mbondi_pb3 and charmm_radii.

Note: There are several differences between the master branch and v3.4. Is the difference due to some specific feature or is it that a merge between the two has not been done?


swails:
The 3.4 branch was created specifically to back-port some fixes before a large release that drops support for a lot of things.

Notably, the master branch has dropped support entirely for Python <3.6 (in particular Python 2.7 support is gone), and also drops support for OpenMM <7.6.  As part of this effort, I cleaned up a lot of code that was only needed to support legacy configurations.

I'm primarily waiting on time to finish up #1157 before cutting a 4.0 release and retiring the v3.4 branch.

Unfortunately the changes you made here will have to be copied into a new branch that branches off master.

PR #1230, opened 2022-02-16
PBRadii 

Valdes-Tresanco-MS:
Sorry for the previous PR, for some reason Pycharm was always creating the new branch from v3.4.

Fix https://github.com/ParmEd/ParmEd/issues/1223

Add `mbondi_pb2`, `mbondi_pb3` and `charmm_radii`.

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -401,7 +401,8 @@ class GromacsTopologyFile(Structure, TopFromStructureMixin, metaclass=FileFormat
                     dihedral_types = dict()
                     exc_types = dict()
                 elif current_section == 'atoms':
-                    molecule.add_atom(*self._parse_atoms(line, params))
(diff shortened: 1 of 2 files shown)


swails:
Thanks!

It is probably also worth adding a few tests here to support any future efforts to refactor.  That can be added later, though.

Valdes-Tresanco-MS:
Sure. We'll be working on that a bit later. Now we are preparing some examples for a possible problem that we have found in the conversion of topologies from Gromacs to Amber when cysteines are forming S-S. This is critical since the VDWAALS energies obtained using the tleap and parmed topologies differ considerably (~7000 kcal for the system in which we noticed the problem). The good news is that the delta (`Complex` - `Receptor` - `Ligand`) is the same for both cases.

I'll write this homework down to do later :)

swails:
Sounds like it's not creating the S-S bond

PR #1235, opened 2022-03-02
Added CONECT records

Valdes-Tresanco-MS:
This is a very basic implementation, so I'm not entirely convinced that it works in all cases. I look forward to your suggestions


Valdes-Tresanco-MS:
Sorry, the same thing is happening to me as the previous time with the creation of the branch.

PR #1236, opened 2022-03-03
Added CONECT records

Valdes-Tresanco-MS:
This is a very basic implementation, so I'm not entirely convinced that it works in all cases. I look forward to your suggestions

Mario S.

### DIFF
--- a/parmed/formats/pdb.py
+++ b/parmed/formats/pdb.py
@@ -782,9 +782,8 @@ class PDBFile(metaclass=FileFormatType):
     @staticmethod
     def write(struct, dest, renumber=True, coordinates=None, altlocs='all',
               write_anisou=False, charmm=False, use_hetatoms=True,
-              standard_resnames=False, increase_tercount=True, write_links=False):
+              standard_resnames=False, increase_tercount=True, write_links=False, conect=True):
         """ Write a PDB file from a Structure instance
-
         Parameters
         ----------
         struct : :class:`Structure`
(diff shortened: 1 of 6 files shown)


parmed/formats/pdb.py
```diff
@@ -837,7 +834,8 @@ def write(struct, dest, renumber=True, coordinates=None, altlocs='all',
             If True, any LINK records stored in the Structure will be written to
             the LINK records near the top of the PDB file. If this is True, then
             renumber *must* be False or a ValueError will be thrown
-
+        conect: bool, default True
+            Write CONECT records for S-S bridges and HETATM
```
swails, line 838:
```suggestion
            Write CONECT records for S-S bridges and HETATM

```

swails (commented):
This looks great!  Do you have a test case you can add for it?

Valdes-Tresanco-MS:
> This looks great! Do you have a test case you can add for it?

I'll try again with a direct case from the PDB and add it because the one I was using has some Na+ ions that also have records. However, I don't think ParmEd has the salt bridging atoms as bonded atoms. So, I will look for a suitable and easy to compare one.

Valdes-Tresanco-MS:
Here is a well-resolved structure with disulfide bonds. It also has a phosphate ion that serves for heteroatoms
Test code. 
Save conect records by default:
```python
import parmed

s = parmed.load_file('3m7q_test.pdb')
s.save('3m7q_parmed.pdb', format='pdb', renumber=False, overwrite=True)
```

Don't save conect records:
```python
import parmed

s = parmed.load_file('3m7q_test.pdb')
s.save('3m7q_parmed.pdb', format='pdb', renumber=False, overwrite=True, conect=False)
```



This file contains `3m7q.pdb` (Original PDB), `3m7q_test.pdb` (PDB without CONECT records) and `3m7q_parmed.pdb` (Saved pdb with CONECT records using this PR)
[parmed_conect_test.zip](https://github.com/ParmEd/ParmEd/files/8209809/parmed_conect_test.zip)

I look forward to your comments
Mario S.

PR #1259, opened 2022-07-26
Add a volume property

bradakstx:
I've been finding it useful to be able to quickly query the box volume
without too much fuss. This just requires a simple query of the box
(if it exists) and the box shape (ifbox). The only wrinkle here is the
general triclinic case where, I'll admit, I don't really know what the
proper math should be. Since that case is really rare, the
implementation here just raises a warning and then assumes a funky
truncated octahedron (as one might get from CHARMM-GUI, say).

I can easily add tests if this change is deemed acceptable.


drroe:
@bradakstx you can use the code from the cpptraj function `CalcFracFromUcell()` (https://github.com/Amber-MD/cpptraj/blob/4d031ad72dfa101b30d6c1820da632ff80e8d19f/src/Box.cpp#L318) to get the volume of a general triclinic cell. In this function `ucell` is a 3x3 matrix stored in a 1D array in row-major order (i.e. the unit cell vectors are a={0,1,2}, b={3,4,5}, c={6,7,8}.

bradakstx:
Thanks @drroe -- you could nullify the whole need for this function if that's available via pytraj!
 
My other question is that `AmberParm.box` is always a six element array with three lengths and three angles -- how does one convert that to the matrix you describe?

drroe:
>  you could nullify the whole need for this function if that's available via pytraj!

I think that it is: https://amber-md.github.io/pytraj/latest/analysis.html#pytraj.volume

> Nvm - answered my own question. You use Box::CalcUcellFromXyzAbg

Yep, that's the one!

parmed/amber/_amberparm.py
```diff
@@ -2286,6 +2286,62 @@ def box(self, value):
...
+            ucell[3] = y * cos(gam * deg2rad)
+            ucell[4] = y * sin(gam * deg2rad)
+        if not isclose(bet, 90.0):
+            ucell[6] = z * cos(alp * deg2rad)
+        yz_cos_alp = 0.0
+        if not isclose(alp, 90.0):
+            yz_cos_alp = x * y * cos(alp * deg2rad)
+        ucell[7] = (yz_cos_alp - ucell[6] * ucell[3]) / ucell[4]
+        ucell[8] = sqrt(z**2 + ucell[6]**2 + ucell[7]**2)
+        return ucell
```
swails, line 2318: Maybe you can re-use some of the functions in `geometry.py`?

bradakstx: Indeed I can.

The next item for discussion - For now I elected to return a float in A^3 rather than a Quantity. Is this objectionable? IMO the Quantity class is useful for internal calculations but I find it cumbersome to require an additional import just to interpret an attribute or method result.

bradakstx: Alternative -- Change `volume` attribute to `volume(use_units)` method and let the user decide.

swails: I've been trying to offer quantitied properties by prefixing with `u`, so `volume` would be a property returning a float as cubic angstroms while `uvolume` would return `volume * u.angstroms**3`.

parmed/amber/_amberparm.py
```diff
@@ -2286,6 +2286,28 @@ def box(self, value):
         else:
             self._box = np.array(value).reshape((-1, 6))[0]
 
+    @property
+    def volume(self):
```
swails, line 2290: Wouldn't it be better to put this on `Structure` instead?

bradakstx: Done.

parmed/amber/_amberparm.py
bradakstx: Done.

PR #1266, opened 2022-08-31
lonepair refinement

JohannesKarwou:
When reading in CHARMM psf files, I experienced problems for certain psf files. I included one, generated by CHARMM-GUI, which does not work as expected (loading with `load_psf` works fine, but processing this file or saving it with `save_psf` produces errors):

[step3_input.zip](https://github.com/ParmEd/ParmEd/files/9460858/step3_input.zip)

I think the problem is, that `_process_lonepair_section` is only executed, if there is a `MOLNT` section, which is sometimes not the case (e.g. when using the `openmm` option in CHARMM-GUI, like for the example provided above)

```
tmp = psfsections['MOLNT'][1]
tag_molecules(self)
if len(tmp) == len(self.atoms):
    # We have a CHARMM PSF file; now do NUMLP/NUMLPH sections
    self._process_lonepair_section(psfsections["NUMLP NUMLPH"])
```

Is there a reason, why `if len(tmp) == len(self.atoms):` is used for deciding whether the `_process_lonepair_section()` function is executed or not?
If not, I suggest a check like this:

```
if len(psfsections["NUMLP NUMLPH"][1]) != 0:
```


swails (approved):
Thanks!

PR #1274, opened 2022-10-31
Drude psf where NUMANISO is zero

florianjoerg:
When writing a CHARMM PSF file which has Drude particles, I ran into an error if the NUMANISO section is zero, because the `idx_section` list is empty in this case. I attached a psf to reproduce the error.
https://github.com/ParmEd/ParmEd/blob/c11babc6f19142134cc0401d946b1b0168073dc7/parmed/formats/psf.py#L317

```
  File "/home/florian/software/ParmEd_fork/minimal_example/test.py", line 3, in <module>
    psf.write_psf("new.psf")
  File "/scratch/data/florian/anaconda3/envs/parmed/lib/python3.11/site-packages/parmed/formats/psf.py", line 345, in write
    if idx_section[-1] == "\n":
       ~~~~~~~~~~~^^^^
IndexError: list index out of range
```
My suggestion would be to execute lines 317 to 348 only if `n_aniso` != 0.

[example.zip](https://github.com/ParmEd/ParmEd/files/9899553/example.zip)


swails:
Thanks!

Issue #1276, opened 2022-11-01
Matching patches to residues


peastman:
I'm investigating a problem (https://github.com/openmm/openmm/issues/3815) in which the ACE patch for the CHARMM polar force field didn't get converted correctly.  Part of the problem is in how it matches patches to residues.

In an OpenMM force field, a patch definition needs to include information that isn't specified in the CHARMM parameter files.  For example, when it lists an `ATOM` record, is that a new atom to be added or an existing atom to be modified?  When it lists a `BOND`, is that a new bond to be added, or is it replacing what was previously an external bond to a different residue?  To determine that information, we apply the patch to an actual residue and compare it to the original unpatched residue.

But which residue to choose?  CHARMM parameter files also don't specify which residues each patch is meant to be applied to.  So we try every possible residue.  Different residues may lead to different definitions for the patch, so we take a vote.  Whatever patch definition the largest number of residues produce, that's what we use.

For the ACE residue, that isn't working.  It's meant to be applied to amino acids, but in principle it could be applied to any residue containing an atom called `N`, which is a lot of them.  For example, lots of types of lipids.  In fact, there turn out to be more of those than there are amino acids.  And since `N` doesn't form an external bond in lipids, it therefore omits the `<RemoveExternalBond>` tag which is essential for applying it to amino acids.

One possible solution to this problem is that when it applies the patch to residues for the purpose of defining the patch information, it only considers ones that came from the same file.  The ACE patch is defined in toppar_drude_master_protein_2019h.str, so it would only consider residues found in that same file, not ones in other files like toppar_drude_lipid_2017c.str.

Do you think that would work?  Are there any situations it would break, where a single file contains patches but no residues those patches are meant to be applied to?  Can you think of a better solution?

peastman:
I don't see any existing mechanism to tell what file a patch or residue was loaded from, so presumably that would require creating one.

peastman:
@swails do you have any suggestions on the best way to handle this?  We need to come up with something, since it's a blocker for OpenMM 8.0.  If absolutely necessary I can put in a hack just for that one force field, but of course I'd much rather fix it in a clean, general way.

swails:
The solution you describe seems a little brittle to me, since I imagine (although don't know) that it's possible for people to create patches in str files that should be applied to residues outside that file.  We can certainly add a 

[...]

PR #1278, opened 2022-11-21
Prefer patch versions that replace external bonds

peastman:
This is a possible solution to #1276.  When identifying possible versions of a patch to write, it prefers ones that replace an external bond.  There are other possible solutions.  This is one I've implemented, and it seems to work.


peastman:
@swails any comments on this PR?

peastman:
Do you have any comments on this?

swails:
If tests pass, let's merge.

PR #1285, opened 2023-01-28
set `python_requires` to `>=3.8`

njzjz:
Python support has been restricted to Python v3.8 and newer, but `python_requires` is not set. Thus, the new version of ParmEd will still be installed when one use `pip` to install ParmEd on an older Python version. By setting `python_requires`, pip with old Python versions will not install the new version of ParmEd.


swails:
Thanks!  I'll cut a bugfix release and remove the old one.

PR #1286, opened 2023-02-03
Improve rdkit coordinate precision

bradakta:
It's possible to get the full precision in coordinates directly from rdkit even though PDB is used as the intermediary format (I know changing this is on the TODO list). 

The real nut to crack here is resolving bond orders - only mol2 and sdf really live or die by them.


parmed/rdkit/rdkit.py
```diff
@@ -30,7 +30,9 @@ def load(rmol):
         from ..formats.pdb import PDBFile
         from rdkit import Chem
         fh = StringIO(Chem.MolToPDBBlock(rmol))
-        return PDBFile.parse(fh)
+        struct = PDBFile.parse(fh)
+        struct.coordinates = rmol.GetConformer().GetPositions()
```
swails, line 34:
```suggestion
        struct.coordinates = rmol.GetConformer(0).GetPositions()
```

swails:
> The real nut to crack here is resolving bond orders - only mol2 and sdf really live or die by them.

#1284 added a lot in this area.  Residue templates from the PDB CCD have bond orders, and I postprocessed them to add atom hybridization perceived by RDKit for each of those templates.  It then maps these types and attributes to atoms in parsed PDB and PDBx/mmCIF files.

ParmEd can now store many of the RDKit attributes (like bond types, etc.), so it's probably worth prioritizing "proper" instantiation from RDKit models.

parmed/rdkit/rdkit.py
bradakta: I'm seeing this all now - I'm doing lots of ParmEd <--> rdkit conversions in order to use different functionalities in the same code. It looks like my old system of ad hoc copying bond orders no longer works with the new QualitativeBondType system -- I guess I'll need to figure this out.

The thing I would like to be able to do is read a PDB of a complex and then extract something like a ligand or cofactor as an rdkit object, but this causes trouble when I want rdkit to perceive bond orders itself.

swails:
`QualitativeBondType` is literally just a copy of RDKit's BondType enum.  All those new properties are exact copies of RDKit enums, so it should hopefully make the translation between ParmEd and RDKit pretty easy.

parmed/rdkit/rdkit.py
swails: > The thing I would like to be able to do is read a PDB of a complex and then extract something like a ligand or cofactor as an rdkit object, but this causes trouble when I want rdkit to perceive bond orders itself.

Something like this should do it:

```python
import parmed as pmd
from rdkit import Chem

struct = pmd.load_file("your_file.pdb", expanded_residue_template_match=True)

ligand = struct[":LIG"]
ligand_rdkit = pmd.rdkit.RDKit.to_mol(ligand)
Chem.SanitizeMol(ligand_rdkit, catchErrors=True)
```

That should give you an RDKit-sanitized molecule pretty quickly and easily.  Note that `to_mol` does things the "right" way (that is, it instantiates an RWMol directly and adds the attributes via the API rather than going through PDB or something similar).

PR #1290, opened 2023-02-17
switch out copy with deepcopy for structure.copy() method

### LINKED ISSUE #1289: Some properties are copied by reference in Structure.copy()
Some of properties in `parmed.Structure.copy()` is actually copied by reference (`atom.atom_type` specifically. This becomes an issue when the copied structure (e.g., apply nbfix) was modified also accidentally altered the original structure, which may lead to other unexpected behavior.

daico007:
Fix #1289. Basically switch out all the `copy()` call with `deepcopy()` in `parmed.Structure.copy()` function, ensuring that nothing is accidentally copied by reference.

### DIFF
--- a/parmed/structure.py
+++ b/parmed/structure.py
@@ -488,7 +488,7 @@ class Structure:
             c.torsion_torsion_types.append(deepcopy(tt))
         c.torsion_torsion_types.claim()
         for at in self.adjust_types:
-            c.adjust_types.append(cdeepcopyopy(at))
+            c.adjust_types.append(deepcopy(at))
         c.adjust_types.claim()
         # Now create the topological objects
         atoms = c.atoms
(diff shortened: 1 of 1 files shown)


parmed/structure.py
```diff
@@ -413,75 +433,76 @@ def copy(self, cls, split_dihedrals=False):
...
-            c.stretch_bend_types.append(copy(st))
+            c.stretch_bend_types.append(deepcopy(st))
         c.stretch_bend_types.claim()
         for tt in self.torsion_torsion_types:
-            c.torsion_torsion_types.append(copy(tt))
+            c.torsion_torsion_types.append(deepcopy(tt))
         c.torsion_torsion_types.claim()
         for at in self.adjust_types:
-            c.adjust_types.append(copy(at))
+            c.adjust_types.append(cdeepcopyopy(at))
```
swails, line 491:
```suggestion
            c.adjust_types.append(deepcopy(at))
```

PR #1303, opened 2023-04-04
Fix bug with restart reporter for OpenMM minimization

nffaruk:
There was a spurious argument in the `RestartReporter` function call for OpenMM minimization causing the error:

>    TypeError: __init__() got multiple values for argument 'write_velocities'

Getting rid of the `parm.ptr('natom')` argument resolves the error and brings the function in line with what is written to the script file (lines 553-554).


swails:
Thanks!

PR #1306, opened 2023-04-11
Flip guess strategy for atomic numbers

bradakta:
Invert the order of guessing so that the first two letters are tried first instead of second -- maybe there's a good reason why this wasn't done this way originally?

This lead to a bug in my code when a mol2 from AmberTools Antechamber was generated with the name/type CL1 and ParmEd kept reading it as carbon. This change fixes my case at least, but maybe it breaks something else I'm not thinking about?

### DIFF
--- a/parmed/periodic_table.py
+++ b/parmed/periodic_table.py
@@ -289,15 +289,14 @@ def element_by_name(name):
     if len(name) == 0:
         return Element[0]
     try:
-        atomic_number = AtomicNum[name[0].upper()]
-    except KeyError:
         sym = name[:2]
+        sym = '%s%s' % (sym[0].upper(), sym[1].lower())
+        atomic_number = AtomicNum[sym]
+    except (KeyError, IndexError):
         try:
-            sym = '%s%s' % (sym[0].upper(), sym[1].lower())
-            atomic_number = AtomicNum[sym]
-        except (KeyError, IndexError):
+            atomic_number = AtomicNum[name[0].upper()]
+        except KeyError:
             atomic_number = 0 # give up
-
     return Element[atomic_number]
 
 # Add some mass aliases here. We need to do it *after* _sorted_masses is created above, since


swails:
Most protein files (like PDBs) don't specify atomic numbers, and `CA` is a *very* common atom name used for carbon atoms.  In the majority of contexts, `CA` is actually carbon, whereas this change will recognize it as calcium instead.

We could get more sophisticated and start adding heuristics to determine element identity, but these are always imperfect.

Better would be to use an intermediate format that remembers the atomic numbers, or store them separately and assign them after ParmEd parses them (e.g., SDF, PDBx/mmCIF, or even Amber OFF libraries).

----

This is going to be really hard functionality to change.  It's one of those choices made long ago knowing there were potential drawbacks like this one, but we can't fix this particular issue without causing many others by changing the original choice.

bradakta:
I see how that's a huge issue. The difference between parsing a mol2 and a pdb is stark indeed. Is it possible to have both guessing schemes available as an option and then change defaults based on the file type? In the meantime, I think you're right - more active bookkeeping on the part of the user is the easiest solution.

drroe:
If a PDB is being written according to the format specification, it should be writing the element symbol (right-justified) in columns 77-78 - this is the unambiguous way to get atomic number from the PDB file. If that column is not present, you can revert to the guessing scheme.

swails:
> If a PDB is being written according to the format specification, it should be writing the element symbol (right-justified) in columns 77-78 - this is the unambiguous way to get atomic number from the PDB file. If that column is not present, you can revert to the guessing scheme.

This is how ParmEd already works, but the majority of PDB files I've worked with in the past do not actually have this information written to the file, so guessing was important to implement.  Since I was using it (and envisioned it being used for) primarily biomolecular simulation focused on large organic molecules, I decided to err on the side of assuming `C<something>` was a carbon rather than trying to be clever about it.

There are certainly improved heuristics that could be employed to make fewer mistakes, but none will ever be perfect and I just never prioritized spending time on it.

swails:
And for what it's worth, ParmEd _will_ write the element information that it has to the PDB file.  So that file format can always be used as a way to store element information.

bradakta:
That's a good point @drroe - there's probably more to it, but I'm seeing the main tension between PDB and MOL2 formats. The latter doesn't really have a standard for element info but programs for handling small molecules seem to use the convention <element symbol in caps><cumulative number of instances>.

To me at least, PDB without an element column feels like the special case rather than the best case of MOL2 above. Again though, it's a valid position to say that MOL2 is just under-specified and the user should keep track of that info for themselves.

PR #1310, opened 2023-05-31
Transfer chain IDs to OpenMM topology.

jpthompson17:
This PR sets the chain IDs in `Structure.topology` to match those in the `Structure` object.


swails:
Thanks! Automerge turned on, so this will go in after the builds pass.

PR #1311, opened 2023-06-08
Basic routines for internal to cartesian coordinate conversion

bradakta:
I hope there is agreement as to how something like this could be useful. I included a test to show how you can place atoms based on internal coordinates using a Z-matrix-like entry.

### DIFF
--- a/parmed/geometry.py
+++ b/parmed/geometry.py
@@ -314,6 +314,53 @@ def _cross(v1, v2):
                      v1[2]*v2[0] - v1[0]*v2[2],
                      v1[0]*v2[1] - v1[1]*v2[0]])
 
+def _unit_diff(a: np.ndarray, b: np.ndarray) -> np.ndarray:
+    """Unit vector going from a to b"""
+    u = (b - a).astype(np.float64)
+    u /= sqrt(np.dot(u, u))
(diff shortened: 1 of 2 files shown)


parmed/geometry.py
```diff
@@ -314,6 +314,53 @@ def _cross(v1, v2):
...
+    u /= sqrt(np.dot(u, u))
+    return u
+
+def _unit_cross(a: np.ndarray, b: np.ndarray) -> np.ndarray:
+    """Unit vector along cross product of a and b"""
+    u = _cross(a, b)
+    u /= sqrt(np.dot(u, u))
+    return u
+
+# Skipping type annotations for Atom bc it causes a circular import
```
swails, line 329: At the bottom of the file you can do this:

```python
if TYPE_CHECKING:  # must be imported from typing
    from .topologyobjects import Atom
```

Then you can use strings as type hints:

```python
def local_axes(atom1: "Atom", atom2: "Atom", atom3: "Atom") -> np.ndarray:
```

swails:
There are a number of places where it'd be useful to do Zmat to cartesian transformations (e.g., CHARMM RTF files).

I'm good adding this.

PR #1314, opened 2023-08-18
Remove unused scripts

swails:


swails:
I'll do this in another PR.  Environment solving is taking way too long.

PR #1315, opened 2023-08-21
Add a library of all CCD residues separate from modified amino/nucleic acids

swails:
This will allow doing bond/atom mapping for all standard 3-residue codes in the PDB.

### DIFF
--- a/.github/workflows/Test.yaml
+++ b/.github/workflows/Test.yaml
@@ -13,11 +13,11 @@ jobs:
     strategy:
       matrix:
         os: ["ubuntu-latest"]
-        python-version: ["3.8", "3.9"]
+        python-version: ["3.9", "3.10"]
 
     steps:
     - name: Check out source code
-      uses: actions/checkout@v2
+      uses: actions/checkout@v3
       with:
         fetch-depth: 0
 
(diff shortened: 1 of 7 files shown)


parmed/structure.py
```diff
@@ -1119,16 +1121,14 @@ def __getitem__(self, selection):
         scan = [x * y for x, y in zip(scan, selection)]
         # Copy all parameters
         struct = type(self)()
+        res_numbers_are_fake = {res.number for res in self.residues} == {0}
```
thomas-dresselhaus, line 1124: I actually think they are `-1` if they are fake. And if you do that here, then you might want to use it in `__repr__`, too.

thomas-dresselhaus (approved):
Thanks a lot, @swails . I'm confident that all issues I found are addressed here.

parmed/structure.py
swails: -1 is the (stupid) not-set default.  The existing logic was only checking if they were 0, which leads me to think that they were set to 0 in one of the parsers.

richardjgowers: necrocommenting here, but -1 is sometimes a valid resid (if residues have been inserted before the canonical 0).  Depending on the source of this (if this is internal book-keeping and you're sure it's never -1) then you might be better off using a safer sentinel value (-MAX_INT or something)

PR #1316, opened 2023-09-12
Preserve residue IDs, insertion codes, and bond orders in OpenMM topology conversions

jpthompson17:
This PR modifies `parmed.openmm.load_topology()` and `Structure.topology.getter()` in order to preserve additional topology information.

Changes to `parmed.openmm.load_topology()`:
- If an OpenMM residue `id` is the string representation of an integer, set the corresponding ParmEd residue `number` to that integer.
- ParmEd residue `insertion_code` <- OpenMM residue `insertionCode`
- If an OpenMM bond carries `type` data, use it to set the corresponding ParmEd bond `order`. Otherwise, use the OpenMM bond `order` (converted to a float).

Changes to `Structure.topology.getter()`:
- OpenMM residue `id` <- string representation of ParmEd residue `number`
- OpenMM residue `insertionCode` <- ParmEd residue `insertion_code`
- Use ParmEd bond `order` to determine OpenMM bond `type` and `order`.


parmed/structure.py
```diff
@@ -1689,17 +1689,41 @@ def topology(self):
...
+                if math.isclose(bond.order, o):
+                    type = t
+            # Determine the bond order (an integer)
+            order = None
+            n = round(bond.order)
+            if math.isclose(bond.order, n):
+                order = n
+            top.addBond(
+                atoms[bond.atom1.idx], atoms[bond.atom2.idx], type=type, order=order
+            )
```
swails, line 1726: I've also recently added a `qualitative_type` that is basically a translation of the RDKit bond types.  Not sure if that would be useful here or not.

swails (approved):
Thanks!

swails:
The build failures are fixed in #1315, so once that is merged this can be rebased and should run to completion.

The failures are a result of downstream updates in dependencies (scipy, AmberTools, and OpenMM).

parmed/structure.py
jpthompson17: Hmm... maybe. Would it make sense to add
```py
qualitative_type = None
if b.type is not None:
    name = repr(b.type).upper()
    if hasattr(QualitativeBondType, name):
        qualitative_type = getattr(QualitativeBondType, name)
```
to `parmed.openmm.load_topology()`, and do something like
```py
type = None
if bond.qualitative_type is not None:
    name = bond.qualitative_type.name.lower().capitalize()
    if hasattr(app, name):
        t = getattr(app, name)
        if t in bond_orders_by_openmm_type:  # ensure t is a recognized type
            type = t
else:  # try determining the type from the bond order
    for t, o in bond_orders_by_openmm_type.items():
        if math.isclose(bond.order, o):
            type = t
```
in `Structure.topology.getter()`?

swails: I'm tempted to say let's hold off on that until and unless it proves useful, and let's just get this change merged.

PR #1327, opened 2023-10-24
Fix setup.py

swails:
The nonstandard residue file was not getting included in the distribution builds because it was compressed and the extension didn't match what was in the setup.py file.


thomas-dresselhaus (approved):
Thanks!

PR #1348, opened 2024-04-24
Fix wildcard sorting when writing Amber parameter sets (fixes #1347)

### LINKED ISSUE #1347: Amber wildcard impropers are not properly sorted when writing to .frcmod/.parm
Thank you for all your hard work maintaining this project!  I recently ran into an issue recently using `AmberParameterSet` from `parmed.amber` to write wildcard dihedrals to a file.  The dihedrals written are not properly sorted such that wildcard atoms appear only at the beginning.

The offending lines of code can be found [here](https://github.com/ParmEd/ParmEd/blob/master/parmed/amber/parameters.py#L782-L797), or as follows:

```python
        for (a1, a2, a3, a4), typ in self.improper_periodic_types.items():
            # Make sure wild-cards come at the beginning
            if a2 == 'X':
                assert a4 == 'X', 'Malformed generic improper!'
                a1, a2, a3, a4 = a2, a4, a3, a1
            elif a4 == 'X':
                a1, a2, a3, a4 = a4, a1, a3, a2
            a1, a2, a4 = sorted([a1, a2, a4])
            if (a1, a2, a3, a4) in written_impropers:
                if written_impropers[(a1, a2, a3, a4)] != typ:
                    raise ValueError('Multiple impropers with the same atom set not allowed')
                continue
            outfile.write(
                f'{a1:<2}-{a2:<2}-{a3:<2}-{a4:<2} {typ.phi_k:14.8f} {typ.phase:8.3f} {typ.per:5.1f}\n'
            )
            written_impropers[(a1, a2, a3, a4)] = typ
        outfile.write('\n')
```

The first two if cases reorder a1, a2, and a4 such that wildcard atoms occur at the beginning.  However, the list is then resorted as `sorted([a1, a2, a4])` which typically places the wildcard "X" at the end of the list after many common atom types.

PR with a quick fix incoming promptly.

murfalo:
Fixes #1347

### DIFF
--- a/parmed/amber/parameters.py
+++ b/parmed/amber/parameters.py
@@ -786,7 +786,9 @@ class AmberParameterSet(ParameterSet, metaclass=FileFormatType):
                 a1, a2, a3, a4 = a2, a4, a3, a1
             elif a4 == 'X':
                 a1, a2, a3, a4 = a4, a1, a3, a2
-            a1, a2, a4 = sorted([a1, a2, a4])
+                a2, a4 = sorted([a2, a4])
+            else:
+                a1, a2, a4 = sorted([a1, a2, a4])
             if (a1, a2, a3, a4) in written_impropers:
                 if written_impropers[(a1, a2, a3, a4)] != typ:
                     raise ValueError('Multiple impropers with the same atom set not allowed')


swails:
Thanks!

PR #1351, opened 2024-05-24
Ignore TURN comments in PDBs

slochower:
PDB files in the protein-ligand benchmarks have `TURN` in their header, which prevents ParmEd from reading the PDB files correctly. I'm not sure how these were prepared, but I think it comes from Maestro. I'm also not sure if this is within spec of the PDB format, but this change seems innocuous and allows processing of the benchmark set in ParmEd.


slochower:
Meant to add this as reference: https://github.com/openforcefield/protein-ligand-benchmark/blob/a0b67b93c4f12f63a8bb417a1f6e4df4dc10c6c5/data/tyk2/01_protein/crd/protein.pdb#L137

swails (approved):

PR #1360, opened 2024-06-26
Bugfixed AtomType equality operator

### LINKED ISSUE #1359: Incorrect equality comparison in AtomType equality operator
In the equality operator for AtomType, True is returned if `abs(self.charge - other.charge) > TINY`.  Surely this should return False?

```python3
    def __eq__(self, other):
            # ...
            elif abs(self.charge - other.charge) > TINY:
                return True
            #...
```
(from topologyobjects.py [lines 5185-5186](https://github.com/ParmEd/ParmEd/blob/7f812971eb21db72bc814447379f585158f10159/parmed/topologyobjects.py#L5185-L5186))

murfalo:
Fixes #1359

### DIFF
--- a/parmed/topologyobjects.py
+++ b/parmed/topologyobjects.py
@@ -5183,7 +5183,7 @@ class AtomType:
                 if self.charge is not other.charge:
                     return False
             elif abs(self.charge - other.charge) > TINY:
-                return True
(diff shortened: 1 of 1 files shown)


swails:
Thanks!

PR #1367, opened 2024-08-27
Fix velocity unit scale for existing vels passed to OpenMM

### LINKED ISSUE #1366: Velocity off by factor of 10 (nm/ps vs Å/ps) when continuing a simulation with OpenMM from an Amber restart file
Hi, I've been experiencing consecutive simulations (e.g. NVT equil -> NPT equil) blowing up when running them through OpenMM with ParmEd. I've narrowed it down to ParmEd not converting the velocities that it gets from the amber restart file (Å/ps) to OpenMM's units (nm/ps).

nffaruk:
Closes #1366 - I thought I'd go ahead and submit a PR for it since there are a couple more fixes/mods I am going to submit anyway.


swails (approved):

nffaruk:
Nice, thanks for looking through and merging the various PRs

PR #1377, opened 2024-09-28
Fixed deepcopy for Atom.

ywu:
This PR is to fix a deepcopy-related bug that results in error for calls like `deepcopy(atom)` if Atom.__deepcopy__ is not defined.

### DIFF
--- a/parmed/topologyobjects.py
+++ b/parmed/topologyobjects.py
@@ -570,6 +570,10 @@ class Atom(_ListItem):
         """ Returns a deep copy of this atom, but not attached to any list """
         return type(self)._copy(self)
 
+    def __deepcopy__(self):
+        """ Returns a deep copy of this atom, but not attached to any list """
+        return type(self)._copy(self)
+
     #===================================================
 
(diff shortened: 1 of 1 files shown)


swails:
Are you sure this is needed? I had thought that deepcopy used `__copy__` if the `__deepcopy__` method wasn't defined. 

Can you share some code that this fixes?

ywu:
I added a unittest to hopefully demonstrate the difference (and fixed a bug in my previous change :blush:): Running the unittest with and without the `Atom.__deepcopy__` method defined, you should see `deepcopy(atom)` doesn't call `Atom._copy` if `Atom.__deepcopy__` is not defined.

swails:
I had operated under the assumption that `copy.deepcopy()` would just use `__copy__` if it `__deepcopy__` wasn't available, so if this is a problem for `Atom`, then it's a problem for every other class, too.

Can you give me a real-world script where the `copy.deepcopy()` semantics produces incorrect behavior?

PR #1391, opened 2025-02-14
Update dev environment and add testing for python 3.13

chrisjonesBSU:
A project I'm working on is looking at adding testing and support for python 3.13. Parmed is one of the dependencies that can't be installed via conda-forge with python 3.13. I played around with installing from source along with python 3.13, and everything seems to work correctly after bumping up the version of `rdkit` in the `environment-dev.yaml` file. I figured it would be worth making a PR that includes this change and adds python 3.13 to the test matrix.

### DIFF
--- a/.github/workflows/Test.yaml
+++ b/.github/workflows/Test.yaml
@@ -31,7 +31,7 @@ jobs:
         python-version: ${{ matrix.python-version }}
 
     - name: Optionally install AmberTools
-      if: ${{ matrix.python-version != 3.12 }}
+      if: ${{ matrix.python-version != 3.13 }}
       shell: bash -l {0}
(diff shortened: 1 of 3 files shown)


chrisjonesBSU:
I made one small change to [an openmm unit test](https://github.com/ParmEd/ParmEd/pull/1391/files#diff-96b01ffe269742337da4acf2653fd04ff70c85e864152a46d016a249c1274813). Per the release notes from [OpenMM 8.2](https://github.com/openmm/openmm/releases/tag/8.2.0) `LangevinMiddleIntegrator` is used instead of `LangevinIntegrator`.

swails (approved):

PR #1400, opened 2025-03-19
Support to modify cmaps based on phi psi mask CNCACN

kbelfon007:
Here we wanted to add support in parmed to 
1) load a frcmod file containing CMAPS
2) modify existing CMAP in the topology file with CMAP from the loaded frcmod
3) add CMAP to the backbone of cyclic amino acids

I added a new object CMAP to make these changes so that I don't make major changes to the code and how it operate. 

Here is an example:
```
# load topology of cyclic ALA ALA TRP ALA ALA
parm alaW.parm7

# load cmaps from file 
loadCmap ./ff19SB_cmaps.frcmod

#print all CMAPS in topology and loaded frcmod file 
printCmaps :*

# replace existing Trp (Leu) CMAP with Lys CMAP at residue 3
assignCmap :2@C :3@N :3@CA :3@C :4@N LYS

# add two new CMAP for the cyclic peptide 
# by default residue 1 and 5 won't have CMAP from leap 
assignCmap :4@C :5@N :5@CA :5@C :1@N ALA
assignCmap :5@C :1@N :1@CA :1@C :2@N ALA

setOverwrite
outparm ALA_Cyclic_mix_myedit.parm7
quit
```
~


parmed/amber/_amberparm.py
```diff
@@ -927,7 +927,333 @@ def load_rst7(self, rst7):
...
+                    
+                            parameters = data[idx+1:idx+n_parm_lines+1]
+                            self.cmaps_frcmod[cmap_title].set_parameters(parameters)
+        except (IndexError, KeyError, ValueError) as e:
+            raise AmberError(f'Error parsing CMAP file {frcmod}: {str(e)}')
+ 
+        #print (self.cmaps_frcmod)
+    #===================================================
+
+    def printCmaps(self):
```
swails, line 1019: Move this code to `actions.py` - it doesn't belong on `AmberParm`.

parmed/amber/_amberparm.py
```diff
@@ -927,7 +927,333 @@ def load_rst7(self, rst7):
         self.hasbox = self.box is not None
         if self.hasvels:
             self.velocities = rst7.vels
+    
+    #===================================================
+
+    def load_cmap_frcmod(self, frcmod):
```
swails, line 933: This code also all belongs in the action class, not the AmberParm class.

parmed/amber/_amberparm.py
```diff
@@ -2237,6 +2563,60 @@ def __setstate__(self, d):
                 setattr(self, attr, d[attr])
 
 
+# ++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+class CMAP(object):
```
swails, line 2567: What is this object used for? There's already a CMAP object in `topologyobjects`.

swails (commented):
It looks like you're trying to support prmtop modifications to support trying out ideas in force field development. This should all be encapsulated in the Action without leaking implementation to AmberParm.

The code you added should all go inside the `Action` class (and maybe a submodule for it) rather than tacking it onto the object (which is a public API used in many places) and calling it from the action.

You can see plenty of Action examples of the pattern that's preferable to a library like ParmEd.

Also, please add some tests so that future refactoring will not inadvertently break this functionality (including moving these functions to the corresponding `Action` subclasses).

Issue #1411, opened 2025-07-29
Gromacs topology input has issue for systems with NBFIX


feiglab:
In testing different force fields and preparing them for use in openMM, I found the following problem when reading Gromacs topology files via ParmEd as input:

I created a Gromacs topology file using Gromacs tools for a peptide system using the Gromacs implementation of the CHARMM c36m force field (which has NBFIX terms).

I read the Gromacs topology using ParmEd's load_file with Gromacs topology (.top) and coordinate (.gro) files as input. I get incorrect Lennard-Jones energies, compared against the following references (all of which agree once cutoffs are treated the same way):
- Energies from Gromacs
- Energies from CHARMM
- Energies from openMM using the CHARMM PSF reader with the CHARMM force field files
- Energies from openMM using an XML forcefield for the same CHARMM force field

I identified (and fixed in my local version) the following issue in the '_omm_nbfixed_force'  function (in structure.py):

NBFIX entries may not be considered due to compression of LJ parameters because the condition in line  2907 only checks one atom type but not both when deciding whether two atom types with the same LJ parameters can be compressed into one. 

The change that fixes the issue would be:

Instead of:
 elif not atype.nbfix:

we need:
 elif not atype.nbfix and not atype2.nbfix:

I note that the same issue applies to GromacsTopFile in the openmm package (which has another issue not present here).

I am happy to make a pull request.

PR #1412, opened 2025-07-29
fix issue with correctly processing NBFIX entries

feiglab:
addresses issue with NBFIX entries read from Gromacs topology file: https://github.com/ParmEd/ParmEd/issues/1411


feiglab:
it looks like test fails because of issue with main branch.

PR #1418, opened 2025-10-13
Load bond orders from rdkit smiles

CalCraven:
This should handle bond order parameters in a parmed structure when loaded via RDKit.


CalCraven:
I don't see how this code change would affect the two failing tests:
FAILED test_openmm_gromacs.py::TestGromacsTop::test_pme_switch - AssertionError: -18588.324341591044 != -18587.09715 with relative tolerance 0.0001 (1.000066)
FAILED test_parmed_unit.py::TestUnits::testMisc - AssertionError: '1.2 m' != 'Quantity(value=1.2, unit=meter)'
- 1.2 m

Could these be failing on master?

PR #1421, opened 2025-12-20
Merge request from Pengfei about C4 functions

pengfeili1:
Hi Jason,

I have made the following edits about ParmEd:

1. Add the onlywater option for add12_6_4 command;
2. add the changeC4AtomTypePair command, which changes the atom-type-pair C4 values;
3. fix a typo in the changeLJPair command.

Can you approve my pull request? Thanks!

Best,
Pengfei


swails:
Can you merge the latest master branch and add a test case to ensure we don't break this feature?

swails:
A few of the tests need to be updated (the `repr()` check can just be removed - it's unimportant, and the relative error check can be loosened so it passes).

I can do this if you can't, but won't have time until end of March, probably.

pengfeili1:
Hi Jason,

I have removed the the repr() check and loosen the relative error check for the test_add12_6_4_onlywater and test_change_c4_atom_type_pair cases.

Let me know whether it works or not.

Best,
Pengfei

swails:
I meant the ones that are failing (see [here](https://github.com/ParmEd/ParmEd/actions/runs/23030970343/job/66889063810))

You should add back the `str()` commands to at least make sure the `__str__` method doesn't throw an exception.

pengfeili1:
When I click the link, it just tells me that there is error, but I could not find the details.

I thought you meant deleting the str() commands for the repr() check issue, because I did not find any repr() in the two test cases. I just added the str() back. If you meant other code except the two cases I added, I could not do that because I am afraid to mess up other code.

I will leave it for you to do the fixes: by the end of March is fine, as long as the new commands appear in the AmberTools2026 release.

swails:
#1430 replaces this PR (incorporates its commits).

Issue #1426, opened 2026-02-12
"ParameterError: Not all cmap parameters found" with the gromacs 2026 and gmx_MMPBSA related with ParmEd


crf2213:
I ran into the same problem that was mentioned in the gmx_MMPBSA forum two weeks ago. Is this related to ParmEd? Can it be solved?

https://github.com/Valdes-Tresanco-MS/gmx_MMPBSA/issues/624

PR #1432, opened 2026-06-29
Cmap gmx26

mattiabondanza:
This is a basic implementation that should solve #1426 and other related issue in depending software. Just tested on an example system prepared in gmx 2026.1 with amber19sb and everything seems ok to me.

### DIFF
--- a/parmed/gromacs/gromacstop.py
+++ b/parmed/gromacs/gromacstop.py
@@ -1253,7 +1253,11 @@ class GromacsTopologyFile(Structure, TopFromStructureMixin, metaclass=FileFormat
             cmap_atom_types = sum([key for key in params.cmap_types], ())
             cmaptypes_w_res  = all(['-' in at for at in cmap_atom_types])
             cmaptypes_wo_res = all(['-' not in at for at in cmap_atom_types])
-            assert cmaptypes_w_res or cmaptypes_wo_res
+            
+            try:
+                assert cmaptypes_w_res or cmaptypes_wo_res
+            except AssertionError:
+                raise ParameterError("All cmaps should be defined either with or without residues."
 
             if cmaptypes_wo_res:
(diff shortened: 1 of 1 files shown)


parmed/gromacs/gromacstop.py
```diff
@@ -1248,11 +1248,76 @@ def update_typelist_from(ptypes, types):
...
+                    c.type.used = True
+                else:
+                    raise ParameterError('Not all cmap parameters found')
+            
+            if cmaptypes_w_res:
+                def _getresidue(a):
+                    return a.residue.name
+                
+                def compare_reskeys(a, b):
+                    if type(a) != tuple:
```
swails, line 1275:
```suggestion
                    if not isinstance(a, tuple):
```

This will break if you end up using a tuple subclass (like NamedTuple). It's better to use `isinstance` checks than type checking this way.

parmed/gromacs/gromacstop.py
```diff
@@ -1248,11 +1248,76 @@ def update_typelist_from(ptypes, types):
...
+                        #print("A not a tuple")
```
swails, line 1276: Delete these commented debug lines.

swails (commented):
Thanks! Just a few recommended changes.

Do you have a test case you can add?

mattiabondanza:
Hi I do have a test but it's a very large system I am using for research. I don't think it's appropriate to use this as test. I think that any protein system prepared with the correct version of gmx/forcefield should work. If I have time I will prepare something in the next days.

vedranmiletic:
@mattiabondanza FWIW, we used PyMOL's [fab](https://pymolwiki.org/Fab) function, e.g. `fab MPCDF`, to build the peptide test set when validating the ported ff19SB and it worked well for the purpose.

